PR #16, opened 2018-11-28
cmd/firectl: adding more verbose error messages and testdata folder

xibz:
This change improves the error messaging when verifying the `firectl`'s configuration structure.
Also, this improves the test story by adding a `testdata` folder that should house things like `.img` file and other potential testing data. `FC_TEST_DATA_PATH` can be set to override the test data folder path to include custom `.img` and` vmlinux` files.


cmd/firectl/main.go
```diff
@@ -26,29 +26,26 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// checkConfig will ensure that the required fields are set
+// in the config. This will also check to ensure that the socket
+// path exists.
 func checkConfig(cfg firecracker.Config) error {
```
mxpv: I think it'll be useful moving `Config` validations to `firecracker` package.

xibz: Cool, moved to `firecracker` package.

mxpv: Can you also validate configuration passed to `NewMachine`? To make sure input is correct.

xibz: Hm, do we want want to validate in there? I think users could call `Validate` themselves on the config to ensure their configuration file is correct, similarly to how it is being done in `firectl`. That way validating is optional. If `firecracker` drops a required field, users could turn off validation until the SDK has been updated. What do you think? I can be swayed either way.

mxpv: I think we need validation at some point. It can be either `NewMachine` or `Init`.
Let's say user forgot to pass `KernelImagePath`, we want to fail as early as possible and not run Firecracker process at all. Also user might forget to call `Validate`, which will add more confusion than flexibility. Doing that in `NewMachine` feels more 'go way' to me. 

@nmeyerhans do you have opinion regarding this?

machine_test.go
```diff
@@ -251,3 +255,9 @@ func TestWaitForSocket(t *testing.T) {
 		t.Error("waitForSocket did not properly detect program exit")
 	}
 }
+
+func init() {
```
samuelkarp: Can you move this to the top of the file?

testdata/drive-2.img
```diff
@@ -0,0 +1 @@
+dd if=/dev/zero of=drive-2.img bs=1k count=102400
```
samuelkarp: Why do you have a dd command here?

xibz: Just cp'd the image file, but will go ahead and remove the command.

Issue #10, opened 2018-11-27
Support additional VM console options


nmeyerhans:
We currently support several console options for the VM, including `stdio` to use the existing IO streams or `xterm` to run in a graphical terminal. Additional ones that could be useful are:

1. Arbitrary file -- Log stdout/stdio to an arbitrary file and read stdin from /dev/null
2. Inherited file descriptor. -- Allow the client to provide existing file descriptors for any or all of the standard IO streams.

Maybe others?

xibz:
closing in result of #17

PR #17, opened 2018-11-29
cmd/firectl: adding support for file based logging

xibz:
This change introduces a new console type of `file` which allows users to specify a files to write logs to using the `log-stdout-file` and `log-stderr-file` environment variables.

Refactors how VMM `exec.Command` is constructed. Prior this was done in `startVMM`, but that led to some coupling with the `CLI`. This PR introduces a new functional option when creating a machine, `WithProcessRunner` that can be used to pass in a custom `exec.Cmd`.

Issue #10


machine.go
```diff
@@ -578,3 +606,41 @@ func (m *Machine) waitForSocket(timeout time.Duration, exitchan chan error) erro
...
+func (m *Machine) Finalize() error {
+	buf := bytes.Buffer{}
+	for _, fn := range m.cleanupFns {
+		if err := fn(); err != nil {
+			buf.WriteString(err.Error())
+			buf.WriteString("\n")
+		}
+	}
+
+	if buf.Len() > 0 {
```
mxpv: I suggest looking at `github.com/hashicorp/go-multierror`

machine.go
```diff
@@ -578,3 +606,41 @@ func (m *Machine) waitForSocket(timeout time.Duration, exitchan chan error) erro
 
 	return <-done
 }
+
+// Finalize is a clean up that needs to be called when the Machine
+// has finished its task.
+func (m *Machine) Finalize() error {
```
mxpv: `Finalize` comes from Java/C#. I suggest using `Close` instead.

machine.go
```diff
@@ -295,16 +324,16 @@ func (m *Machine) startVMM(ctx context.Context) (chan error, error) {
...
-			m.logger.Warnf("firecracker exited: %s", err.Error())
-		} else {
-			m.logger.Printf("firecracker exited: status=0")
-		}
+	m.cleanupFns = append(m.cleanupFns, func() error {
 		os.Remove(m.cfg.SocketPath)
 		os.Remove(m.cfg.LogFifo)
 		os.Remove(m.cfg.MetricsFifo)
+
+		return nil
```
mxpv: nit: Might be useful to use `multierror` here as well?

xibz: Yea, I think so. Will make the change!

nmeyerhans:
We should also add the ability to redirect stderr independently. The VMs console goes to stdout, but the SDK's own output goes to stderr. The flexibility of separating them will be useful.

machine.go
```diff
@@ -578,3 +614,36 @@ func (m *Machine) waitForSocket(timeout time.Duration, exitchan chan error) erro
...
+
+// generateLogFile will create a file using the filename parameter.
+// If the filename is not a file but a directory, then the
+// defaultLogFileName will be used instead.
+func generateLogFile(filename string) (*os.File, error) {
+	info, err := os.Stat(filename)
+	// if the error is not exists, then we assume the user has
+	// provided a file path, including the file, but it does not
+	// exist. In that case this function will create that file.
+	if !os.IsNotExist(err) && err != nil {
```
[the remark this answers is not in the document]
xibz: I changed this to require a directory to be passed in. So, this needs to `Stat` anyways to check if it is a directory. However, changed it to return the error if `Stat` returned one.

cmd/firectl/main.go
```diff
@@ -181,26 +197,85 @@ func main() {
...
+	fn := func() error {
+		return nil
+	}
+
+	switch opts.FcConsole {
+	case ConsoleXterm:
+		bin := builder.Bin()
+		args := builder.Args()
+		builder = builder.WithBin(terminalProgram).
+			WithArgs(append([]string{"-e", bin}, args...))
```
mxpv: I'm not sure if this approach is better, as for me, extra level of indirection makes things harder to understand.
Consider this version:
```go
	switch m.cfg.Console {
	case ConsoleXterm:
		cmd = exec.CommandContext(ctx, xtermPath, "-e", m.binPath)
	case ConsoleStdio:
		cmd = exec.CommandContext(ctx, m.binPath)
		cmd.Stdin = os.Stdin
		cmd.Stdout = os.Stdout
		cmd.Stderr = os.Stderr
	case ConsoleFile:
		stdout, stdout, err := openLogFiles(...)
		if err != nil {
			...
		}
		cmd = exec.CommandContext(ctx, m.binPath)
		cmd.Stdout = stdout
		cmd.Stderr = stderr
	default:
		cmd = exec.CommandContext(ctx, m.binPath)
	}
	
	cmd.Args = append(cmd.Args, "--api-sock", m.cfg	.SocketPath)
```

cmd/firectl/main.go
```diff
@@ -181,26 +197,85 @@ func main() {
...
 	defer vmmCancel()
-	errchan, err := m.Init(ctx)
+
+	errCh, err := m.Init(ctx)
 	if err != nil {
 		log.Fatalf("Firecracker Init returned error %s", err)
-	} else {
-		m.StartInstance(vmmCtx)
 	}
+	m.StartInstance(vmmCtx)
```
mxpv: I think we have to check errors here as well.

cmd/firectl/main.go
```diff
@@ -181,26 +197,85 @@ func main() {
 		Debug:             opts.Debug,
 	}
 
-	m, err := firecracker.NewMachine(fcCfg, firecracker.WithLogger(log.NewEntry(logger)))
+	ctx := context.Background()
+	cmd, cleanup, err := buildCommand(ctx, fcCfg.SocketPath, opts)
```
mxpv: Why do we want this outside `firecracker` package?
And if we do really want, consider approach similar to `FirecrackerClient` (add something like `WithFirecrackerProcessRunner` opt and fallback implementation, if not provided).

xibz: Are you asking why do we want the `buildCommand` to be outside of the `firecracker` package?

mxpv: Why do we force user to manage `Command` stuff.

xibz: Ah, because it is more extensible. The goal for the command builder was to make it easy for users to deal with it.

Users can now specify how they want to start the VMM, which is more powerful than doing it all under the hood. If we want, we can add a `DefaultFirecrackerCommand` that is a prebuild command that users can use. A good example of a user wanting to extend may be they want to remote start a command via SSH. The prior implementation made that impossible to do.

mxpv: `firecracker.NewMachine(cfg, firecracker.WithLogger(), firecracker.WithProcessRunner(cmd))`

mxpv: It makes sense to provide default implementation as well.

xibz: Yea, `WithProcessRunner` sounds like a good idea. If one isn't provided use the default one. What do you think?

mxpv: Yes!

xibz:
Ah, interesting. I'll have to write some tests around that.

cmd/firectl/main.go
xibz: I only changed the `ConsoleXterm` case as that one was a little funky. I've kept the other ones the same aren't difficult to read.

machine.go
```diff
@@ -112,12 +106,12 @@ func (cfg *Config) Validate() error {
 
 // Machine is the main object for manipulating firecracker VMs
 type Machine struct {
-	binPath       string
 	cfg           Config
 	client        Firecracker
 	cmd           *exec.Cmd
 	logger        *log.Entry
 	machineConfig models.MachineConfiguration // The actual machine config as reported by Firecracker
+
```
nmeyerhans: nit: stray whitespace

Issue #12, opened 2018-11-27
Support configuration of the Firecracker MMDS


nmeyerhans:
Firecracker has an internal metadata service that can provide user-supplied data to the VM via an http endpoint. We should support configuration of this service.

PR #18, opened 2018-11-30
Add Metadata support

yuval-k:
*Issue #:* #12

*Description of changes:*
* Adding metadata support in the api & cli
* Added allow mmds to network card (automatically set to enabled in cli if metadata provided)
* Added a test in machine_test.go and tested manually. made sure test pass with `make test`
* Added `make generate` target for convenience and update HACKING.md

### DIFF
--- a/client/swagger.yaml
+++ b/client/swagger.yaml
@@ -257,8 +257,10 @@ paths:
           schema:
             type: object
       responses:
+        201:
+          description: MMDS data store created
         204:
-          description: MMDS data store created/updated.
+          description: MMDS data store updated.
         400:
           description: MMDS data store cannot be created due to bad input.
           schema:
@@ -393,7 +395,7 @@ definitions:
 
   Error:
     properties:
-      faultMessage:
+      fault_message:
         type: string
         description: A description of the error condition
 
--- a/cmd/firectl/main.go
+++ b/cmd/firectl/main.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -107,6 +108,7 @@ func main() {
 		FcCPUCount         int64    `long:"ncpus" short:"c" description:"Number of CPUs" default:"1"`
 		FcCPUTemplate      string   `long:"cpu-template" description:"Firecracker CPU Template (C3 or T2)"`
 		FcMemSz            int64    `long:"memory" short:"m" description:"VM memory, in MiB" default:"512"`
+		FcMetadata         string   `long:"metadata" description:"Firecracker Meatadata for MMDS (json)"`
 		Debug              bool     `long:"debug" short:"d" description:"Enable debug output"`
 		Help               bool     `long:"help" short:"h" description:"Show usage"`
 	}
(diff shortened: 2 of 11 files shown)


yuval-k:
Also note: I updated swagger.yaml due to some differences between it and the implementation.

machine.go
```diff
@@ -530,6 +533,28 @@ func (m *Machine) startInstance(ctx context.Context) error {
 	return err
 }
 
+// SetMetadata sets the machine's metadata for MDDS
+func (m *Machine) SetMetadata(ctx context.Context, metadata interface{}) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
mxpv: Why do we need `cancel` here?

yuval-k: TBH I copied the pattern from `startInstance`. I suspect it's there to cancel out any kept alive connections, but I don't really know.
I'll be happy to remove this from here and `startInstance` if you think that's the correct thing to do.

client/swagger.yaml
```diff
@@ -393,7 +395,7 @@ definitions:
 
   Error:
     properties:
-      faultMessage:
+      fault_message:
```
nmeyerhans: :100: Thanks for noticing this. Please submit your swagger changes to the firecracker project's [swagger file](https://github.com/firecracker-microvm/firecracker/blob/master/api_server/swagger/firecracker.yaml) as well!

cmd/firectl/main.go
```diff
@@ -131,6 +133,13 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
+	var metadata interface{}
+	if opts.FcMetadata != "" {
+		if err := json.Unmarshal([]byte(opts.FcMetadata), &metadata); err != nil {
+			log.Fatalf("Unable to metadata as json: %s", err)
```
nmeyerhans: I think you're missing a word here. "Unable to parse metadata as json", maybe?

machine.go
```diff
@@ -530,6 +533,28 @@ func (m *Machine) startInstance(ctx context.Context) error {
...
+		var message string
+		if respcreated != nil {
+			message = respcreated.Error()
+		}
+		if respnocontent != nil {
+			message = respnocontent.Error()
+		}
+		m.logger.Printf("SetMetadata successful: %s", message)
+	} else {
+		m.logger.Printf("Error setting metadata: %s", err)
```
nmeyerhans: This should be logged at Error severity, and drop "Error" from the message string.

yuval-k: this pattern was too copied from `startInstance` - would you like me to change it as well?

client/swagger.yaml
yuval-k: https://github.com/firecracker-microvm/firecracker/pull/713

machine.go
mxpv: I don't think it's needed. Please do so.

machine.go
```diff
@@ -518,14 +521,30 @@ func (m *Machine) startInstance(ctx context.Context) error {
...
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	resp, err := m.client.CreateSyncAction(ctx, &info)
 	if err == nil {
 		m.logger.Printf("startInstance successful: %s", resp.Error())
 	} else {
-		m.logger.Printf("Error starting instance: %s", err)
+		m.logger.Errorf("Starting instance: %s", err)
```
nmeyerhans: :+1: At some point I'll do an audit of all our messages and make sure the severities and language is correct. Thanks for taking care of this one.

nmeyerhans:
I've tested this with `firectl` and it looks good. Let's squash and cleanup the commits before merging, but otherwise, :shipit:!

yuval-k:
usually what we do internall is use github's "Squash and Merge" feature - that way you get a reference to the PR in the commit log

yuval-k:
Alternatively I can squash and force push. let me know what you prefer.

nmeyerhans:
Squash and force-push, please.

yuval-k:
squash everything to one commit and force pushed

PR #21, opened 2018-12-04
flags log-stdout-file and log-stderr-file are optional

xibz:
Changes that if `--firecracker-console` is set to `file` and none if the `log-*-file` flags were set, then it will simply use the default `io.Writer`s during command construction.


nmeyerhans:
Let's just drop most console handling from firectl. We can handle that from the shell, and we don't need to reconstruct that. I'm more interested in how the SDK helps an application writer manage the console.

xibz:
Done! Also, added the verbose flag that I mentioned since it was odd to have `firecracker-console` to have only one flag. Seemed more natural to have it default to `none` then with `verbose` enabled, use `stdio` and `stderr`.

cmd/firectl/main.go
```diff
@@ -218,16 +212,38 @@ func main() {
...
+			log.Fatalf("Binary, %q, does not exist: %v", opts.FcBinary, err)
+		}
+
+		if err != nil {
+			log.Fatalf("Failed to stat binary, %q: %v", opts.FcBinary, err)
+		}
+
+		if finfo.IsDir() {
+			log.Fatalf("Binary, %q, is a directory", opts.FcBinary)
+		} else if finfo.Mode()&0111 == 0 {
```
samuelkarp: nit: I'd prefer to have the mask as a constant.

mxpv: I'm not sure we need this checks at all. Why not just to rely on `exec.Command` ?

xibz: The errors returned from `exec.Command` can be a little ambiguous. This is to be more specific. For instance if the binary is a directory, `exec.Command` will return "permission denied."

PR #23, opened 2018-12-04
adding fifo logging files and refactoring the SDK

xibz:
*Description of changes:*
`LogFifo` and `MetricsFifo` can now log its contents to a file by setting the `FifoLogWriter` which is an `io.Writer` to handle the fifo streams content.

Some refactors here on how the `Machine` object flows through its instructions. Previously these instructions were all called during the initialization phase. However, to make the `Machine` more flexible and extendable, this introduces handler lists which is a command pattern queue. `Init` is no longer needed, as that is now handled through `Handlers`.

### DIFF
--- a/cmd/firectl/main.go
+++ b/cmd/firectl/main.go
@@ -18,11 +18,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
 	"github.com/firecracker-microvm/firecracker-go-sdk"
+	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 	"github.com/jessevdk/go-flags"
 	log "github.com/sirupsen/logrus"
 )
@@ -43,28 +47,37 @@ const (
 	executableMask = 0111
 )
 
-func parseBlockDevices(entries []string) ([]firecracker.BlockDevice, error) {
-	var devices []firecracker.BlockDevice
-	for _, entry := range entries {
-		var path string
+func parseBlockDevices(entries []string) ([]models.Drive, error) {
+	devices := []models.Drive{}
+
+	for i, entry := range entries {
+		path := ""
+		readOnly := true
+
 		if strings.HasSuffix(entry, ":rw") {
+			readOnly = false
 			path = strings.TrimSuffix(entry, ":rw")
 		} else if strings.HasSuffix(entry, ":ro") {
 			path = strings.TrimSuffix(entry, ":ro")
 		} else {
 			msg := fmt.Sprintf("Invalid drive specification. Must have :rw or :ro suffix")
-			return []firecracker.BlockDevice{}, errors.New(msg)
+			return nil, errors.New(msg)
 		}
+
 		if path == "" {
 			return nil, errors.New("Invalid drive specification")
 		}
-		_, err := os.Stat(path)
-		if err != nil {
+
+		if _, err := os.Stat(path); err != nil {
 			return nil, err
 		}
-		e := firecracker.BlockDevice{
-			HostPath: path,
-			Mode:     "rw",
+
+		e := models.Drive{
+			// i + 2 represents the drive ID. We will reserve 1 for root.
+			DriveID:      firecracker.String(strconv.Itoa(i + 2)),
+			PathOnHost:   firecracker.String(path),
+			IsReadOnly:   firecracker.Bool(readOnly),
+			IsRootDevice: firecracker.Bool(false),
 		}
 		devices = append(devices, e)
 	}
@@ -105,6 +118,76 @@ func parseVsocks(devices []string) ([]firecracker.VsockDevice, error) {
 	return result, nil
 }
 
+func createFifoFileLogs(fifoPath string) (*os.File, error) {
+	return os.OpenFile(fifoPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+}
+
+// handleFifos will see if any fifos need to be generated and if a fifo log
+// file should be created.
+func handleFifos(opts *options) (io.Writer, []func() error, error) {
+	// these booleans are used to check whether or not the fifo queue or metrics
+	// fifo queue needs to be generated. If any which need to be generated, then
+	// we know we need to create a temporary directory. Otherwise, a temporary
+	// directory does not need to be created.
+	generateFifoFilename := false
+	generateMetricFifoFilename := false
+	cleanupFns := []func() error{}
+	var err error
+
+	var fifo io.WriteCloser
+	if len(opts.FcFifoLogFile) > 0 {
+		if len(opts.FcLogFifo) > 0 {
+			return nil, cleanupFns, fmt.Errorf("vmm-log-fifo and firecracker-log cannot be used together")
+		}
+
+		generateFifoFilename = true
+		// if a fifo log file was specified via the CLI then we need to check if
+		// metric fifo was also specified. If not, we will then generate that fifo
+		if len(opts.FcMetricsFifo) == 0 {
+			generateMetricFifoFilename = true
+		}
+
+		if fifo, err = createFifoFileLogs(opts.FcFifoLogFile); err != nil {
+			return fifo, cleanupFns, fmt.Errorf("Failed to create fifo log file: %v", err)
+		}
+
+		cleanupFns = append(cleanupFns, func() error {
+			return fifo.Close()
+		})
+	} else if len(opts.FcLogFifo) > 0 || len(opts.FcMetricsFifo) > 0 {
+		// this checks to see if either one of the fifos was set. If at least one
+		// has been set we check to see if any of the others were not set. If one
+		// isn't set, we will generate the proper file path.
+		if len(opts.FcLogFifo) == 0 {
+			generateFifoFilename = true
+		}
+
+		if len(opts.FcMetricsFifo) == 0 {
+			generateMetricFifoFilename = true
+		}
+	}
+
+	if generateFifoFilename || generateMetricFifoFilename {
+		dir, err := ioutil.TempDir(os.TempDir(), "fcfifo")
+		if err != nil {
+			return fifo, cleanupFns, fmt.Errorf("Fail to create temporary directory: %v", err)
+		}
+
+		cleanupFns = append(cleanupFns, func() error {
+			return os.RemoveAll(dir)
+		})
+		if generateFifoFilename {
+			opts.FcLogFifo = filepath.Join(dir, "fc_fifo")
+		}
+
+		if generateMetricFifoFilename {
+			opts.FcMetricsFifo = filepath.Join(dir, "fc_metrics_fifo")
+		}
+	}
+
+	return fifo, cleanupFns, nil
+}
+
 type options struct {
 	FcBinary           string   `long:"firecracker-binary" description:"Path to firecracker binary"`
 	FcKernelImage      string   `long:"kernel" description:"Path to the kernel image" default:"./vmlinux"`
@@ -117,12 +200,12 @@ type options struct {
 	FcLogFifo          string   `long:"vmm-log-fifo" description:"FIFO for firecracker logs"`
 	FcLogLevel         string   `long:"log-level" description:"vmm log level" default:"Debug"`
 	FcMetricsFifo      string   `long:"metrics-fifo" description:"FIFO for firecracker metrics"`
-	FcCaptureFifoLogs  bool     `long:"capture-fifo-logs" description:"pipes fifo and metric fifo's contents to files"`
 	FcDisableHt        bool     `long:"disable-hyperthreading" short:"t" description:"Disable CPU Hyperthreading"`
 	FcCPUCount         int64    `long:"ncpus" short:"c" description:"Number of CPUs" default:"1"`
 	FcCPUTemplate      string   `long:"cpu-template" description:"Firecracker CPU Template (C3 or T2)"`
 	FcMemSz            int64    `long:"memory" short:"m" description:"VM memory, in MiB" default:"512"`
 	FcMetadata         string   `long:"metadata" description:"Firecracker Meatadata for MMDS (json)"`
+	FcFifoLogFile      string   `long:"firecracker-log" short:"l" description:"pipes the fifo contents to the specified file"`
 	Debug              bool     `long:"debug" short:"d" description:"Enable debug output"`
 	Help               bool     `long:"help" short:"h" description:"Show usage"`
 }
@@ -165,7 +248,7 @@ func main() {
 		if err != nil {
 			log.Fatalf("Unable to parse NIC config: %s", err)
 		} else {
-			log.Printf("Adding tap device %s", tapDev)
+			log.Error("Adding tap device %s", tapDev)
 			allowMDDS := metadata != nil
 			NICs = []firecracker.NetworkInterface{
 				firecracker.NetworkInterface{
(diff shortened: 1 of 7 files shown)


handlers.go
```diff
@@ -0,0 +1,167 @@
...
+	CreateBootSourceHandlerName        = "CreateBootSource"
+	AttachDrivesHandlerName            = "AttachDrives"
+	CreateNetworkInterfacesHandlerName = "CreateNetworkInterfaces"
+	AddVsocksHandlerName               = "AddVsocks"
+)
+
+// StartVMMNamedHandler is a named handler that will
+// handle starting of the VMM. This handler will also
+// set the exit channel on completion.
+var StartVMMNamedHandler = NamedHandler{
```
mxpv: What kind of problem are you trying to solve with this? Why not to call functions directly?

xibz: Makes testing easier when testing individual components also allows for extendable functionality without the `Init` function becoming this huge behemoth. Allows the machine to not require any knowledge of what it is executing as well, making it a little more of a black box of commands. All the machine should care about is whether or not a command has failed.

mxpv: Why not use interfaces instead?

samuelkarp: I think I'm comfortable with the handler approach.  It does add a lot of flexibility; the main concern that I have right now is whether we need it.  But I also don't really object to having it.

machine.go
```diff
@@ -162,116 +158,89 @@ func NewMachine(cfg Config, opts ...Opt) (*Machine, error) {
...
+	return nil
 }
 
 // startVMM starts the firecracker vmm process and configures logging.
-func (m *Machine) startVMM(ctx context.Context) (<-chan error, error) {
+func (m *Machine) startVMM(ctx context.Context) error {
 	m.logger.Printf("Called startVMM(), setting up a VMM on %s", m.cfg.SocketPath)
 
-	exitCh := make(chan error)
+	m.ErrCh = make(chan error)
```
samuelkarp: I don't think I really like `m.ErrCh` being public.  This function uses the channel for internal signalling before returning, and I don't think it wise for that to be exposed while the function is executing.  Having it in the return meant that it was inaccessible until the function returned, but moving it into the struct and making it public loses that.  I think I'd prefer it to just be returned with the function, but if you think that's not possible, an alternative is not assigning it to the public field in the struct until the end of this function, after its internal dependencies are done trying to read from the channel (i.e., after the call to `waitForSocket`).

machine.go
```diff
@@ -162,116 +158,89 @@ func NewMachine(cfg Config, opts ...Opt) (*Machine, error) {
...
 
-	for id, dev := range m.cfg.AdditionalDrives {
-		// id must be increased by 2 because firecracker uses 1-indexed arrays and the root drive occupies position 1.
-		err = m.attachDrive(ctx, dev, id+2, false)
-		if err != nil {
-			m.logger.Errorf("While attaching secondary drive %s, got error %s", dev.HostPath, err)
+func (m *Machine) addVsocks(ctx context.Context, vsocks ...VsockDevice) error {
+	for _, dev := range m.cfg.VsockDevices {
+		if err := m.addVsock(ctx, dev); err != nil {
 			m.stopVMM()
```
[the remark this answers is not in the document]
xibz: Yea, I think that makes a lot of sense. +1

machine.go
```diff
@@ -303,20 +270,26 @@ func (m *Machine) startVMM(ctx context.Context) (<-chan error, error) {
...
 	}
 
 	m.logger.Debugf("returning from startVMM()")
-	return exitCh, nil
+	return nil
+}
+
+// Wait will wait for the microVM to finish execution or if an error was
+// encountered.
+func (m *Machine) Wait() error {
```
samuelkarp: Would it be worth having `Wait` take a `context.Context` so it can be canceled?

machine.go
```diff
@@ -111,6 +103,10 @@ type Machine struct {
 	cmd           *exec.Cmd
 	logger        *log.Entry
 	machineConfig models.MachineConfiguration // The actual machine config as reported by Firecracker
+
+	Metadata interface{}
```
samuelkarp: Can you add godoc for exported fields?

machine.go
xibz: After some thought, I think it makes a lot of sense to do this. Originally I was wondering if we should follow the same signature as the `exec.Cmd` wait call, but due to the difference in complexity I think it makes sense to pass a context in.

samuelkarp: I think so too.  It would be possible to wrap the call to `Wait` like the following:

```go
ctx, _ := context.WithTimeout(context.Background(), timeout)
errCh := make(chan errCh)
go func(){
	errCh <- m.Wait()
}()
select {
	case <- ctx.Done():
		return errors.New("timed out")
	case err <- errCh:
		return err
}
```

But this is a lot of code, and it leaves a dangling goroutine.  The risk of the dangling goroutine is that if you then call `Wait` again, it's nondeterministic which one will actually get the error and return, and which one will block.

xibz:
Added a test for that as I assumed the handler would ignore `nil` values but never added that functionality. I have instead do a `nil` check and will only enable metadata support when `metadata` isn't `nil`.

machine.go
```diff
@@ -112,13 +88,20 @@ type Config struct {
 	// to the microVM.
 	NetworkInterfaces []NetworkInterface
 
+	// CaptureFifoLogsToFile will copy the contents from the given fifo
```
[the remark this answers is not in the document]
xibz: That makes sense. I am not too familiar with unlinking, but from the man pages it states that will delete a name from a file system if no process has it opened. Since we'll have it opened, unlinking would fail?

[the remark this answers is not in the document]
nmeyerhans: No, the directory entry will be removed. `unlink()` doesn't know or care about open files. From the manual:

> If the name was the last link to a file but any processes still have the file open, the file will remain in existence until the last file descriptor referring to it is closed.

and

> If  the name referred to a socket, FIFO, or device, the name for it is removed but processes which have the object open may continue to use it. 

This is precisely the behavior we want.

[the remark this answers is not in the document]
xibz: Done

xibz:
I think you have an existing socket file somewhere.

cmd/firectl/main.go
```diff
@@ -188,23 +195,36 @@ func main() {
 		log.Fatalf("Invalid vsock specification: %s", err)
 	}
 
+	if len(opts.FcFifoLogFile) != 0 {
+		if len(opts.FcLogFifo) == 0 {
+			opts.FcLogFifo = filepath.Join(os.TempDir(), fmt.Sprintf("fifo_%s", randBytes(16)))
```
nmeyerhans: Constructing a file path out of a pure pseudorandom byte sequence is unsafe and prone to failure or unexpected behavior.

Consider the case where a byte corresponds to a non-printing character such as a carriage return or a NULL character. Or where a byte corresponds to a character with special meaning to the filesystem, such as `0x2f`.

You might want to check out what [ioutil.TempFile()](https://golang.org/src/io/ioutil/tempfile.go#L39) is doing. Or use [ioutil.TempDir()](https://godoc.org/io/ioutil#TempDir) to create a temporary directory, then create the fifo in it with an arbitrary valid filename (can be a constant). After you open the fifo for reading, you can unlink the file and remove the temp directory.

xibz: Yea, that's a good point with the `NULL` character and special characters in general. I went ahead and went with the `ioutil.TempDir` approach since that seemed simplest. I also changed the configuration parameter of `FifoLogFile` to `FifoLogWriter` which now gives users the choice on how they want to handle the fifo.

cmd/firectl/main.go
```diff
@@ -188,23 +200,50 @@ func main() {
...
 
+	if len(opts.FcFifoLogFile) != 0 {
+		dir, err := ioutil.TempDir(os.TempDir(), "fcfifo")
+		if err != nil {
+			log.Fatalf("Fail to create temporary directory: %v", err)
+		}
+
+		defer os.RemoveAll(dir)
+
+		if len(opts.FcLogFifo) == 0 {
```
IRCody: Is checking == 0 better than != ""?

handlers.go
```diff
@@ -0,0 +1,247 @@
...
+// BootstrapLoggingHandler is a named handler that will set up fifo logging of
+// firecracker process.
+var BootstrapLoggingHandler = Handler{
+	Name: BootstrapLoggingHandlerName,
+	Fn: func(ctx context.Context, m *Machine) error {
+		m.logger.Debugf(fmt.Sprintf("%s handler executing", BootstrapLoggingHandlerName))
+		if err := m.setupLogging(ctx); err != nil {
+			m.logger.Warnf("setupLogging() returned %s. Continuing anyway.", err)
+		} else {
+			m.logger.Debugf("back from setupLogging")
```
IRCody: nit: I think this could be more descriptive, maybe "setup logging: success" or something since it means err == nil?

handlers.go
```diff
@@ -0,0 +1,247 @@
...
+// Handler represents a named handler that contains a name and a function which
+// is used to execute during the initialization process of a machine.
+type Handler struct {
+	Name string
+	Fn   func(context.Context, *Machine) error
+}
+
+// Run will execute each instruction in the handler list. If an error occurs in
+// any of the handlers, then the list will halt execution and return the error.
+func (l HandlerList) Run(ctx context.Context, m *Machine) error {
```
IRCody: nit: Can this HandlerList func not have the Handler type between it and the other HandlerList funcs?

handlers.go
```diff
@@ -0,0 +1,247 @@
...
+	for _, handler := range l.list {
+		if err := handler.Fn(ctx, m); err != nil {
```
IRCody: If these are always going to be run this way would it make sense to add the log message that says "executing handler x" here instead of inside the handlers for consistency if someone adds custom handlers?

handlers_test.go
```diff
@@ -0,0 +1,414 @@
...
+			Name: "foo",
+		},
+		Handler{
+			Name: "baz",
+		},
+	)
+
+	h.Remove("foo")
+
+	if e, a := 5, h.Len(); e != a {
```
IRCody: Is it worth also testing that the index we expected to be remove was the one removed? like
```go
if h[0] == "foo" {
        t.Errorf("expected foo to be removed but was not")
}
```

machine.go
```diff
@@ -424,15 +391,53 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
+	fifoPipe, err := os.OpenFile(fifoPath, os.O_RDONLY, 0600)
+	if err != nil {
+		return fmt.Errorf("Failed to open fifo path at %q: %v", fifoPath, err)
+	}
+
+	if err := syscall.Unlink(fifoPath); err != nil {
+		logger.Warnf("Failed to unlink %s", fifoPath)
+	}
+
+	logger.Printf("Capturing %q to writer", fifoPath)
```
IRCody: Should this be printf or have a log level with it(infof?)

cmd/firectl/main.go
xibz: Naw, just style

cmd/firectl/main.go
```diff
@@ -188,23 +200,50 @@ func main() {
...
+			opts.FcLogFifo = filepath.Join(dir, "fc_fifo")
+		}
+
+		if len(opts.FcMetricsFifo) == 0 {
+			opts.FcMetricsFifo = filepath.Join(dir, "fc_metrics_fifo")
+		}
+	}
+
+	var fifo *os.File
+	if len(opts.FcFifoLogFile) > 0 {
```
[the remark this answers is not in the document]
xibz: Good catch. Yea, will move that up

handlers.go
xibz: Ah, I think that got squashed when rebasing. Thanks! Will refix.

cmd/firectl/main.go
IRCody: :+1: I don't have a strong opinion on it either way was just curious if there was a benefit I wasn't aware of.

handlers.go
IRCody: The definition for type Handler is inserted between 2 HandlerList funcs (remove/Run). I think it would make sense to keep all the HandlerList functions continuous and have the Handler struct definition either above or below the HandlerList section.

xibz: Ah! Fixed :)

machine.go
```diff
@@ -424,15 +391,53 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
+	fifoPipe, err := os.OpenFile(fifoPath, os.O_RDONLY, 0600)
+	if err != nil {
+		return fmt.Errorf("Failed to open fifo path at %q: %v", fifoPath, err)
+	}
+
+	if err := syscall.Unlink(fifoPath); err != nil {
+		logger.Warnf("Failed to unlink %s", fifoPath)
+	}
+
+	logger.Infof("Capturing %q to writer", fifoPath)
```
nmeyerhans: Let's drop this to Debug severity. We'll do a more complete log cleanup later, but for now let's get new messages in the right place to begin with.

machine.go
```diff
@@ -424,15 +391,53 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
-	if err == nil {
-		m.logger.Printf("Configured VMM logging to %s, metrics to %s: %s",
-			m.cfg.LogFifo, m.cfg.MetricsFifo, resp.Error())
+	if err != nil {
+		return err
 	}
-	return err
+
+	m.logger.Printf("Configured VMM logging to %s, metrics to %s: %s",
+		m.cfg.LogFifo, m.cfg.MetricsFifo, resp.Error())
```
nmeyerhans: Let's drop this to Debug severity, and also drop `resp.Error()`, since it won't contain anything meaningful.

cmd/firectl/main.go
```diff
@@ -188,23 +196,49 @@ func main() {
...
 	}
 
+	var fifo *os.File
+	if len(opts.FcFifoLogFile) != 0 {
+		dir, err := ioutil.TempDir(os.TempDir(), "fcfifo")
+		if err != nil {
+			log.Fatalf("Fail to create temporary directory: %v", err)
+		}
+
+		defer os.RemoveAll(dir)
```
nmeyerhans: FIFO management still feels a bit off. Combinations of certain firectl flags result in weird behavior. For example:

(most flags left out, only log/fifo related flags are shown)


```
$ firectl --firecracker-log=/tmp/y
...
INFO[0000] Configured VMM logging to /tmp/fcfifo558137878/fc_fifo, metrics to /tmp/fcfifo558137878/fc_metrics_fifo: [PUT /logger][204] putLoggerNoContent  
```
The metrics fifo sticks around for the lifetime of the process. Since I never asked for a metrics FIFO, this is surprising to me.

```
$ firectl --firecracker-log=/tmp/log --vmm-log-fifo=/tmp/fc.log --metrics-fifo=/tmp/metrics
...
ERRO[0000] Unable to set up logging: file exists        
WARN[0000] setupLogging() returned file exists. Continuing anyway. 
```
I asked for all 3, and got an unhelpful error.
```
$ firectl --metrics-fifo=/tmp/metrics
...
INFO[0000] VMM logging and metrics disabled.            
```
I asked for only one FIFO, but got nothing at all. Since firecracker wants both FIFOs configured, I think we should create the unspecified one in a tempdir and immediately unlink it. (I realize that this is behavior predates this PR and is largely driven by Firecracker's behavior, but we should handle it better.)

xibz: > The metrics fifo sticks around for the lifetime of the process. Since I never asked for a metrics FIFO, this is surprising to me.

I thought this was the simplest approach, since it does delete it, just until the process ends. Otherwise, we wouldn't be able to delete until we've opened the file. However, this does open the question for a potential change in the design. For instance, if we did something similar to the `FifoLogWriter` being an `io.Writer` with the log and metric fifo, we'd be able to handle that at the `firectl` level. I also think this is the right approach. What do you think?

> I asked for all 3, and got an unhelpful error.

Fixed and was a issue with how interfaces deal with `nil` values.

> I asked for only one FIFO, but got nothing at all. Since firecracker wants both FIFOs configured, I think we should create the unspecified one in a tempdir and immediately unlink it. (I realize that this is behavior predates this PR and is largely driven by Firecracker's behavior, but we should handle it better.)

Fair enough. I'll update. If we go with the design above, I can reuse some of the logic in `firectl`

cmd/firectl/main.go
```diff
@@ -105,6 +109,74 @@ func parseVsocks(devices []string) ([]firecracker.VsockDevice, error) {
...
+		if generateFifoFilename {
+			opts.FcLogFifo = filepath.Join(dir, "fc_fifo")
+		}
+
+		if generateMetricFifoFilename {
+			opts.FcMetricsFifo = filepath.Join(dir, "fc_metrics_fifo")
+		}
+	}
+
+	return fifo, cleanupFns, nil
```
mxpv: I think this can be simplified:
```go
func handleFifos(opts options) (io.Writer, []func() error, error) {
	var (
		generateFifoFilename       = len(opts.FcLogFifo) == 0
		generateMetricFifoFilename = len(opts.FcMetricsFifo) == 0
		cleanupFns                 = []func() error{}
		err                        error
		fifo                       io.WriteCloser
	)

	if len(opts.FcFifoLogFile) > 0 {
		if fifo, err = createFifoFileLogs(opts.FcFifoLogFile); err != nil {
			return fifo, cleanupFns, fmt.Errorf("Failed to create fifo log file: %v", err)
		}

		cleanupFns = append(cleanupFns, func() error {
			return fifo.Close()
		})
	}

	if generateFifoFilename || generateMetricFifoFilename {
		// ...
	}

	return fifo, cleanupFns, nil
}

```

machine.go
```diff
@@ -215,116 +203,96 @@ func NewMachine(cfg Config, opts ...Opt) (*Machine, error) {
...
 
-	if err := m.setupLogging(ctx); err != nil {
-		m.logger.Warnf("setupLogging() returned %s. Continuing anyway.", err)
-	} else {
-		m.logger.Debugf("back from setupLogging")
+// Wait will wait until the firecracker process has finished
+func (m *Machine) Wait(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return nil
```
mxpv: `return ctx.Err()`

cmd/firectl/main.go
```diff
@@ -252,25 +342,19 @@ func main() {
...
 	}
+	defer m.StopVMM()
 
 	// wait for the VMM to exit
-	if err := <-errCh; err != nil {
+	if err := m.Wait(vmmCtx); err != nil {
 		log.Fatalf("startVMM returned error %s", err)
 	}
 	log.Printf("startVMM was happy")
+
```
mxpv: Logs a bit outdated here. s/startVMM/Wait ?

cmd/firectl/main.go
mxpv: According to your logic, if `opts.FcLogFifo` not empty, `generateFifoFilename` will be true, no matter which branch it'll go, you do same checks in both branches. Do I missing something?

mxpv: I think I get it :)

```go
if len(opts.FcLogFifo) > 0 || len(opts.FcMetricsFifo) > 0 || len(opts.FcFifoLogFile) > 0 {
                generateFifoFilename = len(opts.FcLogFifo) == 0
		generateMetricFifoFilename = len(opts.FcMetricsFifo) == 0
}

if len(opts.FcFifoLogFile) > 0 {
               createFifoFileLogs()
               ...
}

```

machine.go
```diff
@@ -215,116 +203,96 @@ func NewMachine(cfg Config, opts ...Opt) (*Machine, error) {
...
+	return nil
+}
 
-	for id, dev := range m.cfg.AdditionalDrives {
-		// id must be increased by 2 because firecracker uses 1-indexed arrays and the root drive occupies position 1.
-		err = m.attachDrive(ctx, dev, id+2, false)
-		if err != nil {
+func (m *Machine) attachDrives(ctx context.Context, rootIndex int, drives ...BlockDevice) error {
+	for id, dev := range drives {
+		if err := m.attachDrive(ctx, dev, id+1, id == rootIndex); err != nil {
```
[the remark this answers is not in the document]
xibz: Yea, that sounds good, +1

cmd/firectl/main.go
```diff
@@ -43,28 +47,35 @@ const (
...
 		}
-		e := firecracker.BlockDevice{
-			HostPath: path,
-			Mode:     "rw",
+
+		e := models.Drive{
+			// i + 2 represents the drive ID. We will reserve 1 for root.
+			DriveID:      firecracker.String(strconv.Itoa(i + 2)),
+			PathOnHost:   firecracker.String(path),
+			IsReadOnly:   firecracker.Bool(false),
```
nmeyerhans: The value here should be set based on the `:rw` or `:ro` suffix on the drive path.

cmd/firectl/main.go
```diff
@@ -105,6 +116,74 @@ func parseVsocks(devices []string) ([]firecracker.VsockDevice, error) {
...
+	generateMetricFifoFilename := false
+	cleanupFns := []func() error{}
+	var err error
+
+	var fifo io.WriteCloser
+	if len(opts.FcFifoLogFile) > 0 {
+		// if a fifo log file was specified via the CLI then we need to check if
+		// the log or metric fifo was also specified. If not, we will then generate
+		// those fifo files.
+		if len(opts.FcLogFifo) == 0 {
```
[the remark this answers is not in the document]
xibz: Good call out. Fixed.

cmd/firectl/main.go
xibz: I am going to leave the function as is since I think it is easier to read.

cmd/firectl/main.go
```diff
@@ -188,23 +261,40 @@ func main() {
 		log.Fatalf("Invalid vsock specification: %s", err)
 	}
 
+	fifo, cleanFns, err := handleFifos(opts)
+	// we call cleanup first due to errors returning at different points which
+	// may result in a file handle being opened.
+	defer func() {
+		for _, fn := range cleanFns {
+			if err := fn(); err != nil {
+				log.Printf("Failed to cleanup: %v", err)
```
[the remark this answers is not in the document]
xibz: What do you mean? Like `os.Exit(1)`?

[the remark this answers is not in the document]
mxpv: No, `log.Error(...)` or `log.WithError(err).Error(...)`

PR #28, opened 2018-12-05
test: Better error message for missing vsock perms

### LINKED ISSUE #25: `make test` fails with "Cannot create vsock device. VhostOpen(VhostOpen(Os { code: 13, kind: PermissionDenied, message: Permission denied }))"
```
$ make test
⋮
DEBU[0002] PUT /actions HTTP/1.1
Host: localhost
User-Agent: Go-http-client/1.1
Content-Length: 32
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip

{"action_type":"InstanceStart"}
 
DEBU[0002] HTTP/1.1 500 Internal Server Error
Transfer-Encoding: chunked
Content-Type: application/json
Date: Wed, 05 Dec 2018 21:20:51 GMT

8e
{
  "fault_message": "Cannot create vsock device. VhostOpen(VhostOpen(Os { code: 13, kind: PermissionDenied, message: Permission denied }))"
}
0

 
ERRO[0002] Starting instance: [PUT /actions][500] createSyncAction default  &{FaultMessage:Cannot create vsock device. VhostOpen(VhostOpen(Os { code: 13, kind: PermissionDenied, message: Permission denied }))} 
WARN[0007] firecracker exited: signal: terminated       
⋮
```

This needs a better error message showing how to fix it.

samuelkarp:
*Issue #, if available:*

Fixes #25

*Description of changes:*

Better error message

### DIFF
--- a/machine_test.go
+++ b/machine_test.go
@@ -21,9 +21,11 @@ import (
 	"flag"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/firecracker-microvm/firecracker-go-sdk/client/operations"
 	"github.com/firecracker-microvm/firecracker-go-sdk/fctesting"
 	"github.com/golang/mock/gomock"
 )
@@ -252,7 +254,15 @@ func testAttachVsock(ctx context.Context, t *testing.T, m *Machine) {
 func testStartInstance(ctx context.Context, t *testing.T, m *Machine) {
 	err := m.startInstance(ctx)
 	if err != nil {
-		t.Errorf("startInstance failed: %s", err)
+		if syncErr, ok := err.(*operations.CreateSyncActionDefault); ok &&
+			strings.HasPrefix(syncErr.Payload.FaultMessage, "Cannot create vsock device") {
+				t.Errorf(`startInstance: %s
+Do you have permission to interact with /dev/vhost-vsock?
+Grant yourself permission with `+"`sudo setfacl -m u:${USER}:rw /dev/vhost-vsock`", syncErr.Payload.FaultMessage)
+			}
+		} else {
+			t.Errorf("startInstance failed: %s", err)
+		}
 	}
 }
 


machine_test.go
```diff
@@ -252,7 +254,15 @@ func testAttachVsock(ctx context.Context, t *testing.T, m *Machine) {
 func testStartInstance(ctx context.Context, t *testing.T, m *Machine) {
 	err := m.startInstance(ctx)
 	if err != nil {
-		t.Errorf("startInstance failed: %s", err)
+		if syncErr, ok := err.(*operations.CreateSyncActionDefault); ok {
+			if strings.HasPrefix(syncErr.Payload.FaultMessage, "Cannot create vsock device") {
+				t.Errorf(`startInstance: %s
+Do you have permission to interact with /dev/vhost-vsock?
```
nmeyerhans: Given that vsock is disabled in default Firecracker builds, do you think it also makes sense to specifically handle the case where we're running with a build that doesn't support it?

samuelkarp: Yeah, that's fair.  What would you suggest for that?

samuelkarp: "Does your installation of Firecracker have support for vsock?  Enable it by building with `cargo build --features vsock`?"

samuelkarp: Different error, will handle in a different PR:

```
    --- FAIL: TestMicroVMExecution/TestAttachVsock (0.00s)
        machine_test.go:250: attaching vsock failed: [PUT /vsocks/{id}][400] putGuestVsockByIdBadRequest  &{FaultMessage:Invalid request method and/or path: PUT /vsocks/foo}
```

samuelkarp: Opened https://github.com/firecracker-microvm/firecracker-go-sdk/issues/29 for tracking.

nmeyerhans:
LGTM

machine_test.go
```diff
@@ -252,7 +254,15 @@ func testAttachVsock(ctx context.Context, t *testing.T, m *Machine) {
 func testStartInstance(ctx context.Context, t *testing.T, m *Machine) {
 	err := m.startInstance(ctx)
 	if err != nil {
-		t.Errorf("startInstance failed: %s", err)
+		if syncErr, ok := err.(*operations.CreateSyncActionDefault); ok &&
+			strings.HasPrefix(syncErr.Payload.FaultMessage, "Cannot create vsock device") {
```
xibz: This seems a little fragile since this message is coming back from `firecracker` and not the SDK. I am a little worried that `firecracker` may update this message which would cause the SDK to start failing. Even if we fixed it, users using the old SDK but newer `firecracker` would have failed builds. I would much prefer trying to validate this in the SDK or maybe wrapping the error received from `firecracker`? Don't know if I like the wrapping idea either

samuelkarp: Yeah, it's definitely fragile.  Unfortunately the errors in the Firecracker API are not modeled any more than just a string.  I opted to put this in the test instead of the SDK itself precisely because I can't guarantee that they won't change; I don't want to guarantee that the SDK functions if I'm relying on strings.  I want to fix it here so that anyone trying to build the SDK can validate what they're seeing, and then propose that we add explicit error modeling to the Firecracker API.

xibz: Alright, that sounds like a plan! :shipit:

samuelkarp: This is the current API model:

https://github.com/firecracker-microvm/firecracker-go-sdk/blob/ec0a04065cf1307665ea30a1ae82b9a14da54401/client/swagger.yaml#L396-L400

PR #36, opened 2018-12-08
godoc: add new and enhance existing documentation

samuelkarp:
godoc: add new and enhance existing documentation

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,29 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+/*
+Package firecracker provides a library to interact with the Firecracker API.
+
+Firecracker is an open-source virtualization technology that is purpose-built
+for creating and managing secure, multi-tenant containers and functions-based
+services.  See https://firecracker-microvm.github.io/ for more details.
+
+This library requires Go 1.11 and can be used with Go modules.
+
+BUG(aws): There are some Firecracker features that are not yet supported by the
+SDK.  These are tracked as GitHub issues with the firecracker-feature label:
+https://github.com/firecracker-microvm/firecracker-go-sdk/issues?q=is%3Aissue+is%3Aopen+label%3Afirecracker-feature
+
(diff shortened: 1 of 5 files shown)


machine.go
```diff
@@ -115,32 +162,41 @@ type Machine struct {
 
 // Logger returns a logrus logger appropriate for logging hypervisor messages
 func (m *Machine) Logger() *log.Entry {
-	// return m.logger.WithField("subsystem", "firecracker-go-sdk")
-	return m.logger
+	return m.logger.WithField("subsystem", "firecracker-go-sdk")
```
xibz: nit: Can you the const `userAgent` here?

machine.go
```diff
@@ -90,21 +137,21 @@ func (cfg *Config) Validate() error {
 	}
 
 	if _, err := os.Stat(cfg.KernelImagePath); err != nil {
-		return fmt.Errorf("Failed to stat kernal image path, %q: %v", cfg.KernelImagePath, err)
+		return fmt.Errorf("failed to stat kernal image path, %q: %v", cfg.KernelImagePath, err)
 	}
 	if _, err := os.Stat(cfg.RootDrive.HostPath); err != nil {
-		return fmt.Errorf("Failed to stat host path, %q: %v", cfg.RootDrive.HostPath, err)
+		return fmt.Errorf("failed to stat host path, %q: %v", cfg.RootDrive.HostPath, err)
```
mxpv: Can we use `errors.Wrap` instead?

doc.go
```diff
@@ -0,0 +1,29 @@
...
+
+Firecracker is an open-source virtualization technology that is purpose-built
+for creating and managing secure, multi-tenant containers and functions-based
+services.  See https://firecracker-microvm.github.io/ for more details.
+
+This library requires Go 1.11 and can be used with Go modules.
+
+BUG(aws): There are some Firecracker features that are not yet supported by the
+SDK.  These are tracked as GitHub issues with the firecracker-feature label:
+https://github.com/firecracker-microvm/firecracker-go-sdk/issues?q=is%3Aissue+is%3Aopen+label%3Afirecracker-feature.
```
mxpv: Dot at the end makes link invalid. Wouldn't it be more convenient to have this note in README ?

machine.go
mxpv: Major benefit - it keeps call stack info, which sometimes makes life of lot easier.
It's not super important for this specific case, but in general it's better to have this information than not to.

samuelkarp: I'd prefer to leave it out for now, both because it's an extra dependency that I don't want us to take just yet and because it's out of scope for this PR (I changed one character in that line...).

doc.go
samuelkarp: I'll fix the dot.  It's already in the README; I copied it from there.

machine.go
samuelkarp: Sure

doc.go
samuelkarp: Done

machine.go
samuelkarp: Done

Issue #37, opened 2018-12-10
tests: panic when firecraker binary is not found


samuelkarp:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x490a0e]

goroutine 13 [running]:
testing.tRunner.func1(0xc00025e300)
	/home/skarp/.gobin/go/src/testing/testing.go:792 +0x387
panic(0x8779e0, 0xc9dbb0)
	/home/skarp/.gobin/go/src/runtime/panic.go:513 +0x1b9
os.(*Process).signal(0x0, 0x9758a0, 0x96df98, 0x0, 0x0)
	/home/skarp/.gobin/go/src/os/exec_unix.go:56 +0x2e
os.(*Process).Signal(0x0, 0x9758a0, 0x96df98, 0xc0004be480, 0x4f542d)
	/home/skarp/.gobin/go/src/os/exec.go:131 +0x3f
github.com/firecracker-microvm/firecracker-go-sdk.(*Machine).stopVMM(0xc000143800, 0x795601, 0xc00013c450)
	/home/skarp/projects/firecracker-go-sdk/machine.go:383 +0xa8
github.com/firecracker-microvm/firecracker-go-sdk.(*Machine).StopVMM(0xc000143800, 0xf, 0xc63015)
	/home/skarp/projects/firecracker-go-sdk/machine.go:377 +0x2b
github.com/firecracker-microvm/firecracker-go-sdk.testStopVMM(0x976a00, 0xc000028118, 0xc00025e300, 0xc000143800)
	/home/skarp/projects/firecracker-go-sdk/machine_test.go:295 +0x2f
github.com/firecracker-microvm/firecracker-go-sdk.TestMicroVMExecution.func12(0xc00025e300)
	/home/skarp/projects/firecracker-go-sdk/machine_test.go:143 +0x46
testing.tRunner(0xc00025e300, 0xc00000c640)
	/home/skarp/.gobin/go/src/testing/testing.go:827 +0xbf
created by testing.(*T).Run
	/home/skarp/.gobin/go/src/testing/testing.go:878 +0x353
exit status 2
```

PR #38, opened 2018-12-10
machine: StopVMM should not panic on nil Process 

samuelkarp:
*Issue #, if available:*
Fixes https://github.com/firecracker-microvm/firecracker-go-sdk/issues/37

*Description of changes:*
* Check for nil Process before calling Signal
* Use a different channel for signaling the end of TestMicroVMExecution when startVMM fails
* Skip waiting for stabilization if the VMM isn't running
* Update HACKING.md

### DIFF
--- a/machine_test.go
+++ b/machine_test.go
@@ -118,13 +118,15 @@ func TestMicroVMExecution(t *testing.T) {
 
 	vmmCtx, vmmCancel := context.WithTimeout(ctx, 30*time.Second)
 	defer vmmCancel()
-	var exitchannel <-chan error
+	exitchannel := make(chan error)
 	go func() {
-		var err error
-		exitchannel, err = m.startVMM(vmmCtx)
+		exitCh, err := m.startVMM(vmmCtx)
 		if err != nil {
+			close(exitchannel)
 			t.Fatalf("Failed to start VMM: %v", err)
 		}
+		exitchannel <- <-exitCh
+		close(exitchannel)
 	}()
 	time.Sleep(2 * time.Second)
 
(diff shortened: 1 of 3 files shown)


machine_test.go
```diff
@@ -118,13 +118,15 @@ func TestMicroVMExecution(t *testing.T) {
...
+	exitchannel := make(chan error)
 	go func() {
-		var err error
-		exitchannel, err = m.startVMM(vmmCtx)
+		exitCh, err := m.startVMM(vmmCtx)
 		if err != nil {
+			close(exitchannel)
 			t.Fatalf("Failed to start VMM: %v", err)
 		}
+		exitchannel <- <-exitCh
```
mxpv: Can you give some context here? Why do we need two channels?

mxpv: I'm somewhat confused with this block of code :)

Is this what we want?
```go
vmmCtx, vmmCancel := context.WithTimeout()
...
go func() {

    exitCh, err := m.startVMM(vmmCtx)
    if err != nil {
        t.Fatalf("Failed to start VMM: %v", err)
    } else {
         select {
         case <-vmm.Ctx.Done():
              t.Fatal();
         case <-exitCh:
              OK
        }
    }

}


...

t.Run("TestStartInstance", func(t *testing.T) { testStartInstance(vmmCtx, t, m) })

// We can call this safely, if exit==nil, Fatal is called.
<-exitCh

```

Am I missing something?

samuelkarp: `t.Fatal` only stops execution of the goroutine, not of the whole test, so that will block forever.

mxpv: Oh. I didn't know what :(

PR #42, opened 2018-12-21
adding support for jailer

### LINKED ISSUE #11: Support the firecracker jailer
Firecracker provides a [jailer](https://github.com/firecracker-microvm/firecracker/blob/master/docs/jailer.md) that enhances the security boundary of the Firecracker process. We don't currently support it at all, but should.

xibz:
Adds support for firecracker's jailer. This allows for more secure virtualization through creation of namespaces and cgroups. Jailer can be turned on by setting the `EnableJailer` flag. This also implements a very naive mapper and issue #56 is to implement a smarter mapper.

`go test` requires sudo since jailer requires sudo...

Fixes #11

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,17 @@ steps:
       - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
       - 'ln -s /usr/local/bin/firecracker testdata/firecracker'
       - 'ln -s /usr/local/bin/jailer testdata/jailer'
-      - "make test EXTRAGOARGS='-v -count=1'"
+      - 'sudo ip tuntap add fc-test-tap${BUILDKITE_BUILD_NUMBER} mode tap user $(sudo id -u buildkite-agent)'
+      - "DISABLE_ROOT_TESTS=true FC_TEST_TAP=fc-test-tap${BUILDKITE_BUILD_NUMBER} make test EXTRAGOARGS='-v -count=1'"
+      - 'sudo ip tuntap del fc-test-tap${BUILDKITE_BUILD_NUMBER} mode tap'
+
+  - label: ':hammer: root tests'
+    commands:
+      - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
+      - 'ln -s /usr/local/bin/firecracker testdata/firecracker'
+      - 'ln -s /usr/local/bin/jailer testdata/jailer'
+      - 'sudo ip tuntap add fc-root-tap${BUILDKITE_BUILD_NUMBER} mode tap user $(sudo id -u buildkite-agent)'
+      - "sudo FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} make test EXTRAGOARGS='-v -count=1'"
+      - 'sudo ip tuntap del fc-root-tap${BUILDKITE_BUILD_NUMBER} mode tap'
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
--- a/HACKING.md
+++ b/HACKING.md
@@ -6,14 +6,17 @@ Testing
 
 Tests are written using Go's testing framework and can be run with the standard
 `go test` tool.  If you prefer to use the Makefile, `make test EXTRAGOARGS=-v`
-will run tests in verbose mode.
+will run tests in verbose mode. By default, the unit tests require root
+privileged access. This can be disabled by setting the `DISABLE_ROOT_TESTS`
+environment variable.
 
 You need some external resources in order to run the tests, as described below:
 
 1. A firecracker and jailer binary (tested with 0.12.0) must either be
    installed as `./testdata/firecracker` or the path must be specified through
    the `FC_TEST_BIN` environment variable. The jailer requires go test to be
-   run with sudo and can also be turned off with the `test.root-disable` flag.
+   run with sudo and can also be turned off by setting the `DISABLE_ROOT_TESTS`
+   env flag.
 2. Access to hardware virtualization via `/dev/kvm` and `/dev/vhost-vsock`
    (ensure you have mode `+rw`!)
 3. An uncompressed Linux kernel binary that can boot in Firecracker VM (Must be
--- a/HACKING.md
+++ b/HACKING.md
@@ -10,9 +10,10 @@ will run tests in verbose mode.
 
 You need some external resources in order to run the tests, as described below:
 
-1. A firecracker binary (tested with 0.10.1), but any recent version should
-   work.  Must either be installed as `./testdata/firecracker` or the path must
-   be specified through the `FC_TEST_BIN` environment variable.
+1. A firecracker and jailer binary (tested with 0.12.0) must either be
+   installed as `./testdata/firecracker` or the path must be specified through
+   the `FC_TEST_BIN` environment variable. The jailer requires go test to be
+   run with sudo and can also be turned off with the `test.root-disable` flag.
 2. Access to hardware virtualization via `/dev/kvm` and `/dev/vhost-vsock`
    (ensure you have mode `+rw`!)
 3. An uncompressed Linux kernel binary that can boot in Firecracker VM (Must be
--- a/example_test.go
+++ b/example_test.go
@@ -220,7 +220,7 @@ func ExampleNetworkInterface_rateLimiting() {
 	}
 }
 
-func ExampleJailer_withBuilder() {
+func ExampleJailerCommandBuilder() {
 	ctx := context.Background()
 	// Creates a jailer command using the JailerCommandBuilder.
 	b := firecracker.NewJailerCommandBuilder().
@@ -264,6 +264,7 @@ func ExampleJailer_withBuilder() {
 		panic(err)
 	}
 
-	tCtx, _ := context.WithTimeout(ctx, time.Minute)
+	tCtx, cancelFn := context.WithTimeout(ctx, time.Minute)
+	defer cancelFn()
 	m.Wait(tCtx)
 }
--- a/example_test.go
+++ b/example_test.go
@@ -99,11 +99,7 @@ func ExampleDrivesBuilder() {
 	}
 
 	ctx := context.Background()
-	m, err := firecracker.NewMachine(ctx, cfg)
-	if err != nil {
-		panic(fmt.Errorf("failed to create new machine: %v", err))
-	}
-
+	m := firecracker.NewMachine(ctx, cfg)
 	if err := m.Start(ctx); err != nil {
 		panic(fmt.Errorf("failed to initialize machine: %v", err))
 	}
@@ -141,11 +137,7 @@ func ExampleDrivesBuilder_driveOpt() {
 	}
 
 	ctx := context.Background()
-	m, err := firecracker.NewMachine(ctx, cfg)
-	if err != nil {
-		panic(fmt.Errorf("failed to create new machine: %v", err))
-	}
-
+	m := firecracker.NewMachine(ctx, cfg)
 	if err := m.Start(ctx); err != nil {
 		panic(fmt.Errorf("failed to initialize machine: %v", err))
 	}
@@ -249,6 +241,9 @@ func ExampleJailer_withBuilder() {
 
 	// Passes the custom jailer command into the constructor
 	m := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(b.Build(ctx)))
+	// This does not copy any of the files over to the rootfs since a process
+	// runner was specified. This examples assumes that the files have been
+	// properly mounted.
 	if err := m.Start(ctx); err != nil {
 		panic(err)
 	}
--- a/example_test.go
+++ b/example_test.go
@@ -45,11 +45,7 @@ func ExampleWithProcessRunner_logging() {
 		WithStderr(stderr).
 		Build(ctx)
 
-	m, err := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(cmd))
-	if err != nil {
-		panic(fmt.Errorf("failed to create new machine: %v", err))
-	}
-
+	m := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(cmd))
 	defer os.Remove(cfg.SocketPath)
 
 	if err := m.Start(ctx); err != nil {
@@ -211,4 +207,45 @@ func ExampleNetworkInterface_rateLimiting() {
 	if err := m.Wait(ctx); err != nil {
 		panic(err)
 	}
+}
+
+func ExampleJailer_withBuilder() {
+	ctx := context.Background()
+	// Creates a jailer command using the JailerCommandBuilder.
+	b := firecracker.JailerCommandBuilder{}.
+		WithID("my-test-id").
+		WithUID(123).
+		WithGID(100).
+		WithNumaNode(0).
+		WithExecFile("/usr/local/bin/firecracker").
+		WithChrootBaseDir("/tmp").
+		WithStdout(os.Stdout).
+		WithStderr(os.Stderr)
+
+	const socketPath = "/tmp/firecracker/my-test-id/api.socket"
+	cfg := firecracker.Config{
+		SocketPath:      socketPath,
+		KernelImagePath: "./vmlinux",
+		Drives: []models.Drive{
+			models.Drive{
+				DriveID:      firecracker.String("1"),
+				IsRootDevice: firecracker.Bool(true),
+				IsReadOnly:   firecracker.Bool(false),
+				PathOnHost:   firecracker.String("/path/to/root/drive"),
+			},
+		},
+		MachineCfg: models.MachineConfiguration{
+			VcpuCount: 1,
+		},
+		DisableValidation: true,
+	}
+
+	// Passes the custom jailer command into the constructor
+	m := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(b.Build(ctx)))
+	if err := m.Start(ctx); err != nil {
+		panic(err)
+	}
+
+	tCtx, _ := context.WithTimeout(ctx, time.Minute)
+	m.Wait(tCtx)
 }
--- a/example_test.go
+++ b/example_test.go
@@ -52,7 +52,11 @@ func ExampleWithProcessRunner_logging() {
 		WithStderr(stderr).
 		Build(ctx)
 
-	m := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(cmd))
+	m, err := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(cmd))
+	if err != nil {
+		panic(fmt.Errorf("failed to create new machine: %v", err))
+	}
+
 	defer os.Remove(cfg.SocketPath)
 
 	if err := m.Start(ctx); err != nil {
@@ -99,7 +103,11 @@ func ExampleDrivesBuilder() {
 	}
 
 	ctx := context.Background()
-	m := firecracker.NewMachine(ctx, cfg)
+	m, err := firecracker.NewMachine(ctx, cfg)
+	if err != nil {
+		panic(fmt.Errorf("failed to create new machine: %v", err))
+	}
+
 	if err := m.Start(ctx); err != nil {
 		panic(fmt.Errorf("failed to initialize machine: %v", err))
 	}
@@ -137,7 +145,11 @@ func ExampleDrivesBuilder_driveOpt() {
 	}
 
 	ctx := context.Background()
-	m := firecracker.NewMachine(ctx, cfg)
+	m, err := firecracker.NewMachine(ctx, cfg)
+	if err != nil {
+		panic(fmt.Errorf("failed to create new machine: %v", err))
+	}
+
 	if err := m.Start(ctx); err != nil {
 		panic(fmt.Errorf("failed to initialize machine: %v", err))
 	}
@@ -211,7 +223,7 @@ func ExampleNetworkInterface_rateLimiting() {
 func ExampleJailer_withBuilder() {
 	ctx := context.Background()
 	// Creates a jailer command using the JailerCommandBuilder.
-	b := firecracker.JailerCommandBuilder{}.
+	b := firecracker.NewJailerCommandBuilder().
 		WithID("my-test-id").
 		WithUID(123).
 		WithGID(100).
@@ -240,7 +252,11 @@ func ExampleJailer_withBuilder() {
 	}
 
 	// Passes the custom jailer command into the constructor
-	m := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(b.Build(ctx)))
+	m, err := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(b.Build(ctx)))
+	if err != nil {
+		panic(fmt.Errorf("failed to create new machine: %v", err))
+	}
+
 	// This does not copy any of the files over to the rootfs since a process
 	// runner was specified. This examples assumes that the files have been
 	// properly mounted.
--- a/example_test.go
+++ b/example_test.go
@@ -19,6 +19,13 @@ func ExampleWithProcessRunner_logging() {
 		MachineCfg: models.MachineConfiguration{
 			VcpuCount: 1,
 		},
+		JailerCfg: firecracker.JailerConfig{
+			GID:      firecracker.Int(100),
+			UID:      firecracker.Int(100),
+			ID:       "my-micro-vm",
+			NumaNode: firecracker.Int(0),
+			ExecFile: "/path/to/firecracker",
+		},
 	}
 
 	// stdout will be directed to this file
--- a/fctesting/utils.go
+++ b/fctesting/utils.go
@@ -1,0 +1,41 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package fctesting
+
+import (
+	"flag"
+	"os"
+	"testing"
+)
+
+var rootDisabled bool
+
+func init() {
+	flag.BoolVar(&rootDisabled, "test.root-disable", false, "Disables tests that require root")
+}
+
+// RequiresRoot will ensure that tests that require root access are actually
+// root. In addition, this will skip root tests if the test.root-disable is set
+// to true
+func RequiresRoot(t testing.TB) {
+	if rootDisabled {
+		t.Skip("skipping test that requires root")
+	}
+
+	if e, a := 0, os.Getuid(); e != a {
+		t.Fatal("This test must be run as root. " +
+			"To disable tests that require root, " +
+			"run the tests with the -test.root-disable flag.")
+	}
+}
--- a/fctesting/utils.go
+++ b/fctesting/utils.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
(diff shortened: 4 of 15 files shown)


jailer.go
```diff
@@ -0,0 +1,376 @@
...
+
+	return cmd
+}
+
+// Jail will set up proper handlers and remove configuuration validation due to
+// stating of files
+func jail(ctx context.Context, m *Machine, cfg Config) {
+	m.Handlers.Validation = m.Handlers.Validation.Remove(ValidateCfgHandlerName)
+
+	id := StringValue(cfg.JailerCfg.ID)
```
IRCody: Any chance this is unset resulting in id = ""?

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+	b := JailerCommandBuilder{}.
+		WithID(id).
+		WithUID(IntValue(cfg.JailerCfg.UID)).
+		WithGID(IntValue(cfg.JailerCfg.GID)).
+		WithNumaNode(IntValue(cfg.JailerCfg.NumaNode)).
+		WithExecFile(StringValue(cfg.JailerCfg.ExecFile)).
+		WithDaemonize(cfg.JailerCfg.Daemonize).
+		WithSecCompLevel(cfg.JailerCfg.SecCompLevel).
+		WithStdout(os.Stdout).
+		WithStderr(os.Stderr)
```
IRCody: Are these all safe to use with the zero value for the type?

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+
+	return args
+}
+
+// ID will return the command arguments regarding the id.
+func (b JailerCommandBuilder) ID() []string {
+	return []string{
+		"--id",
+		b.id,
+	}
```
mxpv: Are there any use cases when user wants to read back parts of command line? Or these can be private?

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+
+// Args returns the specified set of args to be used
+// in command construction.
+func (b JailerCommandBuilder) Args() []string {
+	args := []string{}
+	args = append(args, b.ID()...)
+	args = append(args, b.UID()...)
+	args = append(args, b.GID()...)
+	args = append(args, b.ExecFile()...)
+	args = append(args, b.NumaNode()...)
```
mxpv: What if user forgets to provide something? Will this make incorrect command line?

example_test.go
```diff
@@ -67,3 +64,44 @@ func ExampleWithProcessRunner_logging() {
 		panic(err)
 	}
 }
+
+func ExampleJailer_withBuilder() {
```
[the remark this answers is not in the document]
xibz: Yea, goal here was how configurable the jailer builder is.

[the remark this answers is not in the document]
samuelkarp: It'd be good to have an example that is the simplest working way as well (unless that's already covered by another example because the jailer is used by default?).

machine.go
```diff
@@ -191,23 +194,23 @@ func (m Machine) LogLevel() string {
...
-		WithSocketPath(cfg.SocketPath).
-		Build(ctx)
 	m.Handlers = defaultHandlers
+	m.logger = log.NewEntry(logger)
+	if cfg.DisableJailer {
+		m.cmd = defaultFirecrackerVMMCommandBuilder.
+			WithSocketPath(m.cfg.SocketPath).
+			Build(ctx)
+	} else {
+		jail(ctx, m, cfg)
```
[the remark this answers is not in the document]
xibz: Yea, good call out.

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+	// SecCompLevelBasic prohibits syscalls not whitelisted by Firecracker.
+	SecCompLevelBasic = SecCompLevelValue(1)
+	// SecCompLevelAdvanced adds further checks on some of the parameters of the
+	// allowed syscalls.
+	SecCompLevelAdvanced = SecCompLevelValue(2)
+)
+
+// JailerConfig is jailer specific configuration needed to execute the jailer.
+type JailerConfig struct {
+	GID           *int
```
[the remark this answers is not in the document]
xibz: Yea, jailer let's you run as root. You can set everything to 0

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+	UID           *int
+	ID            *string
```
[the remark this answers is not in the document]
xibz: Yes, `""` is valid.

jailer.go
xibz: Validation handles this. Only way this would fail is if the user disables or removes the validation.

jailer.go
xibz: I think if users want to log individual components of the builder for debugging purposes may be nice

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+func (b JailerCommandBuilder) WithStdin(stdin io.Reader) JailerCommandBuilder {
+	b.stdin = stdin
+	return b
+}
+
+// Build will build a jailer command.
+func (b JailerCommandBuilder) Build(ctx context.Context) *exec.Cmd {
+	cmd := exec.CommandContext(
+		ctx,
+		"jailer",
```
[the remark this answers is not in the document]
xibz: Hm, should it? I definitely can see the `firecracker` binary being configurable, but less for this. What do you think?

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+
+	if stderr := b.Stderr(); stderr != nil {
+		cmd.Stderr = stderr
+	}
+
+	return cmd
+}
+
+// Jail will set up proper handlers and remove configuuration validation due to
+// stating of files
```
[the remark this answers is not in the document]
xibz: Ah, that's a good point. Will definitely put a comment about that.

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+				return err
+			}
+
+			var rootDrive models.Drive
+			rootDriveIndex := 0
+			for i, drive := range m.cfg.Drives {
+				if BoolValue(drive.IsRootDevice) {
+					rootDrive = drive
+					rootDriveIndex = i
+					break
```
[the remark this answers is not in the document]
xibz: firecracker fails. Probably should add that in the validation.

jailer.go
xibz: These will never be nil if the validation handler is ran. So that shouldn't be a concern.

jailer.go
xibz: `""` is a valid id according to firecracker's API.

handlers.go
```diff
@@ -102,11 +116,40 @@ var defaultValidationHandlerList = HandlerList{}.Append(
...
 		Name: ValidateCfgHandlerName,
 		Fn: func(ctx context.Context, m *Machine) error {
-			// ensure that the configuration is valid for the
-			// FcInit handlers.
+			// ensure that the configuration is valid for the FcInit handlers.
 			return m.cfg.Validate()
 		},
 	},
+	Handler{
+		Name: "validate.JailerCfg",
```
IRCody: This should be a const I think.

jailer.go
IRCody: So it is run by default (the validation handler) but is not guaranteed to be run right?

jailer.go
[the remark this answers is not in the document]
IRCody: I think it should be configurable.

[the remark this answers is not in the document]
samuelkarp: I think the `jailer` needs to be configurable for the same reason that the `firecracker` binary needs to be configurable: development work (like what we're doing) or people who choose to install it in nonstandard locations.

jailer.go
xibz: Yea, it isn't guaranteed to run if the user explicitly turns it off. We can do a small check to see whether or not `UID` and `GID` are set and if they aren't, return an error.

jailer.go
```diff
@@ -0,0 +1,376 @@
...
+
+	if len(cfg.JailerCfg.NetNS) > 0 {
+		b = b.WithNetNS(cfg.JailerCfg.NetNS)
+	}
+
+	m.cmd = b.Build(ctx)
+	// TODO: hmm, do we actually want to overwrite the socket path?
+	m.cfg.SocketPath = filepath.Join(rootfs, "api.socket")
+
+	m.Handlers.Validation = m.Handlers.Validation.Remove(ValidateCfgHandlerName)
```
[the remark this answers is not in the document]
xibz: Added a comment addressing why this was removed:

```go
  // ValidateCfgHandlerName handler needs to be removed due to checking for    
  // non-existant drive paths and socket path. Since the jailer uses relative    
  // links according to its rootfs then the user would also need those files    
  // relative to running this. 
```

machine_test.go
```diff
@@ -49,9 +49,11 @@ const (
 var testDataPath = "./testdata"
 
 var skipTuntap bool
+var rootEnabled bool
 
 func init() {
 	flag.BoolVar(&skipTuntap, "test.skip-tuntap", false, "Disables tests that require a tuntap device")
+	flag.BoolVar(&skipTuntap, "test.root", false, "Enables tests that require root")
```
samuelkarp: Can you reverse this?  I'd much prefer to have root enabled by default and disabled explicitly so that anyone who *is* running without root knows that they're skipping tests.

machine_test.go
```diff
@@ -60,21 +62,109 @@ func init() {
...
+	m.Handlers.Validation = m.Handlers.Validation.Clear()
 
 	if m == nil {
 		t.Errorf("NewMachine did not create a Machine")
 	}
 }
 
+func TestJailerMicroVMExecution(t *testing.T) {
+	if !rootEnabled {
+		return
```
samuelkarp: Use `t.Skip` when skipping tests.

HACKING.md
```diff
@@ -10,9 +10,9 @@ will run tests in verbose mode.
 
 You need some external resources in order to run the tests, as described below:
 
-1. A firecracker binary (tested with 0.10.1), but any recent version should
-   work.  Must either be installed as `./testdata/firecracker` or the path must
-   be specified through the `FC_TEST_BIN` environment variable.
+1. A firecracker and jailer binary (tested with 0.12.0), but any recent version
```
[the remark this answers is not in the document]
xibz: Done

xibz:
Yea, my test suite passes o.o

nmeyerhans:
What is `testdata/firecracker` when your tests pass? Are you overriding the firecracker binary location (e.g. with `FC_TEST_BIN`)? Or the testdata path (`FC_TEST_DATA_PATH`)?

Because by default, `getFirecrackerBinaryPath()` will return `testdata/firecracker` for the binary path, but elsewhere we treat that path as a directory, and fail as a result:

```
=== RUN   TestJailerMicroVMExecution/TestAttachRootDrive
INFO[0002] Attaching drive testdata/root-drive.img, slot 0, root true. 
DEBU[0002] PUT /drives/0 HTTP/1.1
Host: localhost
User-Agent: Go-http-client/1.1
Content-Length: 100
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip

{"drive_id":"0","is_read_only":true,"is_root_device":true,"path_on_host":"testdata/root-drive.img"}
 
ERRO[0002] Attach drive failed: testdata/root-drive.img: Put http://localhost/drives/0: dial unix testdata/firecracker/test-id/api.socket: connect: not a directory 
ERRO[0002] While attaching drive testdata/root-drive.img, got error Put http://localhost/drives/0: dial unix testdata/firecracker/test-id/api.socket: connect: not a directory 
```

xibz:
Yea, I was overriding it. I am going to have an update which changes the default test binary name to `firecracker.test`

samuelkarp:
Can you leave the default binary name as `firecracker`?  `firecracker.test` won't be on my `$PATH`.

xibz:
Ah okay, another option is to change the chroot path to be `./testdata/jailer/`. I'll go ahead and do that

jailer.go
```diff
@@ -0,0 +1,372 @@
...
+
+			m.cfg.Drives[rootDriveIndex].PathOnHost = &rootdriveFileName
+			m.cfg.KernelImagePath = kernelImageFileName
+			return nil
+		},
+	})
+}
+
+func linkFileToRootFS(cfg JailerConfig, dst, src string) error {
+	if err := os.Link(src, dst); err != nil {
```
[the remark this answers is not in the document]
samuelkarp: If it's a plain file but on a different filesystem we could use a bind-mount instead of copying.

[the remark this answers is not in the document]
nmeyerhans: Actually, I think we always need to copy. The file contains a filesystem. Unless we're doing something with a squashfs+read/write upper layer (which we're not, currently), this file can't safely be shared between concurrent VMs.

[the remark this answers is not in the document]
samuelkarp: Do we want to make the copying part of the SDK or make that requirement of non-concurrent reuse a thing exposed to the client?  I'm asking because we want to avoid copying when we're doing this with firecracker-containerd.

samuelkarp:
The jailer doesn't seem to be loaded from testdata.

```
$ ls testdata/jailer
testdata/jailer
$ make testgo test 
WARN[0000] Failed handler "baz": baz error              
DEBU[0000] Called NewMachine()                          
DEBU[0000] Called NewMachine()                          
DEBU[0000] Called Machine.Start()                       
DEBU[0000] Running handler validate.JailerCfg           
DEBU[0000] Running handler fcinit.StartVMM              
INFO[0000] Called startVMM(), setting up a VMM on testdata/jailer/firecracker/test-id/api.socket 
ERRO[0000] Failed to start VMM: exec: "jailer": executable file not found in $PATH 
WARN[0000] Failed handler "fcinit.StartVMM": exec: "jailer": executable file not found in $PATH 
--- FAIL: TestJailerMicroVMExecution (0.00s)
    machine_test.go:166: Failed to start VMM: exec: "jailer": executable file not found in $PATH
...
$ ls testdata/jailer
ls: cannot access 'testdata/jailer': No such file or directory
```

I assume this is because you're using `testdata/jailer` as a directory in TestJailerMicroVMExecution, you've also called out that `testdata/jailer` is a valid path for the jailer in the README.  Can you change the path in TestJailerMicroVMExecution?

samuelkarp:
Still seeing failures:

<details>

```
$ make test
go test 
WARN[0000] Failed handler "baz": baz error              
DEBU[0000] Called NewMachine()                          
DEBU[0000] Called NewMachine()                          
DEBU[0000] Called Machine.Start()                       
DEBU[0000] Running handler validate.JailerCfg           
DEBU[0000] Running handler fcinit.StartVMM              
INFO[0000] Called startVMM(), setting up a VMM on testdata/jailer-test/firecracker/test-id/api.socket 
DEBU[0000] VMM started socket path is testdata/jailer-test/firecracker/test-id/api.socket 
DEBU[0000] Setting up signal handler                    
INFO[0000] firecracker exited: status=0                 
DEBU[0000] returning from startVMM()                    
DEBU[0000] Running handler fcinit.BootstrapLogging      
DEBU[0000] Created metrics and logging fifos.           
DEBU[0000] PUT /logger HTTP/1.1
Host: localhost
User-Agent: Go-http-client/1.1
Content-Length: 135
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip

{"level":"Debug","log_fifo":"testdata/firecracker.log","metrics_fifo":"testdata/firecracker-metrics","options":null,"show_level":true}
 
WARN[0000] setupLogging() returned Put http://localhost/logger: dial unix testdata/jailer-test/firecracker/test-id/api.socket: connect: no such file or directory. Continuing anyway. 
DEBU[0000] Running handler fcinit.CreateMachine         
DEBU[0000] PUT /machine-config HTTP/1.1
Host: localhost
User-Agent: Go-http-client/1.1
Content-Length: 56
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip

{"cpu_template":"T2","mem_size_mib":256,"vcpu_count":2}
 
ERRO[0000] PutMachineConfiguration returned [PUT /machine-config][204] putMachineConfigurationNoContent  
WARN[0000] Failed handler "fcinit.CreateMachine": Put http://localhost/machine-config: dial unix testdata/jailer-test/firecracker/test-id/api.socket: connect: no such file or directory 
--- FAIL: TestJailerMicroVMExecution (0.01s)
    machine_test.go:182: Failed to start VMM: Put http://localhost/machine-config: dial unix testdata/jailer-test/firecracker/test-id/api.socket: connect: no such file or directory

```
</details>

fctesting/utils.go
```diff
@@ -0,0 +1,39 @@
...
+// RequiresRoot will ensure that tests that require root access are actually
+// root. In addition, this will skip root tests if the test.root-disable is set
+// to true
+func RequiresRoot(t testing.TB) {
+	if rootDisabled {
+		t.Skip("skipping test that requires root")
+	}
+
+	if e, a := 0, os.Getuid(); e != a {
+		t.Fatal("This test must be run as root.")
```
samuelkarp: nit: Maybe add "Add the `-test.root-disable` flag to your `go test` invocation to skip this test instead."

jailer.go
```diff
@@ -0,0 +1,451 @@
...
+
+	if b.daemonize {
+		args = append(args, "--daemonize")
+	}
+
+	return args
+}
+
+// Bin .
+func (b JailerCommandBuilder) Bin() string {
```
mxpv: nit: this needs a proper comment

jailer.go
```diff
@@ -0,0 +1,451 @@
...
+	if len(b.bin) == 0 {
+		return defaultJailerBin
+	}
+
+	return b.bin
+}
+
+// WithBin .
+func (b JailerCommandBuilder) WithBin(bin string) JailerCommandBuilder {
```
mxpv: This needs a comment

jailer.go
```diff
@@ -0,0 +1,451 @@
...
+	args = append(args, b.GID()...)
+	args = append(args, b.ExecFile()...)
+	args = append(args, b.NumaNode()...)
+
+	if len(b.chrootBaseDir) > 0 {
+		args = append(args, b.ChrootBaseDir()...)
+	}
+
+	if len(b.netNS) > 0 {
+		args = append(args, b.NetNS()...)
```
mxpv: ```args = append(args, []string{"--net-ns", b.netNS}```
would be a lot more understandable than adding an extra func for each argument

xibz: It allows for defaults if they are function based and hides that from the `Build` method allowing for a simpler view what what is being built. Maybe a better name rather than `NetNS`, maybe `GetCommandArgsNetNS` or something?

mxpv: I'm not sure I follow. You check defaults with `if len(b.netNS) > 0` and `NetNS` not called anywhere else except here.
So basically this can be
```go
struct Builder {
  NetNS string
}

func (b*builder) Build() string {
  if NetNS != "" {
    args = append(args, "--net-ns", b.netNS)
  }
}
```

xibz: I can add a constructor that will solve the defaults issue. Currently there is no constructor, so users would always have to set some value.

mxpv: Yep. I think constructor is a good place to init defaults 👍

samuelkarp:
I'm getting panics when I run tests here instead of failures:
<details>

```
=== RUN   TestMicroVMExecution/TestStopVMM
time="2019-01-29T16:13:32-08:00" level=warning msg="firecracker exited: signal: terminated"
panic: close of nil channel

goroutine 80 [running]:
github.com/firecracker-microvm/firecracker-go-sdk.(*Machine).startVMM.func3(0xc00017e060, 0xc000174480)
	/home/local/ANT/skarp/projects/Go-firecracker/machine.go:370 +0x73
created by github.com/firecracker-microvm/firecracker-go-sdk.(*Machine).startVMM
	/home/local/ANT/skarp/projects/Go-firecracker/machine.go:367 +0x60f
FAIL	github.com/firecracker-microvm/firecracker-go-sdk	7.041s

```
</details>

(I expect failures right now because I have not set the permissions properly for the test dependencies, but it should not be panicking)

xibz:
@samuelkarp - Ah, found it. Was an issue with an earlier PR. Didn't allocate the channel ><. Fixed.

samuelkarp:
Might want to fix the commit author on the commit you just pushed.

xibz:
@samuelkarp - Good catch! Fixed.

HACKING.md
```diff
@@ -10,9 +10,10 @@ will run tests in verbose mode.
 
 You need some external resources in order to run the tests, as described below:
 
-1. A firecracker binary (tested with 0.10.1), but any recent version should
-   work.  Must either be installed as `./testdata/firecracker` or the path must
-   be specified through the `FC_TEST_BIN` environment variable.
+1. A firecracker and jailer binary (tested with 0.12.0) must either be
+   installed as `./testdata/firecracker` or the path must be specified through
+   the `FC_TEST_BIN` environment variable. The jailer requires go test to be
+   run with sudo and can also be turned off with the `test.root-disable` flag.
```
samuelkarp: Can you update this with the environment variable?

fctesting/utils.go
```diff
@@ -0,0 +1,44 @@
...
+var rootDisabled bool
+
+func init() {
+	if v := os.Getenv(rootDisableEnvName); len(v) != 0 {
+		rootDisabled = true
+	}
+}
+
+// RequiresRoot will ensure that tests that require root access are actually
+// root. In addition, this will skip root tests if the test.root-disable is set
```
samuelkarp: Can you update this?

fctesting/utils.go
```diff
@@ -0,0 +1,44 @@
...
+// to true
+func RequiresRoot(t testing.TB) {
+	if rootDisabled {
+		t.Skip("skipping test that requires root")
+	}
+
+	if e, a := 0, os.Getuid(); e != a {
+		t.Fatal("This test must be run as root. " +
+			"To disable tests that require root, " +
+			"run the tests with the -test.root-disable flag.")
```
samuelkarp: Can you update this?

jailer.go
```diff
@@ -0,0 +1,388 @@
...
+	// restrictive they should be. Possible values are:
+	//
+	//	0 : (default): disabled.
+	//	1 : basic filtering. This prohibits syscalls not whitelisted by Firecracker.
+	//	2 : advanced filtering. This adds further checks on some of the
+	//			parameters of the allowed syscalls.
+	SeccompLevel SeccompLevelValue
+
+	// DevMapperStrategy will dictate how files are transfered to the root drive.
+	DevMapperStrategy HandlersAdaptor
```
samuelkarp: I think we need a different name for this.  "DevMapper" to me implies [device mapper](https://en.wikipedia.org/wiki/Device_mapper), which this does not use.

How about FileMapperStrategy or ChrootStrategy?

HACKING.md
```diff
@@ -6,13 +6,15 @@ Testing
 
 Tests are written using Go's testing framework and can be run with the standard
 `go test` tool.  If you prefer to use the Makefile, `make test EXTRAGOARGS=-v`
-will run tests in verbose mode.
+will run tests in verbose mode. By default, the unit tests require root
+priviledged access. This can be disabled by setting the `DISABLE_ROOT_TESTS`.
```
samuelkarp: nit: privileged

HACKING.md
```diff
@@ -6,13 +6,15 @@ Testing
...
 
 You need some external resources in order to run the tests, as described below:
 
-1. A firecracker binary (tested with 0.10.1), but any recent version should
-   work.  Must either be installed as `./testdata/firecracker` or the path must
-   be specified through the `FC_TEST_BIN` environment variable.
+1. A firecracker and jailer binary (tested with 0.12.0) must either be
+   installed as `./testdata/firecracker` or the path must be specified through
+   the `FC_TEST_BIN` environment variable. The jailer requires go test to be
+   run with sudo and can also be turned off with the `test.root-disable` flag.
```
samuelkarp: `test.root-disable` needs to be updated to `DISABLE_ROOT_TESTS` environment variable

jailer.go
```diff
@@ -0,0 +1,388 @@
...
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+)
+
+const (
+	defaultJailerPath = "/srv/jailer/firecracker"
```
samuelkarp: I don't really think I like `/srv`; I'm not sure it really fits into the [standard definition](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard).  What about `/var/lib/jailer/firecracker`?

@nmeyerhans Do you have thoughts on this?

xibz: The SDK isn't setting that. If not chroot base directory was provided when executing the jailer, it'll go to `/srv/jailer/firecracker/`.

We can optionally have our own default, but that would mean we would always be setting the chroot base directory.

samuelkarp: This default comes from the jailer itself?  If it comes from the jailer itself, then I don't think we should override the default.

xibz: Cool, yea, from the jailer itself. I added a comment to make that more clear.

machine_test.go
```diff
@@ -61,18 +65,164 @@ func TestNewMachine(t *testing.T) {
...
+func TestJailerMicroVMExecution(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+	}
+	fctesting.RequiresRoot(t)
+
+	jailerUID := 123
+	jailerGID := 100
+	var err error
+	if v := os.Getenv("SUDO_UID"); v != "" {
```
samuelkarp: const?

machine_test.go
```diff
@@ -61,18 +65,164 @@ func TestNewMachine(t *testing.T) {
...
+		if jailerUID, err = strconv.Atoi(v); err != nil {
+			t.Fatalf("Failed to parse 'SUDO_UID'")
+		}
+	}
+
+	if v := os.Getenv("SUDO_GID"); v != "" {
```
samuelkarp: const?

xibz:
> Some minor comments, but looks really good other than those.

Done

samuelkarp:
Since we now have CI, would you mind rebasing on master and seeing that the new tests pass?  You'll probably need to change the .buildkite/pipeline.yml to make a distinction between root and non-root tests, similar to [here](https://github.com/firecracker-microvm/firecracker-containerd/blob/83b8a18b81ae606fe4c6997021dd50c8438ef193/.buildkite/pipeline.yml#L25-L33).  It's also possible that the build infrastructure might need to change; I can work with you on that, but I tried to make it work out of the box (see the `ln -s` lines in the existing .buildkite/pipeline.yml to link the right binaries.

.buildkite/pipeline.yml
```diff
@@ -31,6 +31,7 @@ steps:
       - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
       - 'ln -s /usr/local/bin/firecracker testdata/firecracker'
       - 'ln -s /usr/local/bin/jailer testdata/jailer'
-      - "make test EXTRAGOARGS='-v -count=1'"
+      - 'sudo chown 999:999 testdata/vmlinux'
+      - "sudo make test EXTRAGOARGS='-v -count=1'"
```
samuelkarp: Let's leave this step alone, and make a new one that is specifically for root tests (the way that I did [here](https://github.com/firecracker-microvm/firecracker-containerd/blob/83b8a18b81ae606fe4c6997021dd50c8438ef193/.buildkite/pipeline.yml#L25-L33)).

.buildkite/pipeline.yml
```diff
@@ -31,6 +31,7 @@ steps:
       - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
       - 'ln -s /usr/local/bin/firecracker testdata/firecracker'
       - 'ln -s /usr/local/bin/jailer testdata/jailer'
-      - "make test EXTRAGOARGS='-v -count=1'"
+      - 'sudo chown 999:999 testdata/vmlinux'
```
samuelkarp: Do you need to `chown` it?  999 is not guaranteed to be the UID of the current user.

.buildkite/pipeline.yml
```diff
@@ -31,6 +31,8 @@ steps:
       - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
       - 'ln -s /usr/local/bin/firecracker testdata/firecracker'
       - 'ln -s /usr/local/bin/jailer testdata/jailer'
-      - "make test EXTRAGOARGS='-v -count=1'"
+      - 'chown 999:999 testdata/vmlinux'
+      - 'chown 999:999 testdata/root-drive.img'
+      - "sudo TMPDIR=/local/tmp make test EXTRAGOARGS='-v -count=1'"
```
samuelkarp: Please see my previous comment.  I'd like to have this addressed before we merge.

.buildkite/pipeline.yml
xibz: Was hard coding for testing. I managed to move it into code. However, seems to be another issue with the tests.

.buildkite/pipeline.yml
xibz: Sounds good. Will move that into a root section

.buildkite/pipeline.yml
samuelkarp: Want to pair-program on this tomorrow?  Let's get this resolved and merged!

PR #43, opened 2018-12-26
Adding drives builder

xibz:
Adds a drives builder to help users with building a series of drives
without needing to know the intricate details of firecracker's API.

### DIFF
--- a/drives.go
+++ b/drives.go
@@ -1,0 +1,65 @@
+package firecracker
+
+import (
+	"strconv"
+
+	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
+)
+
+const rootDriveName = "root-drive"
+
+// DrivesBuilder is a builder that will build an array of drives used to set up
+// the firecracker microVM. The DriveID will be an incrementing number starting
+// at one
+type DrivesBuilder struct {
+	rootDrive models.Drive
+	drives    []models.Drive
+}
+
+// NewDrivesBuilder will return a new DrivesBuilder with a given rootfs.
+func NewDrivesBuilder(rootDrivePath string) DrivesBuilder {
+	return DrivesBuilder{}.WithRootDrive(rootDrivePath)
+}
+
+// DriveOpt represents an optional function used to allow for specific
+// customization of the models.Drive structure.
+type DriveOpt func(*models.Drive)
+
+// WithRootDrive will set the given builder with the a new root path. The root
+// drive will be set to read and write by default.
+func (b DrivesBuilder) WithRootDrive(rootDrivePath string, opts ...DriveOpt) DrivesBuilder {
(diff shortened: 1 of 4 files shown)


drives.go
```diff
@@ -0,0 +1,63 @@
+package firecracker
+
+import (
+	"strconv"
+
+	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
+)
+
+// DrivesBuilder is a builder that will build an array of drives used to set up
+// the firecracker microVM. The DriveID will be an incrementing number starting
```
samuelkarp: (I'm not familiar enough with the Firecracker API yet) What are the requirements around the DriveID? Is it just uniqueness or are there other requirements too?

mxpv: Have a look on conversation [here](https://github.com/firecracker-microvm/firecracker-containerd/pull/49). It adds some context to this changes.

xibz: Based on the docs it'll stomp over any pre-existing drive with the same ID, and no requirements I see other than `DriveID` not being empty which is returned from firecracker's API.

samuelkarp: @mxpv I saw that, but that didn't answer my question.  @xibz's response is what I was looking for.

drives.go
```diff
@@ -0,0 +1,65 @@
...
+	return DrivesBuilder{}.WithRootFS(rootfsPath)
+}
+
+// DriveOpt represents an optional function used to allow for specific
+// customization of the models.Drive structure.
+type DriveOpt func(*models.Drive)
+
+// WithRootFS will set the given builder with the a new root path. The rootfs
+// will be set to read and write by default.
+func (b DrivesBuilder) WithRootFS(rootfsPath string, opts ...DriveOpt) DrivesBuilder {
```
samuelkarp: Is this strictly a (single) filesystem, or is this a device (or device image) that may contain a partition table with multiple filesystems inside?

xibz: It is whatever firecracker's API supports. I know it supports a single filesystem, but where can I find a device image to test with?

samuelkarp: @ilianaw published a guide [here](https://iweller.people.a2z.com/con404/) ("Building an AMI").

samuelkarp: I think instead of testing, we could also just rename the function to be `WithRootDrive` instead of `WithRootFS` and be fine.  (Same thing for the `rootDeviceName` constant.)

drives.go
```diff
@@ -0,0 +1,65 @@
+package firecracker
+
+import (
+	"strconv"
+
+	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
+)
+
+const rootDriveName = "rootfs"
```
samuelkarp: Can you change this to "root" or "root-drive"?

PR #44, opened 2018-12-27
Adding custom templates for go-swagger

xibz:
This gets rid of the need for mockgen and uses go-swagger to instead
generate a mock client. In addition, this change also gets rid of the
wrapper client and instead uses the client.Firecracker instead. Also,
the change also generates an interface in the operations package and
also needed some updating to the default go-swagger template to use the
interface instead of types directly.

This PR fixes two things that are known issues with mockgen. The first is that mockgen does not generate correctly when using structures in the same package. Additionally, mockgen takes at least 30 minutes to generate for Go 1.11.


firecracker.go
```diff
@@ -63,88 +56,5 @@ func NewFirecrackerClient(socketPath string, logger *logrus.Entry, debug bool) *
 
 	httpClient.SetTransport(transport)
 
-	return &FirecrackerClient{client: httpClient}
-}
-
-func (f *FirecrackerClient) PutLogger(ctx context.Context, logger *models.Logger) (*ops.PutLoggerNoContent, error) {
```
samuelkarp: So I get that this is a wrapper and it's going to be tedious to maintain, but are we really ready to get rid of it?  The wrapper does a couple of things that I think are valuable:

* For trivial operations (like PutLogger), it gets rid of the need to manage parameter structs like `PutLoggerParams`.  The result of this is a more natural Go API that involves function arguments rather than structs.
* For many of the operations, we set default timeout values.  Without something like this, every caller here would need to worry about the timeouts.

You can already see the effect this has on callers when you look at the changes that you had to make to `machine.go`.

xibz: > For trivial operations (like PutLogger), it gets rid of the need to manage parameter structs like PutLoggerParams. The result of this is a more natural Go API that involves function arguments rather than structs.

Yea, but I don't know if that's worth keeping for a minor improvement in ease of use with now needing to maintain this. The goal of this PR was to eliminate any upkeep of structures when the swagger definition changes. And this could also be fixed to adding a factory function if needed.

> For many of the operations, we set default timeout values. Without something like this, every caller here would need to worry about the timeouts.

For the timeouts, I would be more inclined to agree to keep this if we were using those operation in multiple spots. However, since it is only being used in one location, I don't see any reason to keep it.  If there is a case where we need to call the operation in two spots that can be mitigated with helper functions.

xibz:
@mxpv - I can't remember what failed, but it occurred once when Noah and I were trying to build against the latest swagger. The main reason for this change isn't that, but to eliminate any upkeep when the swagger definition has been updated.

firecracker.go
xibz: Actually I'll go ahead and do that, add a customization in the template to add a `PutLogger` customization.

xibz:
Closing this PR as I realized removing this isn't the best path forward.

xibz:
@mxpv - Here's the [issue]. Found it trying to regenerate with new PR. Looks like it is a bug with Go 1.11 and gomock. May reopen this and keep the mock client generation stuff as I think mockgen is a little too complicated for what we need.

machine_test.go
```diff
@@ -138,7 +135,22 @@ func TestMicroVMExecution(t *testing.T) {
 		exitchannel <- m.Wait(vmmCtx)
 		close(exitchannel)
 	}()
-	time.Sleep(2 * time.Second)
+
+	vmmAlive := false
+	const maxRetry = 5
+	for i := 0; i < maxRetry; i++ {
```
IRCody: Is this worth generalizing since it's used in a couple of places as something like`func waitVmmAlive(client, maxRetries) error`?

Then in the tests anywhere you want to wait for VMM to startup you can do:
```go
if err := waitVmmAlive(m.client, maxRetries); err != nil {
        t.Fatal(err)
}
```

xibz: Done

utils.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	defaultAliveVMMCheckDur = 10 * time.Millisecond
+)
+
+// waitForAliveVMM will check for periodically to see if the firecracker VMM is
+// alive. If the VMM takes too long in starting, an error signifying that will
+// be returned.
+func waitForAliveVMM(ctx context.Context, client *Client, maxRetries int) error {
+	t := time.NewTicker(defaultAliveVMMCheckDur)
```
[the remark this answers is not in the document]
xibz: Yea, I can add it. Since this isn't limited to just tests, although that's where it is only used, it would be a good idea to have.

client/client_transports_test.go
```diff
@@ -1,27 +1,39 @@
...
 
 func TestNewUnixSocketTransport(t *testing.T) {
 	done := make(chan bool)
 
 	expectedMessage := "PUT /logger HTTP/1.1\r\nHost: localhost\r\nUser-Agent: Go-http-client/1.1\r\nContent-Length: 0\r\nAccept: application/json\r\nContent-Type: application/json\r\nAccept-Encoding: gzip\r\n\r\n"
 	socketPath := "testingUnixSocket.sock"
-	addr, _ := net.ResolveUnixAddr("unix", socketPath)
-	listener, _ := net.ListenUnix("unix", addr)
-	defer listener.Close()
+	addr, err := net.ResolveUnixAddr("unix", socketPath)
```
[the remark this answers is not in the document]
xibz: Done!

utils.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+const (
+	defaultAliveVMMCheckDur = 10 * time.Millisecond
+)
+
+// waitForAliveVMM will check for periodically to see if the firecracker VMM is
+// alive. If the VMM takes too long in starting, an error signifying that will
+// be returned.
+func waitForAliveVMM(ctx context.Context, client *Client, maxRetries int) error {
```
samuelkarp: I'd prefer to see this use `<-ctx.Done` for cancellation instead of a `maxRetries`.  Callers can then use `context.WithDeadline` to achieve a timeout.

xibz: Done!

utils.go
```diff
@@ -0,0 +1,29 @@
...
+func waitForAliveVMM(ctx context.Context, client *Client) error {
+	t := time.NewTicker(defaultAliveVMMCheckDur)
+	defer t.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-t.C:
+			if _, err := client.GetMachineConfig(); err == nil {
```
mxpv: This is almost same as `waitForSocket`. Would it make sense to refactor/leave one waiter?

machine.go
```diff
@@ -543,23 +532,21 @@ func (m *Machine) EnableMetadata(metadata interface{}) {
...
-	if err == nil {
-		var message string
-		if respnocontent != nil {
-			message = respnocontent.Error()
-		}
-		m.logger.Printf("SetMetadata successful: %s", message)
-	} else {
+	if _, err := m.client.PutMmds(ctx, metadata); err != nil {
 		m.logger.Errorf("Setting metadata: %s", err)
+		return nil
```
mxpv: Why do you return `nil` here?  `return err` ?

firecracker.go
```diff
@@ -16,109 +16,239 @@ package firecracker
...
+		opt(mc)
+	}
 
 	return f.client.Operations.PutMachineConfiguration(mc)
 }
 
-func (f *FirecrackerClient) PutGuestBootSource(ctx context.Context, source *models.BootSource) (*ops.PutGuestBootSourceNoContent, error) {
+// PutGuestBootSourceOpt is a functional option to be used for the
+// PutGuestBootSource API in setting any additional optional fields.
+type PutGuestBootSourceOpt func(*ops.PutGuestBootSourceParams)
```
mxpv: It looks like the comment for this is outdated

firecracker_test.go
```diff
@@ -0,0 +1,55 @@
...
+		WithSocketPath(socketpath).
+		Build(ctx)
+
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("failed to start firecracker vmm: %v", err)
+	}
+
+	defer func() {
+		if err := cmd.Process.Kill(); err != nil {
+			t.Errorf("failed to kill process")
```
mxpv: It makes sense to log error here.

utils.go
xibz: Yea, issue is the `exitchan` handles early exit of the machine which isn't needed for the utility. The code is small and won't ever change, so I am not too worried about it.

firecracker.go
```diff
@@ -16,109 +16,239 @@ package firecracker
...
+type Client struct {
 	client *client.Firecracker
 }
 
-// NewFirecrackerClient creates a FirecrackerClient
-func NewFirecrackerClient(socketPath string, logger *logrus.Entry, debug bool) *FirecrackerClient {
-	unixTransport := client.NewUnixSocketTransport(socketPath, logger, debug)
-	firecracker := client.New(unixTransport, strfmt.NewFormats())
+// NewClient creates a Client
+func NewClient(socketPath string, logger *logrus.Entry, debug bool, opts ...ClientOpt) *Client {
```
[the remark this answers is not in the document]
xibz: This will require some thought due to how the code is currently laid out.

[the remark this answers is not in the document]
xibz: So, after some thought, we probably should leave as is. The reason for this is we should make it incredibly easy to enable/disable logging and debugging. If we decide to go the other route and require functional options, it would need to be at the transport level. Meaning it would force users to create a custom transport with a custom API client then finally create a firecracker client. That seems like a lot of work to enable logging and debugging. What do you think?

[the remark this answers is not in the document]
mxpv: As for me the only parameter we should force user to pass - `socketPath` which is necessary for client to work properly. Everything else is nice to have customizations, but they are not required for correct functioning. So it's still good to able to configure debugging, to can be hidden to opts.

samuelkarp:
This does not build for me:

<details>

```
$ make test EXTRAGOARGS='-v -count=1'
go test ./... -v -count=1
# github.com/firecracker-microvm/firecracker-go-sdk [github.com/firecracker-microvm/firecracker-go-sdk.test]
./machine.go:30:2: imported and not used: "github.com/firecracker-microvm/firecracker-go-sdk/client/operations" as ops
FAIL	github.com/firecracker-microvm/firecracker-go-sdk [build failed]
=== RUN   TestNewUnixSocketTransport
--- PASS: TestNewUnixSocketTransport (0.00s)
PASS
ok  	github.com/firecracker-microvm/firecracker-go-sdk/client	0.020s
?   	github.com/firecracker-microvm/firecracker-go-sdk/client/models	[no test files]
?   	github.com/firecracker-microvm/firecracker-go-sdk/client/operations	[no test files]
?   	github.com/firecracker-microvm/firecracker-go-sdk/fctesting	[no test files]
Makefile:25: recipe for target 'all-tests' failed
make: *** [all-tests] Error 2
```

</details>

xibz:
Fixed.

PR #45, opened 2018-12-28
seperating out integration tests from unit tests

xibz:
This change makes modification to separate out integration tests from
unit tests. This makes it a little nicer to run only unit tests when
developing and finally running integration as a last check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,15 @@ EXTRAGOARGS:=
 
 all: build
 
-generate build clean test:
+test: all-tests
+
+unit-tests:
+	go test -short $(EXTRAGOARGS)
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -16,7 +16,15 @@ EXTRAGOARGS:=
...
+unit:
+	go test $(EXTRAGOARGS)
+
+integration:
+	go test -tags integration $(EXTRAGOARGS)
+
+generate build clean:
 	go $@ $(EXTRAGOARGS)
 
 .PHONY: all generate clean build test
```
samuelkarp: Add `unit` and `integration` here

PR #47, opened 2019-01-03
Adds a factory function for buiilding a HTTP client for firecracker u…

kshitij8:
…sing unix socket

*Issue #, if available:*

*Description of changes:*
It should be a common pattern to want to configure the firecracker client on a unix socket (as visible in firecracker.go)  
Anyone who wants a customised behaviour than what is provided by the `FirecrackerClient` would be able use it.


client/firecracker_client_factory_test.go
```diff
@@ -0,0 +1,47 @@
...
+	listener, _ := net.ListenUnix("unix", addr)
+	defer listener.Close()
+
+	go func() {
+		conn, err := listener.AcceptUnix()
+		if err != nil {
+			t.Error(err)
+		}
+
+		for {
```
xibz: Does this for loop need to be there? My assumption would be `Read` blocks until something can be read.

client/firecracker_client_factory_test.go
```diff
@@ -0,0 +1,47 @@
...
+	done := make(chan bool)
+
+	expectedMessage := "PUT /logger HTTP/1.1\r\nHost: localhost\r\nUser-Agent: Go-http-client/1.1\r\nContent-Length: 0\r\nAccept: application/json\r\nContent-Type: application/json\r\nAccept-Encoding: gzip\r\n\r\n"
+	socketName := "testingUnixSocket.sock"
+	addr, _ := net.ResolveUnixAddr("unix", socketName)
+	listener, _ := net.ListenUnix("unix", addr)
+	defer listener.Close()
+
+	go func() {
+		conn, err := listener.AcceptUnix()
```
xibz: Unix file needs to be closed and cleaned up.

client/firecracker_client_factory_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+		for {
+			buf := make([]byte, 512)
+			nr, err := conn.Read(buf)
+			if err != nil {
+				return
+			}
+
+			data := string(buf[0:nr])
+			assert.Equal(t, expectedMessage, data, "expectedMessage received on socket is different than what was sent")
```
xibz: can we just do a straight compare here to be consistent with the other tests?

```go
if e, a := expectedMessage, data; e != a {
    t.Errorf("expected message to be %q, but received %q", e, a)
}
```

client/firecracker_client_factory.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/sirupsen/logrus"
+
+	httptransport "github.com/go-openapi/runtime/client"
+)
+
+// NewHttpClientForUNIXSocket creates a new firecracker HTTP client configured at the specified a Unix socketPath.
+func NewHttpClientForUNIXSocket(socketPath string, logger *logrus.Entry, debug bool) *Firecracker {
```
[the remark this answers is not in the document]
kshitij8: My take on that was, if anyone will care about a "UnixSocketTransport" or will they care more about a ready to use client. The difference between the two being formats can be passed explicitly with the suggestion above and a consumer can use sets in the unixTransport (which we already expose arguments).  
   
Another reason was, that it exposes clients to OpenAPIs: `runtime.Client` which didn't seem very nice if we go by the idea that we should keep the surface simpler and easy to use.

client/firecracker_client_factory_test.go
kshitij8: If I'm not mistaken, defer is closing the unix file.  
I was able to verify that the socket file is being deleted after test execution

client/firecracker_client_factory_test.go
kshitij8: Sure! Happy to get rid of the assert if we're staying staying away from testify for now.

samuelkarp: For something like string comparison, testify's diff on errors is pretty useful.  I'd lean toward keeping this as `assert.Equal`.

client/firecracker_client_factory_test.go
kshitij8: Indeed! Verified that it's not needed. 
Removed

client/firecracker_client_factory_test.go
xibz: @samuelkarp - If that is how we want to move forward, we probably want to take some time and update the rest of the tests to use the testify package. I am more for less dependencies the better for an SDK.

client/firecracker_client_factory.go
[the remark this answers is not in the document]
kshitij8: > Not sure I follow since this returns a http.Transport and creates the same client to begin with

The return type is: [*Runtime](https://github.com/go-openapi/runtime/blob/master/client/runtime.go#L200)  

The New accepts the interface: [runtime.ClientTransport](https://github.com/go-openapi/runtime/blob/master/client_operation.go#L38)

> It'll simplify the function and prevent the function from doing too much.

I agree with the idea of not having the client do too much, in this case the only extra thing `NewHttpClientForUNIXSocket` does over the proposed `NewUnixSocketTransport` is: `return client.New(unixTransport, strfmt.NewFormats())`  

I can update the PR based on whatever makes more sense to expose as a function, one which returns a UnixTransport or a UnixClient or both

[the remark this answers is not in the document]
xibz: Yea, I prefer having this be just `NewUnixSocketTransport`. Again, thank you for taking the time to do this and whenever you get the chance, please update.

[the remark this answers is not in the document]
kshitij8: Sure! Updated

kshitij8:
Updated the PR

Issue #52, opened 2019-01-10
API calls take exactly the timeout of time


samuelkarp:
API calls made through the Unix socket appear to take exactly 500ms, which is the value of [`firecrackerRequestTimeout`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/1585d53f793483c96d5d9058ca705d97d1b0558b/firecracker.go#L28).  I can observe this by manipulating the timeout in the [`runtime.ClientOperation`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/1585d53f793483c96d5d9058ca705d97d1b0558b/client/client_transports_test.go#L50).

xibz:
I can't seem to reproduce this on my home machine. I pulled #53 locally and change the time to 5 minutes and moved the submit to a go routine and it passed o.o

samuelkarp:
Can you try changing the time to 5 minutes without making any other changes?  Moving the submit to a goroutine would hide the symptom; the bytes are submitted to the socket effectively instantly (which is why I was able to get away with 1ms) but then submit continues to block until the timeout.

PR #53, opened 2019-01-11
client: Set operation timeout to 1ms

samuelkarp:
*Issue #, if available:*
Related: https://github.com/firecracker-microvm/firecracker-go-sdk/issues/52

*Description of changes:*
client: Set operation timeout to 1ms
Validated by running `make unit-tests EXTRAGOARGS='-v -count=1'`

### DIFF
--- a/client/client_transports_test.go
+++ b/client/client_transports_test.go
@@ -1,22 +1,40 @@
 package client
 
 import (
+	"net"
+	"os"
+	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


xibz:
I just realize the `time.After` channel isn't really useful since the `Submit` call is doing all the work so basically, the timer is waiting for nothing. Let's move the `unixTransport.Submit(testOperation())` into a go routine which should now fail the test if it takes longer than 100ms.

xibz:
Closing this PR in favor of #54

PR #54, opened 2019-01-11
client: fixing test bug with hanging requests

### LINKED ISSUE #52: API calls take exactly the timeout of time

xibz:
Fixing test bug with hanging requests

Prior tests caused the HTTP client to hang waiting for a response. This
will now properly write a response and closing the request.

Fixes #52

### DIFF
--- a/client/client_transports_test.go
+++ b/client/client_transports_test.go
@@ -1,42 +1,45 @@
 package client
 
 import (
+	"context"
+	"net"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+
 	"github.com/firecracker-microvm/firecracker-go-sdk/client/operations"
 	"github.com/go-openapi/runtime"
 	"github.com/stretchr/testify/assert"
-	"net"
-	"testing"
-	"time"
 )
+
+const expectedEndpointPath = "/test-operation"
 
 func TestNewUnixSocketTransport(t *testing.T) {
 	done := make(chan bool)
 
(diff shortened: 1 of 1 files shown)


client/client_transports_test.go
```diff
@@ -1,42 +1,42 @@
...
-
-		data := string(buf[0:nr])
-		assert.Equal(t, expectedMessage, data, "expectedMessage received on socket is different than what was sent")
-		done <- true
+	listener, err := net.Listen("unix", socketPath)
+	if err != nil {
+		panic(err)
+	}
+
+	const expectedEndpoint = "/test-operation"
```
samuelkarp: You've defined this as a const but you're only using it once.  Do you want to move it out of the scope of this function and use it here and in the `testOperation()` function?

client/client_transports_test.go
```diff
@@ -1,42 +1,42 @@
...
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			assert.Equal(t, expectedEndpoint, req.URL.String())
+			w.WriteHeader(http.StatusOK)
+			close(done)
+		}),
+	}
+
+	go server.Serve(listener)
+	defer func() {
+		listener.Close()
```
samuelkarp: I think you should add `server.Shutdown()` above this line.

PR #55, opened 2019-01-12
add support for rate limiting

xibz:
Adds support for rate limiting for inbound and outbound data streams of
firecracker. This also introduces a TokenBuilder which can be passed
into the NewRateLimiter factory function to return a new rate limiter.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/firecracker-microvm/firecracker-go-sdk"
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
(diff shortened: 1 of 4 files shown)


example_test.go
```diff
@@ -151,3 +152,63 @@ func ExampleDrivesBuilder_DriveOpt() {
 		panic(err)
 	}
 }
+
+func ExampleNetworkInterfaces_RateLimiting() {
+	// construct the limitations of the bandwidth for firecracker
+	bandwidthB := firecracker.TokenBucketBuilder{}.
+		WithInitialSize(1024 * 1024).
```
samuelkarp: I'm not super wild about this naming.  The interface is very generic, but it makes it hard to see what the limiter is limiting; is this constrained to 1024*1024 bytes of bandwidth per 30s?  I have to rely on the name of the variable `bandwidthB` to guess what it's limiting and infer that the limit unit is bytes.

xibz: Yea, that makes sense. I've went ahead and made the changes.

PR #57, opened 2019-01-23
Make machine.Wait() safe for multiple calls

samuelkarp:
*Description of changes:*
Prior to this change, multiple calls to `Wait()` could result in the error being lost, and for subsequent callers to wait until their context was canceled rather than waiting for the VMM to exit.  This change makes it such that multiple callers of `Wait()` will all block for the VMM exit (or their individual context cancellation) and will receive the same error (for non-canceled contexts).

This behavior was causing `TestMicroVMExecution` to take 30 seconds every time, which was the length of the `vmmCtx` timeout.  After this change, `TestMicroVMExecution` takes ~7 seconds, which is the length of the calls to `time.Sleep` in that test.

The other commits in this pull request are either in support of making `Wait()` safe (like enforcing that `Start()` can only be called once) or are cleaning up code.

Please view the commits individually to associate them with the change called out in the commit message.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -134,17 +134,18 @@ func (cfg *Config) Validate() error {
 
 // Machine is the main object for manipulating Firecracker microVMs
 type Machine struct {
+	// Metadata is the associated metadata that will be sent to the firecracker
+	// process
+	Metadata interface{}
+	// Handlers holds the set of handlers that are run for validation and start
+	Handlers Handlers
+
 	cfg           Config
 	client        Firecracker
 	cmd           *exec.Cmd
 	logger        *log.Entry
 	machineConfig models.MachineConfiguration // The actual machine config as reported by Firecracker
-
-	// Metadata is the associated metadata that will be sent to the firecracker
-	// process
-	Metadata interface{}
-	errCh    chan error
-	Handlers Handlers
+	errCh         chan error
 }
 
 // Logger returns a logrus logger appropriate for logging hypervisor messages
--- a/machine.go
+++ b/machine.go
@@ -518,11 +518,6 @@ func (m *Machine) addVsock(ctx context.Context, dev VsockDevice) error {
 	return nil
 }
 
-// StartInstance starts the Firecracker microVM
-func (m *Machine) StartInstance(ctx context.Context) error {
-	return m.startInstance(ctx)
-}
-
 func (m *Machine) startInstance(ctx context.Context) error {
 	info := models.InstanceActionInfo{
 		ActionType: models.InstanceActionInfoActionTypeInstanceStart,
--- a/machine.go
+++ b/machine.go
@@ -22,6 +22,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"strconv"
+	"sync"
 	"syscall"
 	"time"
 
@@ -33,6 +34,8 @@ import (
 const (
 	userAgent = "firecracker-go-sdk"
 )
+
+var ErrAlreadyStarted = errors.New("firecracker: machine already started")
 
 // Firecracker is an interface that can be used to mock
 // out an Firecracker agent for testing purposes.
@@ -146,6 +149,7 @@ type Machine struct {
 	logger        *log.Entry
 	machineConfig models.MachineConfiguration // The actual machine config as reported by Firecracker
 	errCh         chan error
+	startOnce     sync.Once
 }
 
 // Logger returns a logrus logger appropriate for logging hypervisor messages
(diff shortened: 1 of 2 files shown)


machine.go
```diff
@@ -148,8 +148,12 @@ type Machine struct {
 	cmd           *exec.Cmd
 	logger        *log.Entry
 	machineConfig models.MachineConfiguration // The actual machine config as reported by Firecracker
-	errCh         chan error
-	startOnce     sync.Once
+	// startOnce ensures that the machine can only be started once
+	startOnce sync.Once
+	// startCh is a channel which gets closed when the VMM is started
```
IRCody: exitCh?

samuelkarp: Oops, deleted the wrong lines.  startCh was in an earlier draft of this commit, but I removed it as it was no longer useful.

samuelkarp: Fixed.

machine.go
```diff
@@ -226,22 +237,35 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
 // Start will iterate through the handler list and call each handler. If an
 // error occurred during handler execution, that error will be returned. If the
 // handlers succeed, then this will start the VMM instance.
+// Start may only be called once per Machine.  Subsequent calls will return
+// ErrAlreadyStarted.
 func (m *Machine) Start(ctx context.Context) error {
 	m.logger.Debug("Called Machine.Start()")
+	alreadyStarted := true
+	m.startOnce.Do(func() {
```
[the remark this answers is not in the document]
samuelkarp: @nmeyerhans and I were talking about this yesterday and we're both leaning toward enforcing the limit here.  Fundamentally, a Firecracker microVM is a one-shot thing; each microVM is tied to exactly one VMM process, and then when the VMM process exits the microVM is gone.  There isn't a notion of restarting a microVM in Firecracker; the closest that you can get is to reuse the same settings and the same disk image, and then launch a new one.

I think it makes sense to tie the `Machine` object to represent exactly one microVM, and to have this same restriction.  As a convenience, we might want to add a `Copy` or `Clone` method that spits out another independent `Machine` if you do want to "restart" it in the future.

machine.go
```diff
@@ -196,7 +205,9 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
 		return nil, err
 	}
 
-	m := &Machine{}
+	m := &Machine{
+		exitCh: make(chan struct{}),
```
xibz: Why not have this be an `chan error`? On a successful flow of the `Machine` you simply could just the close the channel and any error can be sent through it.The reason why is the introduction of the `err` field seems unneeded if we use this and allows for multiple errors to be sent down the channel if we wanted.

samuelkarp: A channel has no persistence; once you read an item out of the channel, it's gone.  To support `Wait` returning the same error to every caller, we need to store it in a variable instead of a channel.

machine.go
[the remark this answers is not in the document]
xibz: I'm really torn here. I think some things make sense and I get since you can't start an instance multiple times it might make sense to reflect that in code. However, there is a distinction in that the `Machine` will not start an instance multiple times but start the whole life cycle multiple times. I wonder if it makes sense to leave the control up to the user. If they want to go through the `Machine`'s start and stop in cycles, they can. I am not too sure which approach is better, but I am leaning more towards letting users control that. Thoughts?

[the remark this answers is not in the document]
samuelkarp: I think it's an easier programming model (both for us as authors of this package and for users) to assert that a `Machine`'s lifecycle is one-way.  It makes it clear that each instance is independent and cannot be restarted, and makes it easier to reason about (i.e., you get guarantees that you can program against).  For users who want to go in cycles, I think a `Copy` or `Clone` convenience method makes sense.

machine.go
xibz: My biggest concern is potentially stomping on top of pre-existing errors set on `m.err`. If we could make it impossible to stomp over errors, I'd be more in favor of this.

samuelkarp: `m.err` is written exactly once, on line 355 inside `startVMM`.  The behavior of `Start` only being called once should make this reasonably safe, but I could make it safer with another `sync.Once` in `startVMM` if you're concerned about it.

machine.go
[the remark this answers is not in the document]
xibz: It definitely is an easier programming model for us, but I don't see it as being any more easier on the user at least. However, if the `Machine` object ever does get complex, it'll make that easier to reason about. I just don't know if it'll ever get that complex. Meh, I'm okay with it lol.

machine.go
xibz: Naw, don't think we need another `sync.Once` for this. We had something similar in the AWS SDK for Go which was on requests. Requests also had a `Copy` method, but we would still stomp over the errors from time to time which made debugging some issues difficult. I am okay with taking this as is for now since the error isn't exposed which will allow us to change it if we can think of a better method of addressing of the stomping over errors.

PR #67, opened 2019-02-22
Syncing with v0.15.0 of Firecracker

xibz:
Two changes were required in updating to v0.15.0 in that the first changed some values to be pointers due to allowing them to have default values. 
There is now validation that the drive name has to be alphanumeric and underscores. Due to this the root drive name, root-drive, has been renamed to root_drive.

Also fixes a bug where `StopVMM` was racing in `TestMicroVMExecution` and would sometimes succeed because the defer would happen last and do nothing with the error. However, if the defer happened before the `TestStopVMM`, then that's when the error is captured and fails.


samuelkarp:
Our CI is not currently testing Firecracker v0.15.0.  Do you want help getting the CI to use the new version?

xibz:
Yea, I would love to do that.

machine.go
```diff
@@ -397,11 +397,11 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
-		LogFifo:       m.cfg.LogFifo,
-		Level:         m.cfg.LogLevel,
-		MetricsFifo:   m.cfg.MetricsFifo,
-		ShowLevel:     true,
-		ShowLogOrigin: false,
+		LogFifo:       String(m.cfg.LogFifo),
+		Level:         String(m.cfg.LogLevel),
+		MetricsFifo:   String(m.cfg.MetricsFifo),
+		ShowLevel:     Bool(true),
+		ShowLogOrigin: Bool(false),
```
nmeyerhans: The commit message doesn't shed any light on why we need these changes. Are they actually required as part of updating to support v0.15.0?

xibz: Yea, the newly generated swagger models changed those to be pointers. I think it has something to do with setting default values and required fields, ie if a field isn't set, nil, set to default value.

PR #71, opened 2019-02-26
Adding support for testing against v0.15.0 firecracker

xibz:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -29,8 +29,8 @@ steps:
   - label: ':hammer: tests'
     commands:
       - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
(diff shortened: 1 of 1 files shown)


xibz:
Talked with @nmeyerhans and discussed why these tests will never pass unless we update to the SDK code base to be compatiable with v0.15.0. 

The issue is v0.12.0 of the SDK will cause Firecracker v0.15.0 to panic and in updating the SDK to be compatible we need v0.12.0 no longer works. So, the best choice here is to merge the code base in and retest the v0.15.0 changes.

samuelkarp:
> So, the best choice here is to merge the code base in and retest the v0.15.0 changes.

I strongly disagree with this.  The best choice would be to combine this PR and https://github.com/firecracker-microvm/firecracker-go-sdk/pull/72 so that you had a single PR (separate commits) that would successfully update for compatibility and test to ensure that it worked.

PR #72, opened 2019-02-26
syncing with v0.15.0 of Firecracker

xibz:
Two changes were required in updating to v0.15.0 in that the first changed some values to be pointers due to allowing them to have default values.
There is now validation that the drive name has to be alphanumeric and underscores. Due to this the root drive name, root-drive, has been renamed to root_drive.

Also fixes a bug where `StopVMM` was racing in `TestMicroVMExecution` and would sometimes succeed because the defer would happen last and do nothing with the error. However, if the defer happened before the `TestStopVMM`, then that's when the error is captured and fails.


samuelkarp:
Please rebase this so that it tests Firecracker v0.15.0.

samuelkarp:
`TestJailerMicroVMExecution` seems to be failing consistently:

<details>

```
=== RUN   TestJailerMicroVMExecution
time="2019-02-26T22:22:24Z" level=debug msg="Called NewMachine()"
time="2019-02-26T22:22:24Z" level=debug msg="Called Machine.Start()"
time="2019-02-26T22:22:24Z" level=debug msg="Marking Machine as Started"
time="2019-02-26T22:22:24Z" level=debug msg="Running handler validate.JailerCfg"
time="2019-02-26T22:22:24Z" level=debug msg="Running handler fcinit.StartVMM"
time="2019-02-26T22:22:24Z" level=info msg="Called startVMM(), setting up a VMM on /tmp/jailer-test221159510/firecracker/b/api.socket"
time="2019-02-26T22:22:24Z" level=debug msg="VMM started socket path is /tmp/jailer-test221159510/firecracker/b/api.socket"
time="2019-02-26T22:22:24Z" level=debug msg="Setting up signal handler"
time="2019-02-26T22:22:27Z" level=warning msg="Failed handler \"fcinit.StartVMM\": Firecracker did not create API socket /tmp/jailer-test221159510/firecracker/b/api.socket: context deadline exceeded"
--- FAIL: TestJailerMicroVMExecution (3.04s)
    machine_test.go:246: Failed to start VMM: Firecracker did not create API socket /tmp/jailer-test221159510/firecracker/b/api.socket: context deadline exceeded
```
</details>

Note that https://github.com/firecracker-microvm/firecracker-go-sdk/issues/73 is making retries more difficult to run in CI.

xibz:
Fixed

machine_test.go
```diff
@@ -333,7 +334,6 @@ func TestMicroVMExecution(t *testing.T) {
 	case <-exitchannel:
 		// if we've already exited, there's no use waiting for the timer
 	}
-	t.Run("TestStopVMM", func(t *testing.T) { testStopVMM(ctx, t, m) })
```
samuelkarp: What's the rationale for removing this test?

machine_test.go
```diff
@@ -353,9 +353,7 @@ func TestStartVMM(t *testing.T) {
 		t.Fatalf("failed to create new machine: %v", err)
 	}
 
-	defer m.StopVMM()
```
samuelkarp: What's the rationale for removing this?

machine_test.go
```diff
@@ -396,7 +394,6 @@ func TestStartVMMOnce(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	defer m.StopVMM()
```
samuelkarp: What's the rationale for removing this?

xibz: Calling `NewMachine` won't start the `VMM`, so no need to call `StopVMM`.

machine_test.go
xibz: isn't removing it, moving it above that way it occurs before exitchannel. 

Also fixes a bug where `StopVMM` was racing in `TestMicroVMExecution` and would sometimes succeed because the defer would happen last and do nothing with the error. However, if the defer happened before the `TestStopVMM`, then that's when the error is captured and fails.

version.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package firecracker
+
+// Version represents the current version of the SDK.
+const Version = "v0.15.0"
```
mxpv: I think we don't need `v` here. [Semver](https://semver.org/) expects `MAJOR.MINOR.PATCH` format.

xibz: Yep! Good catch

machine_test.go
samuelkarp: The point of having `TestStopVMM` in the current location is so that the VM does start up completely before we try to stop it; see line 330-336.  Do you believe that we no longer need to test this scenario?

xibz: The other option is having the test run in a `defer` and removing `defer StopVMM()` as well. That would keep to ensure that testing shutdown of a fulling started VM is still there. Thoughts?

Issue #79, opened 2019-02-28
Test each individual handler


xibz:
In the `handlers.go` file there are predefined handlers that need to be tested, and since some handlers are optionally ran they may not be tested in our current integration tests. By writing tests for each defined handler allows certainty that the handler behave correctly.

PR #80, opened 2019-02-28
Adds handler tests, removes Metadata field, and removes machine_mock.go

xibz:
This PR is broken into three commits.

1) Adds the handlers tests to test each individual handler API call is actually called and has proper values
2) Removes the Metadata field which isn't used anywhere
3) Removes the machine_mock.go file as that is no longer used.

Addresses #79 


machine.go
```diff
@@ -131,9 +131,6 @@ func (cfg *Config) Validate() error {
 
 // Machine is the main object for manipulating Firecracker microVMs
 type Machine struct {
-	// Metadata is the associated metadata that will be sent to the firecracker
-	// process
-	Metadata interface{}
```
nmeyerhans: Alternatively, have we considered actually storing the user-provided metadata here? There could be use cases having it available in this struct. But considering that metadata can be used to pass sensitive data to the VM, it may also be reasonable for it to be write-only from the SDK's perspective. I just want to make sure we're fully aware of what we're doing before we make this change.

xibz: Yea, I thought about it some but I can't really think of a good use case where it makes sense. For instance there are two method, `EnableMetadata` and `SetMetadata` where `EnableMetadata` will inject the metadata handler into the machine's handler stack so when `Start` is called it'll automatically call that API. Then you have `SetMetadata` which calls the API directly. I think with the two methods we cover what we need.

And the concerns with it only being writable is covered by the two operations. I think with that said, removing it makes the most sense.

nmeyerhans: > Yea, I thought about it some but I can't really think of a good use case where it makes sense. For instance there are two method, `EnableMetadata` and `SetMetadata` where `EnableMetadata` will inject the metadata handler into the machine's handler stack so when `Start` is called it'll automatically call that API. Then you have `SetMetadata` which calls the API directly. I think with the two methods we cover what we need.

How so? None of those options facilitate the retrieval of the metadata from within the SDK, which was basically what my question was about.

xibz: > How so? None of those options facilitate the retrieval of the metadata from within the SDK, which was basically what my question was about.

Ah, sorry. Yea, I think that makes sense, but shouldn't the `Metadata` field be added when we add `GET` operations to metadata? Rather than having it do nothing currently? That seems a little confusing. The SDK currently has no way of getting the metadata.

nmeyerhans: GET from the API is probably the right place to grab this.

PR #83, opened 2019-03-06
Adding version test

xibz:
This test will ensure that the CHANGELOG.MD version and the version
specified in the version.go are matching.

### DIFF
--- a/version_test.go
+++ b/version_test.go
@@ -1,0 +1,32 @@
+package firecracker
+
+import (
+	"io/ioutil"
+	"regexp"
+	"testing"
+)
+
+func TestVersionAndChangelogSync(t *testing.T) {
+	const changelogFilename = "CHANGELOG.md"
+	b, err := ioutil.ReadFile(changelogFilename)
+	if err != nil {
+		t.Fatalf("Failed to read %s: %v", changelogFilename, err)
+	}
+
+	clVersion := getVersion(b)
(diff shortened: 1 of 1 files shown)


samuelkarp:
I don't really know that this is precisely what I want.  I've been thinking about changing version.go after a release to be x.y.z+1-dev (i.e., after the 0.15.1 release changing it to 0.15.2-dev) indicating that it's a semver prerelease (it's not precisely 0.15.1 since changes were made, and it's not yet 0.15.2).

I do want to say that 1511111 is a very good git commit short hash though.

xibz:
@samuelkarp - haha didn't even notice the hash. Okay, that sounds like a plan. I was worried about potentially forgetting to update the version.go file for a release. So I wanted to add some sort of validation there. I'll go ahead and close the PR.

mxpv:
My thought on this:
- It feel like it should be a CI step rather than a unit test
- This makes changelog format pretty strict (worth to update docs?), for example firecracker has [different](https://raw.githubusercontent.com/firecracker-microvm/firecracker/master/CHANGELOG.md) format.
- I bet there are bunch of existing tools for changelog verification.

samuelkarp:
> This makes changelog format pretty strict (worth to update docs?), for example firecracker has different format.

I like their "Unreleased" section at the top and would want to do something like that for us as well.

PR #86, opened 2019-03-07
Updating jailer to be compatible with v0.16.0

### LINKED ISSUE #92: How to use metrics and logs fifo with the jailer?
Whatever I do, I can't seem to get logs and metrics to work with the SDK. They work fine if I use the API manually.

I get the following:

```
WARN[0004] setupLogging() returned [PUT /logger][400] putLoggerBadRequest  &{FaultMessage:Could not open logging fifo: Failed to open pipe. Error: entity not found}. Continuing anyway.
```

The jailer requires putting files in its `chrootBasePath + "/root"`, which makes sense. I don't think this is working with the SDK.

While I'm here, what's the difference between using `EnableJailer` and using `WithProcessRunner(jailer)`?

Example program:

```go
	ctx := context.Background()
	vmmCtx, vmmCancel := context.WithCancel(ctx)
	defer vmmCancel()
	devices := []models.Drive{}
	rootDrive := models.Drive{
		DriveID:      firecracker.String("1"),
		PathOnHost:   &"./rootfs",
		IsRootDevice: firecracker.Bool(true),
		IsReadOnly:   firecracker.Bool(false),
	}
	devices = append(devices, rootDrive)
	jailer := firecracker.NewJailerCommandBuilder().
		WithID("my-test-id").
		WithUID(123).
		WithGID(100).
		WithNumaNode(0).
		WithExecFile("./firecracker").
		WithChrootBaseDir("./jails").
		WithStdin(os.Stdin).
		WithStdout(os.Stdout).
		WithStderr(os.Stderr).
		Build(vmmCtx)

	fcCfg := firecracker.Config{
		SocketPath:      "./jails/firecracker/my-test-id/api.socket",
		KernelImagePath: "./hello-vmlinux.bin",
		KernelArgs:      "console=ttyS0 reboot=k panic=1 pci=off",
		Drives:          devices,
		LogFifo:         "./logs.fifo",
		MetricsFifo:     "./metrics.fifo",
		LogLevel:        "Info",
		NetworkInterfaces: []firecracker.NetworkInterface{firecracker.NetworkInterface{
			MacAddress:  "AA:FC:00:00:00:01",
			HostDevName: "tap0",
		}},
		MachineCfg: models.MachineConfiguration{
			VcpuCount:  2,
			HtEnabled:  true,
			MemSizeMib: 256,
		},
	}

	log.Println("Creating VM Machine")
	m, err := firecracker.NewMachine(vmmCtx, fcCfg, firecracker.WithLogger(log.NewEntry(logger)), firecracker.WithProcessRunner(jailer))
	if err != nil {
		return err
	}

	if err := m.Start(vmmCtx); err != nil {
		return err
	}
	defer m.StopVMM()

	// wait for the VMM to exit
	if err := m.Wait(vmmCtx); err != nil {
		return err
	}
	log.Printf("Start machine was happy")
	return nil
```

(this code might be incomplete)

Yields the following logs:

```
INFO[0003] Creating VM Machine
DEBU[0003] Called NewMachine()
DEBU[0003] Called Machine.Start()
DEBU[0003] Marking Machine as Started
DEBU[0003] Running handler validate.Cfg
DEBU[0003] Running handler fcinit.StartVMM
INFO[0003] Called startVMM(), setting up a VMM on ./jails/firecracker/my-test-id/api.socket
DEBU[0003] VMM started socket path is ./jails/firecracker/my-test-id/api.socket
DEBU[0003] Setting up signal handler
DEBU[0004] returning from startVMM()
DEBU[0004] Running handler fcinit.BootstrapLogging
DEBU[0004] Created metrics and logging fifos.
2019-04-08T19:04:31.865532096 [my-test-id:ERROR] Received Error on synchronous Put request on "/logger" with body "{\"level\":\"Info\",\"log_fifo\":\"./logs.fifo\",\"metrics_fifo\":\"./metrics.fifo\",\"options\":[],\"show_level\":true,\"show_log_origin\":false}\n". Status code: 400 Bad Request.
WARN[0004] setupLogging() returned [PUT /logger][400] putLoggerBadRequest  &{FaultMessage:Could not open logging fifo: Failed to open pipe. Error: entity not found}. Continuing anyway.
DEBU[0004] Running handler fcinit.CreateMachine
```

I tried various combinations of manually creating a remove the fifos, but didn't have any luck. I can't control much of the flow with the SDK.

xibz:
Fixes #92 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # [ Unreleased ]
 
-* Fixes bug where default socketpath would always be used when not using jailer (#84)
+* Fixes bug where context was not being used at all during startVM (#86)
+* Updates the jailer's socket path to point to the unix socket in the jailer's workspace (#86)
+* Fixes bug where default socketpath would always be used when not using jailer (#84).
 
 # 0.15.1
 
--- a/jailer.go
+++ b/jailer.go
@@ -94,6 +94,8 @@ type JailerConfig struct {
 	Stdout io.Writer
 	// Stderr specifies the IO writer for STDERR to use when spawning the jailer.
 	Stderr io.Writer
+	// Stdin specifies the IO reader for STDIN to use when spawning the jailer.
+	Stdin io.Reader
 }
 
 // JailerCommandBuilder will build a jailer command. This can be used to
@@ -288,14 +290,14 @@ func (b JailerCommandBuilder) Build(ctx context.Context) *exec.Cmd {
 // Jail will set up proper handlers and remove configuration validation due to
 // stating of files
 func jail(ctx context.Context, m *Machine, cfg *Config) error {
-	rootfs := ""
+	jailerWorkspaceDir := ""
 	if len(cfg.JailerCfg.ChrootBaseDir) > 0 {
-		rootfs = filepath.Join(cfg.JailerCfg.ChrootBaseDir, "firecracker", cfg.JailerCfg.ID)
+		jailerWorkspaceDir = filepath.Join(cfg.JailerCfg.ChrootBaseDir, "firecracker", cfg.JailerCfg.ID, rootfsFolderName)
 	} else {
-		rootfs = filepath.Join(defaultJailerPath, cfg.JailerCfg.ID)
+		jailerWorkspaceDir = filepath.Join(defaultJailerPath, cfg.JailerCfg.ID, rootfsFolderName)
 	}
 
-	cfg.SocketPath = filepath.Join(rootfs, "api.socket")
+	cfg.SocketPath = filepath.Join(jailerWorkspaceDir, "api.socket")
 
 	stdout := cfg.JailerCfg.Stdout
 	if stdout == nil {
@@ -305,6 +307,11 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 	stderr := cfg.JailerCfg.Stderr
 	if stderr == nil {
 		stderr = os.Stderr
+	}
+
+	stdin := cfg.JailerCfg.Stdin
+	if stdin == nil {
+		stdin = os.Stdin
 	}
 
 	m.cmd = NewJailerCommandBuilder().
@@ -318,6 +325,7 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 		WithSeccompLevel(cfg.JailerCfg.SeccompLevel).
 		WithStdout(stdout).
 		WithStderr(stderr).
+		WithStdin(stdin).
 		Build(ctx)
 
 	if err := cfg.JailerCfg.ChrootStrategy.AdaptHandlers(&m.Handlers); err != nil {
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -45,8 +45,8 @@ steps:
   - label: ':hammer: root tests'
     commands:
       - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
-      - 'cp /usr/local/bin/firecracker-v0.15.0 testdata/firecracker'
-      - 'cp /usr/local/bin/jailer-v0.15.0 testdata/jailer'
+      - 'cp /usr/local/bin/firecracker-v0.16.0 testdata/firecracker'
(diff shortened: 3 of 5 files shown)


xibz:
Test will fail until `v0.16.0` of firecracker is added

samuelkarp:
> Test will fail until v0.16.0 of firecracker is added to the CI system

Let me know if you need help with this.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # [ Unreleased ]
 
-* Fixes bug where default socketpath would always be used when not using jailer (#84)
+* Updates jailer's socket path to point to the root drive.
```
nmeyerhans: This does not make sense. Socket path should not point to a root drive.

jailer.go
```diff
@@ -318,6 +318,9 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 		WithSeccompLevel(cfg.JailerCfg.SeccompLevel).
 		WithStdout(stdout).
 		WithStderr(stderr).
+		WithStdout(os.Stdout).
```
nmeyerhans: Shouldn't this replace the WithStdout call on line 319?

jailer.go
```diff
@@ -318,6 +318,9 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
...
+		WithStderr(os.Stderr).
```
nmeyerhans: We're already calling WithStderr on line 320.

jailer.go
```diff
@@ -290,9 +290,9 @@ func (b JailerCommandBuilder) Build(ctx context.Context) *exec.Cmd {
 func jail(ctx context.Context, m *Machine, cfg *Config) error {
 	rootfs := ""
 	if len(cfg.JailerCfg.ChrootBaseDir) > 0 {
-		rootfs = filepath.Join(cfg.JailerCfg.ChrootBaseDir, "firecracker", cfg.JailerCfg.ID)
+		rootfs = filepath.Join(cfg.JailerCfg.ChrootBaseDir, "firecracker", cfg.JailerCfg.ID, rootfsFolderName)
```
nmeyerhans: It's only tangentially related to the change you're making here, but I think `rootfsFolderName` is the wrong name for this variable. It implies that the contents of this directory make up a root filesystem, but that's not the case. It's where we link the kernel and root drive image, but that's not the same thing a rootfs.

Feel free to change this here if you want, otherwise we should do this in a later PR.

CHANGELOG.md
xibz: I'll reword to `Updates the jailer's socket path to point to the socket path in the jailer's workspace`

jailer.go
xibz: Ah this was a rebase issue. I'll go ahead and fix that.

jailer.go
xibz: Same here

jailer.go
xibz: I'll go ahead and change it. Maybe `jailerWorkspacePath`?

nmeyerhans: How about `jailerWorkspaceDir`?

jeromegn: This confused me a lot when I was trying to figure out how things worked. I passed in the wrong path entirely to the NaiveChrootStrategy

jailer.go
jeromegn: Can we add support for Stdin while we're at it?

jeromegn:
The tests are not passing because logging is setup before the linking handler. The logging handler also attempts to `client.PutLogger` which means it'll send the request to the socket before the logs and metrics fifos are linked in the root directory.

The `BootstrapLoggingHandler` handler likely needs to be refactored in 2 steps where one creates the fifos and the other one sends the request. The NaiveChrootStrategy can then inject a handler between the 2 to link logs files properly.

jeromegn:
I've addressed some of my concerns in #96

xibz:
@jeromegn - I've went ahead and removed the fix for fifos here. It seems like there is an issue with the jailer and put logger API which causes linking of the fifos to fail. I opened an issue [here](https://github.com/firecracker-microvm/firecracker/issues/1100)

PR #87, opened 2019-03-09
Adding v0.15.0 firecracker docker image

xibz:
Adds a docker file that allows for testing of v0.15.0 of Firecracker.


fctesting/sandbox/fc-latest
```diff
@@ -0,0 +1,68 @@
...
+
+# Install the Rust toolchain
+#
+# Install 1.32.0 due to Firecracker crashing with 148 code
+RUN mkdir "$TMP_BUILD_DIR" \
+    && curl https://sh.rustup.rs -sSf | sh -s -- -y \
+    	&& rustup install 1.32.0 \
+	&& rustup override set 1.32.0 \
+        && rustup target add x86_64-unknown-linux-musl \
+    && cd /
```
mxpv: I suggest utilizing docker's [multi-stage](https://docs.docker.com/develop/develop-images/multistage-build/) builds here. I think this will allow to drop rust installation and significantly reduce apt-get install deps.

Something like:
```Dockerfile
FROM rust:latest as fc-build

# Build FC
RUN cargo build --release --features vsock --target-dir "$TMP_BUILD_DIR"
...

FROM go:1.12
...
COPY --from  fc-build firecracker-binary . # Copy FC binary from previous stage
...
```

xibz: Does the `rust:latest` have musl with it? And we don't want latest as it doesn't work with firecracker, we want `1.32`.

mxpv: I'm not sure about `musl`.
As for versions, you can use `rust:1.32`

xibz: Yea, that's why rustup was used to get musl. rustup will automatically grab rust and the like, so I don't think adding the `FROM rust:1.32` will save anything, cause we'd still need to grab rustup and install musl and rebuild.

nmeyerhans: The rustup commands from the existing dockerfile will install musl and 1.32 as the defaults.

Makefile
```diff
@@ -27,4 +27,15 @@ all-tests:
 generate build clean:
 	go $@ $(EXTRAGOARGS)
 
-.PHONY: all generate clean build test unit-tests all-tests
+sandbox-test-fc:
```
nmeyerhans: This rule does two things: Build a container image and run it. It would be best to split this into distinct rules for each task. You might want to use a "stamp" file to avoid having the image creation rule be PHONY. Since it does actually create an artifact, this makes sense (the artifact being the image stored within Docker, rather than a file on the filesystem)

fctesting/sandbox/fc-latest
mxpv: > cause we'd still need to grab rustup and install musl

`rust:1.32` [already contains](https://github.com/rust-lang/docker-rust/blob/97a72441aad0ec26edf5175382c3d0022c84ba7d/Dockerfile-debian.template#L19) `rustc` and `rustup`
`rustup target add x86_64-unknown-linux-musl` adds `rust-std` component.
<details>

% rustup -v target add x86_64-unknown-linux-musl   
verbose: read metadata version: '12'
info: downloading component 'rust-std' for 'x86_64-unknown-linux-musl'
verbose: downloading file from: 'https://static.rust-lang.org/dist/2019-01-17/rust-std-1.32.0-x86_64-unknown-linux-musl.tar.xz'
verbose: downloading with curl
verbose: checksum passed
info: installing component 'rust-std' for 'x86_64-unknown-linux-musl'
verbose: creating temp directory: ...

</details>

xibz: @mxpv - Ah cool! Will use that then :)

xibz: Ah, so apparently the multistage builds are two separate containers. I need them to be the same since the binaries from the first stage needs to be in the build directory for the second stage. So, Im going to keep this as is.

mxpv: This is where you use `COPY --from prev_stage_name . . ` to transfer artifacts between stages.

Makefile
xibz: > You might want to use a "stamp" file to avoid having the image creation rule be PHONY. Since it does actually create an artifact, this makes sense (the artifact being the image stored within Docker, rather than a file on the filesystem)

What exactly is a stamp file?

fctesting/sandbox/fc-latest
xibz: Okay, I separated the steps out which I think is nicer. However, I still needed to build the toolchain since no toolchain was provided ><

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        musl-tools
+
+# Install the Rust toolchain
+#
+# Install 1.32.0 due to Firecracker crashing with 148 code
+RUN mkdir "$TMP_BUILD_DIR" \
+    	&& rustup install 1.32.0 \
+	&& rustup override set 1.32.0 \
```
mxpv: I think these 2 lines (install/override) are no longer needed.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+ENV FC_TEST_DATA_PATH="$TMP_BUILD_DIR"
+ARG FIRECRACKER_GO_SDK_SRC_DIR="/firecracker-go-sdk"
+ADD . /firecracker-go-sdk/
+COPY --from=build "$TMP_BUILD_DIR" "$TMP_BUILD_DIR"
+
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        binutils-dev \
+        git \
+        make
```
mxpv: `golang:1.12` already has `git` and `make`. I suspect binutils are also no longer needed. Hence can we drop this entire `RUN`?

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+FROM rust:1.32 as build
+
+ARG TMP_BUILD_DIR
+ARG FIRECRACKER_SRC_DIR="/firecracker"
+ARG FIRECRACKER_BUILD_DIR="$FIRECRACKER_SRC_DIR/build"
+ARG CARGO_REGISTRY_DIR="$FIRECRACKER_BUILD_DIR/cargo_registry"
+ARG FIRECRACKER_VERSION="0.15.0"
+
+ENV CARGO_HOME=/usr/local/rust
+ENV RUSTUP_HOME=/usr/local/rust
```
mxpv: Same here. as we use `rust:1.32`, these lines are no longer needed.

Makefile
nmeyerhans: > What exactly is a stamp file?

It's a technique to avoid having to declare a target PHONY. Make generally assumes that, after the execution of the recipes associated with a target, a file will exist with a name matching the name of the target. Some rules, though, don't create files, but have some other side-effect (e.g. yours, with a side-effect that there's now a new image registered with Docker).

A stamp file is a file that you can create upon successful creation of a rule that would otherwise be PHONY. Its name matches that of the target, so from make's perspective, the target was successfully rebuild and doesn't need to get rebuilt. So, if you consider the following makefile:

```
do_phony_stuff:
	@echo I ran, and I will keep running.
	# Here I might do some time consuming activity

do_stamp_stuff:
	@echo I ran, but only once
	# Here I might do some time consuming activity
	@touch do_stamp_stuff

clean:
	-rm do_stamp_stuff

.PHONY: do_phony_stuff
```

There are two targets that are basically the same. One is phony, one uses a stamp target. If you run the phony target multiple times, it is executed multiple times:
```
$ make do_phony_stuff
I ran, and I will keep running.
# Here I might do some time consuming activity
$ make do_phony_stuff
I ran, and I will keep running.
# Here I might do some time consuming activity
$ make do_phony_stuff
I ran, and I will keep running.
# Here I might do some time consuming activity
$ make do_phony_stuff
I ran, and I will keep running.
# Here I might do some time consuming activity
$ make do_phony_stuff
I ran, and I will keep running.
# Here I might do some time consuming activity
```

Whereas the target that uses a stamp only runs once:
```
$ make do_stamp_stuff
I ran, but only once
# Here I might do some time consuming activity
$ make do_stamp_stuff
make: 'do_stamp_stuff' is up to date.
$ make do_stamp_stuff
make: 'do_stamp_stuff' is up to date.
$ make do_stamp_stuff
make: 'do_stamp_stuff' is up to date.
```

Of course, dependencies can cause it to be rebuilt, as can the non-presense of the stamp file (which is why you should be sure the remove it in the `clean` recipe)

Because Docker caches the result of builds, this isn't super important here. The first call to `docker build` may be time consuming, but subsequent calls will be fairly fast. Still, this is a good practice to be employ in general.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+WORKDIR firecracker
+RUN git checkout tags/v"$FIRECRACKER_VERSION"
+RUN cargo build --release --features vsock --target-dir "$TMP_BUILD_DIR"
+RUN cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/firecracker "$TMP_BUILD_DIR"
+RUN cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/jailer "$TMP_BUILD_DIR"
+
+FROM golang:1.12
+
+ARG TMP_BUILD_DIR
+ENV FC_TEST_DATA_PATH="$TMP_BUILD_DIR"
```
mxpv: `TMP_BUILD_DIR` and `FC_TEST_DATA_PATH` are defined above and can be reused.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+ARG FIRECRACKER_GO_SDK_SRC_DIR="/firecracker-go-sdk"
+ADD . /firecracker-go-sdk/
+COPY --from=build "$TMP_BUILD_DIR" "$TMP_BUILD_DIR"
+
+WORKDIR "$FIRECRACKER_GO_SDK_SRC_DIR"
+RUN cp -rf testdata/* "$TMP_BUILD_DIR"
+RUN curl -fsSL -o "$TMP_BUILD_DIR"/vmlinux https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin
```
sipsma: You can use `ADD https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin <destination inside container>`, which will result in the build checking to see if the remote file changed and re-download it if so.

Makefile
```diff
@@ -27,4 +27,19 @@ all-tests:
...
+	docker build -f fctesting/sandbox/Dockerfile -t "firecracker" .
+
+sandbox-test-fc-run:
+	docker run \
+		--init \
+		--rm \
+		--privileged \
+		--security-opt seccomp=unconfined \
+		--ulimit core=0 \
+		--device=/dev/kvm:/dev/kvm \
```
sipsma: Are you sure this `--device` is needed? I just learned recently that when you use `--privileged`, Docker will automatically create all the device nodes present under the host's `/dev` inside the container ([docs](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities))

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        musl-tools
+
+# Add musl to rust std
+RUN mkdir "$TMP_BUILD_DIR" \
+    	&& rustup install 1.32.0 \
+	&& rustup override set 1.32.0 \
+        && rustup target add x86_64-unknown-linux-musl \
+    && cd /
```
sipsma: nit: I don't believe the `cd /` is needed here, I believe each `RUN` command starts out in the dir specified by `WORKDIR` in the Dockerfile, so you shouldn't need to change back to `/` at the end of a `RUN`

Makefile
```diff
@@ -27,4 +27,20 @@ all-tests:
 generate build clean:
 	go $@ $(EXTRAGOARGS)
 
+sandbox-test-fc-build:
+	docker build -f fctesting/sandbox/Dockerfile -t "firecracker" .
+	@touch sandbox-test-fc-build
+
+sandbox-test-fc-run:
+	docker run \
```
samuelkarp: Given the options here to grant pretty much all privileges and run as root, I'd like the make target to fail when invoked as a non-root user.  You should be able to do this by inserting `test "$(shell id --user)" -eq 0` above the `docker run` line.

Makefile
```diff
@@ -27,4 +27,20 @@ all-tests:
 generate build clean:
 	go $@ $(EXTRAGOARGS)
 
+sandbox-test-fc-build:
+	docker build -f fctesting/sandbox/Dockerfile -t "firecracker" .
```
samuelkarp: Let's not call the image "firecracker".  How about "localhost/firecracker-go-sdk-sandbox`?

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG TMP_BUILD_DIR=/tmp/build
+
+FROM rust:1.32 as build
+
+ARG TMP_BUILD_DIR
```
samuelkarp: Did you mean to declare this `ARG` a second time?

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG FIRECRACKER_VERSION="0.15.0"
+
+# Install system dependecies
+#
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        musl-tools
+
+# Add musl to rust std
+RUN mkdir "$TMP_BUILD_DIR" \
```
samuelkarp: Why do you need to use "$TMP_BUILD_DIR"?  Did you want to make this variable instead of picking a constant path?

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG TMP_BUILD_DIR=/tmp/build
+
+FROM rust:1.32 as build
```
samuelkarp: I'd pick "as firecracker-build" instead of "as build" since this Dockerfile describes multiple build environments.

fctesting/sandbox/Dockerfile
samuelkarp: `ADD` has some surprising behavior around auto-extraction of tarballs and always creating a separate layer, so I don't really have a problem avoiding it.  See [this best-practices document](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) for some details.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG TMP_BUILD_DIR
+ENV FC_TEST_DATA_PATH="$TMP_BUILD_DIR"
+ARG FIRECRACKER_GO_SDK_SRC_DIR="/firecracker-go-sdk"
+ADD . /firecracker-go-sdk/
+COPY --from=build "$TMP_BUILD_DIR" "$TMP_BUILD_DIR"
+
+WORKDIR "$FIRECRACKER_GO_SDK_SRC_DIR"
+RUN cp -rf testdata/* "$TMP_BUILD_DIR"
+RUN curl -fsSL -o "$TMP_BUILD_DIR"/vmlinux https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin
+RUN go mod download
```
samuelkarp: I assume you have this here to aid in cache-reuse?  Since the `docker build` doesn't depend (in the Makefile) on source files changing, this may not be very effective; new dependencies can be added and not be cached in the image.  I think there are a few options instead:

* Update the Makefile to express dependencies on the source code (or at *least* the `go.mod` and `go.sum` files) so that we automatically rebuild the image when those change
* Avoid this step in the image and instead mount in the cache directory (and probably copy it, so we don't mess with the user's actual cache directory) during the `docker run` step

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN git checkout tags/v"$FIRECRACKER_VERSION"
+RUN cargo build --release --features vsock --target-dir "$TMP_BUILD_DIR"
+RUN cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/firecracker "$TMP_BUILD_DIR"
+RUN cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/jailer "$TMP_BUILD_DIR"
+
+FROM golang:1.12
+
+ARG TMP_BUILD_DIR
+ENV FC_TEST_DATA_PATH="$TMP_BUILD_DIR"
+ARG FIRECRACKER_GO_SDK_SRC_DIR="/firecracker-go-sdk"
```
samuelkarp: What's the `ARG` for?  Do you expect this to be overridden?

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ADD . /firecracker-go-sdk/
```
samuelkarp: This should probably be close to the last step in the image, so that we can get as much layer cache reuse from the Docker image as possible.  If you want to do caching of the Go dependencies (see my other comment), you could structure it like this:

```Dockerfile
COPY ["go.mod", "go.sum", "/firecracker-go-sdk/"]
RUN go mod download
COPY [".", "/firecracker-go-sdk/"]
```

Makefile
xibz: @sipsma - No idea, I copied this from firecracker :D. I can remove it and see if it still works.

fctesting/sandbox/Dockerfile
xibz: Yea, if I don't say that variable is in this scope it doesn't exist. I hope that makes sense.

fctesting/sandbox/Dockerfile
xibz: Apparently I have to reference it with `ARG TMP_BUILD_DIR` or I get an empty value

fctesting/sandbox/Dockerfile
xibz: Naw, it could be a constant path, just thought `ENV` were specifically for environment variables and didn't know what is considered to be a constant other than `ENV`. The `TMP_BUILD_DIR` is used as the test folder where it searches for firecracker, jailer, kernel image, and root files.

fctesting/sandbox/Dockerfile
xibz: No, don't expect these to be overridden. Is Docker's equivalent of constant the `ENV` keyword?

fctesting/sandbox/Dockerfile
xibz: Ah, that is my misunderstanding on how docker uses that. I will update.

fctesting/sandbox/Dockerfile
samuelkarp: From the [documentation](https://docs.docker.com/engine/reference/builder/#arg):

> The `ARG` instruction defines a variable that users can pass at build-time to the builder with the `docker build` command using the `--build-arg <varname>=<value>` flag. If a user specifies a build argument that was not defined in the Dockerfile, the build outputs a warning.

If you're not using the `ARG` command to allow overrides with `--build-arg`, please replace it with `ENV`.

fctesting/sandbox/Dockerfile
samuelkarp: From the [documentation](https://docs.docker.com/engine/reference/builder/#arg):

> The `ARG` instruction defines a variable that users can pass at build-time to the builder with the `docker build` command using the `--build-arg <varname>=<value>` flag. If a user specifies a build argument that was not defined in the Dockerfile, the build outputs a warning.

If you're not using the `ARG` command to allow overrides with `--build-arg`, please replace it with `ENV`.

fctesting/sandbox/Dockerfile
samuelkarp: I would probably just write the literal `/build` instead of using an environment variable.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG TMP_BUILD_DIR=/tmp/build
+
+FROM rust:1.32 as firecracker-build
+
+ARG TMP_BUILD_DIR
+ARG FIRECRACKER_SRC_DIR="/firecracker"
+ARG FIRECRACKER_BUILD_DIR="$FIRECRACKER_SRC_DIR/build"
+ARG CARGO_REGISTRY_DIR="$FIRECRACKER_BUILD_DIR/cargo_registry"
+ARG FIRECRACKER_VERSION="0.15.0"
```
samuelkarp: While I think you want to get rid of most of the `ARG` statements, this one makes a lot of sense to keep.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        musl-tools
+
+# Add musl to rust std
+RUN mkdir "$TMP_BUILD_DIR" \
+    	&& rustup install 1.32.0 \
+	&& rustup override set 1.32.0 \
+        && rustup target add x86_64-unknown-linux-musl
+
+# Install Firecracker and set binaries to TMP_BUILD_DIR
+RUN git clone https://github.com/firecracker-microvm/firecracker.git
```
samuelkarp: Is there any particular reason you have these as separate `RUN` statements instead of chaining them together?  In general, each directive (like `RUN`, `ENV`, etc) adds a new layer to the resulting image.  When you're building images, you generally want to reduce the total number of layers except when you want to share layers between disparate images (think a base OS or language image).  This `firecracker-build` one is internal only and you never really push it as an artifact, but introducing separate layers still has a performance impact during the build as you cause Docker's storage driver (layer storage component) to create new snapshots for each one.

I'd rewrite like this:

```Dockerfile
RUN git clone https://github.com/firecracker-microvm/firecracker.git \
    cd firecracker \
    git checkout tags/v"$FIRECRACKER_VERSION" \
    cargo build --release --features vsock --target-dir /build \
    ...
```

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG TMP_BUILD_DIR=/build
```
samuelkarp: Can you remove this as `ARG`?  You're not intending for it to be an overridden argument.

fctesting/sandbox/Dockerfile
xibz: I went with using an `ENV` just due to it being easier to change if we ever do decide to move it from `/build`.

fctesting/sandbox/Dockerfile
xibz: Had no idea that having multiple RUN commands caused such overhead was the main reason. I've went ahead and updated it based on your feedback.

Makefile
```diff
@@ -27,4 +27,20 @@ all-tests:
 generate build clean:
 	go $@ $(EXTRAGOARGS)
 
+sandbox-test-fc-build:
```
nmeyerhans: Assuming the image build is something a person would ever trigger manually, I'd suggest doing the actual work in a `sandbox-test-fc-build-stamp` rule. The final step of that rule's recipe should be to touch a file of the same name. Then add a `sandbox-test-fc-build` PHONY rule that doesn't do anything but declares a dependency on the `-stamp` variant. Having `stamp` in the file name makes the file's purpose a little more apparent.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+# Install Firecracker and set binaries to TMP_BUILD_DIR
+RUN git clone https://github.com/firecracker-microvm/firecracker.git \
+        && cd firecracker \
+        && git checkout tags/v"$FIRECRACKER_VERSION" \
+        && cargo build --release --features vsock --target-dir "$TMP_BUILD_DIR" \
+        && cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/firecracker "$TMP_BUILD_DIR" \
+        && cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/jailer "$TMP_BUILD_DIR"
+
+FROM golang:1.12
```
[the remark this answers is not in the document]
xibz: I don't see why we shouldn't update. I've been using 1.12.0 for a while now, at least.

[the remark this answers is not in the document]
nmeyerhans: Works for me. 1.12 it is.

fctesting/sandbox/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+ENV TMP_BUILD_DIR="/build"
+ENV FC_TEST_DATA_PATH="$TMP_BUILD_DIR"
+ARG FIRECRACKER_GO_SDK_SRC_DIR="/firecracker-go-sdk"
+COPY --from=firecracker-build "$TMP_BUILD_DIR" "$TMP_BUILD_DIR"
+
+WORKDIR "$FIRECRACKER_GO_SDK_SRC_DIR"
+COPY ["./testdata/", "$TMP_BUILD_DIR"]
+COPY ["go.mod", "go.sum", "/firecracker-go-sdk/"]
+RUN curl -fsSL -o "$TMP_BUILD_DIR"/vmlinux https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin \
```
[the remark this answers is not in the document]
xibz: Done along with the rest of the other issues as well.

Makefile
```diff
@@ -27,4 +27,27 @@ all-tests:
 generate build clean:
```
samuelkarp: You should remove "clean" from this line.

Makefile
```diff
@@ -27,4 +27,27 @@ all-tests:
...
 	go $@ $(EXTRAGOARGS)
 
-.PHONY: all generate clean build test unit-tests all-tests
+clean:
+	go $@ $(EXTRAGOARGS)
```
samuelkarp: Since you don't have a variable argument, you can just write `go clean $(EXTRAGOARGS)`.

Makefile
```diff
@@ -27,4 +27,27 @@ all-tests:
...
+	# --force to prevent erroring on stamp file not existing
+	rm --force sandbox-test-fc-build-stamp
```
samuelkarp: You can also write `rm sandbox-test-fc-build-stamp ||:` prevent an error.  `||` is a logical or and `:` is a builtin for true.

xibz:
Closing due to inactivity and age.

PR #90, opened 2019-04-03
A few minor fixes in SDK

mxpv:
*Description of changes:*
- Don't create logger instance inside `NewMachine` if using `WithLogger` option (as it'll be overridden)
- `exitCh` wasn't closed under certain conditions
- ~Added public `ExitCh` to track VM exit outside SDK~

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -200,14 +200,9 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) 
 	m := &Machine{
 		exitCh: make(chan struct{}),
 	}
-	logger := log.New()
-
-	if cfg.Debug {
-		logger.SetLevel(log.DebugLevel)
-	}
 
 	m.Handlers = defaultHandlers
-	m.logger = log.NewEntry(logger)
+
 	if cfg.EnableJailer {
 		m.Handlers.Validation = m.Handlers.Validation.Append(JailerConfigValidationHandler)
 		if err := jail(ctx, m, &cfg); err != nil {
@@ -222,6 +217,15 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) 
 
 	for _, opt := range opts {
 		opt(m)
+	}
+
+	if m.logger == nil {
+		logger := log.New()
+		if cfg.Debug {
+			logger.SetLevel(log.DebugLevel)
+		}
+
+		m.logger = log.NewEntry(logger)
 	}
 
 	if m.client == nil {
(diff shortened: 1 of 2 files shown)


machine.go
```diff
@@ -263,6 +267,11 @@ func (m *Machine) Shutdown(ctx context.Context) error {
 	return m.sendCtrlAltDel(ctx)
 }
 
+// ExitC returns machine's exit channel
+func (m *Machine) ExitCh() <-chan struct{} {
```
samuelkarp: If you expose m.exitCh here, it can be drained outside the SDK and break `Wait`.  The channel semantics (delivery once) isn't really what we want; instead, we want a mechanism where callers can subscribe to notifications and they can be delivered without breaking the SDK's use of exitCh.

mxpv: SDK closes `exitCh` when VM exits instead of writing to it, so there is no drain.
We can have as many readers as we need, they all will be blocked until `close` call.

mxpv: Added a comment to reflect this.

machine.go
```diff
@@ -332,6 +342,11 @@ func (m *Machine) startVMM(ctx context.Context) error {
 		os.Remove(m.cfg.LogFifo)
 		os.Remove(m.cfg.MetricsFifo)
 		errCh <- err
+
+		// Notify subscribers that there will be no more values.
+		// When err is nil, two reads are performed (waitForSocket and close exitCh goroutine),
+		// second one never ends as it tries to read from empty channel.
+		close(errCh)
```
samuelkarp: It looks like line 375 can also write to this channel.  I think we need strong assurance that the channel won't be closed then, as writing to a closed channel causes a panic.

samuelkarp: Ignore me...

xibz:
Out of curiosity what is the use case for needing to expose the `exitCh`?

mxpv:
> Out of curiosity what is the use case for needing to expose the exitCh?

In fc-control I'd like to be notified when VM shuts down by someone else to do proper cleanup (like this vm id is no longer valid).

xibz:
> > Out of curiosity what is the use case for needing to expose the exitCh?
> 
> In fc-control I'd like to be notified when a VM shuts down by someone else to do proper cleanup (like this vm id is no longer valid).

The `Machine` object has a `Wait` method that takes a context. Can that not be used?

mxpv:
Actually yeah. That will do the job too.

mxpv:
I've reverted changes with `ExitCh` as `Wait` can be used instead. Other changes are still relevant I guess.

PR #91, opened 2019-04-04
Adding machine interface

xibz:
This adds a machine interface that can be used to mock operations of
machine. This interface is subject to change as it reflects the Machine
structure.

Removes `EnableMetadta` as it is confusing being along side `SetMetadata`.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -588,10 +588,15 @@ func (m *Machine) sendCtrlAltDel(ctx context.Context) error {
 	return err
 }
 
-// EnableMetadata will append or replace the metadata handler and only will be
-// called when calling the Machine Start operation.
-func (m *Machine) EnableMetadata(metadata interface{}) {
-	m.Handlers.FcInit = m.Handlers.FcInit.Swappend(NewSetMetadataHandler(metadata))
+// SetMetadata sets the machine's metadata for MDDS
+func (m *Machine) SetMetadata(ctx context.Context, metadata interface{}) error {
+	if _, err := m.client.PutMmds(ctx, metadata); err != nil {
+		m.logger.Errorf("Setting metadata: %s", err)
+		return err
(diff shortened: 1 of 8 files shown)


mxpv:
We can use `var _ MachineIface = (*Machine)(nil)` to make sure the interface is up to date.

machine.go
```diff
@@ -133,9 +133,9 @@ func (cfg *Config) Validate() error {
 type Machine struct {
 	// Handlers holds the set of handlers that are run for validation and start
 	Handlers Handlers
+	Client   *Client
```
samuelkarp: Why is client made public now?

xibz: Client needs to be public since there isn't a way to call things like `PutMmds` due to removal of `SetMetadata`. It should be public if users want to use the higher level client we've created without the need of creating one when they already have a `Machine` object.

nmeyerhans: The whole point of the the SDK is to isolate the the generated interface behind an abstraction layer. I don't think we want to be exposing these details, and I certainly don't think we want to force the user to interact directly with the generated code in order to access the complete functionality.

Also, this is a breaking change, so if we're going to remove SetMetadata, we should be careful about it.

xibz: Yea, this is a breaking change, but having two methods, `EnableMetadata` and `SetMetadata` is confusing. They aren't accessing with the generated code also. They are interacting with the client we wrote.

mxpv: What if we leave `SetMetadata` and get rid of `EnableMetadata` ? (for example we can add `initialMetadata` field to `Config` instead)

xibz: Either one is fine to get rid of, if any at all. Just depends on what other think on the breaking change. Worst case is we can deprecate `EnableMetadata` and add more clear documentation

PR #93, opened 2019-04-10
Updating jailer example to use the jailer config

xibz:
Previous jailer example led to some confusion, #92, due to not knowing the need to manually set up the proper jailer workspace. New example uses the jailer config in which allows the SDK to do all the setting up.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
+
+	log "github.com/sirupsen/logrus"
 
 	"github.com/firecracker-microvm/firecracker-go-sdk"
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
@@ -19,13 +22,6 @@ func ExampleWithProcessRunner_logging() {
 		Drives:          firecracker.NewDrivesBuilder("/path/to/rootfs").Build(),
 		MachineCfg: models.MachineConfiguration{
 			VcpuCount: firecracker.Int64(1),
-		},
-		JailerCfg: firecracker.JailerConfig{
-			GID:      firecracker.Int(100),
-			UID:      firecracker.Int(100),
-			ID:       "my-micro-vm",
-			NumaNode: firecracker.Int(0),
-			ExecFile: "/path/to/firecracker",
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


jeromegn:
@xibz are you sure this is an accurate example?

Looking at the test code, it looks like much more is involved: https://github.com/firecracker-microvm/firecracker-go-sdk/blob/5ac6d3f23ba4b362def6a064d79ad2812cb5de57/machine_test.go#L95-L249

- Copying the rootfs, vmlinux, etc. in the appropriate places.
- Setting the correct UID and GID on these files.
- Creating the log files (again, approriate UID, GID)
- Configure the jailer via JailerCommandBuilder

Your example appears to show that this is not needed, but it won't quite work without *some* of the additional steps.

You do need `EnableJailer` on the config though! Else, it won't start anything (afaict)

xibz:
@jeromegn - Thanks for noticing the missing `EnableJailer`! I've went ahead and also included some logic to do some validation. The `NaiveChrootStrategy` handles copying the rootfs and kernel image to the correct places. The reason why the tests does this is to create a new workspace for the test. The `JailerCommandBuilder` is not needed when setting the `EnableJailer` flag and specifying the proper jailer configuration.

example_test.go
```diff
@@ -220,51 +224,86 @@ func ExampleNetworkInterface_rateLimiting() {
...
+			ChrootBaseDir:  path,
+			ChrootStrategy: firecracker.NewNaiveChrootStrategy(pathToWorkspace, kernelImagePath),
 		},
-		DisableValidation: true,
 	}
 
-	// Passes the custom jailer command into the constructor
-	m, err := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(b.Build(ctx)))
+	// This will stat the kernel image and take a look at its GID and UID. If the
+	// GID and UID do not matach the GID and UID provided in the configuration,
```
samuelkarp: s/matach/match/

example_test.go
```diff
@@ -220,51 +224,86 @@ func ExampleNetworkInterface_rateLimiting() {
...
+			NumaNode:       firecracker.Int(0),
+			ChrootBaseDir:  path,
+			ChrootStrategy: firecracker.NewNaiveChrootStrategy(pathToWorkspace, kernelImagePath),
 		},
-		DisableValidation: true,
 	}
 
-	// Passes the custom jailer command into the constructor
-	m, err := firecracker.NewMachine(ctx, cfg, firecracker.WithProcessRunner(b.Build(ctx)))
+	// This will stat the kernel image and take a look at its GID and UID. If the
```
samuelkarp: Do we really require the UID/GID to match, or just require that the target UID/GID has permission to read the kernel image?

example_test.go
```diff
@@ -220,51 +224,86 @@ func ExampleNetworkInterface_rateLimiting() {
...
+	}
+
+	if kernelImageInfo.Uid != uint32(uid) || kernelImageInfo.Gid != uint32(gid) {
+		panic(fmt.Errorf("Kernel image does not have the proper UID or GID.\n"+
+			"To fix this simply run:\n"+
+			"sudo chown %d:%d %s",
+			uid, gid, fcCfg.KernelImagePath))
+	}
+
+	// Check that each drive has the correct UID and GID.
```
samuelkarp: Same question here, though for writable drives I presume we care about the write bit as well.

example_test.go
xibz: Yes, the jailer will return errors otherwise if the UID and GID do not match.

example_test.go
xibz: Removed comment as it seemed unnecessary since all we need is a readable kernel image

example_test.go
xibz: Fixed!

example_test.go
```diff
@@ -221,52 +217,71 @@ func ExampleNetworkInterface_rateLimiting() {
...
+		panic(fmt.Errorf("Failed to open kernel image: %v", err))
+	}
+	f.Close()
+
+	// Check each drive is readable and writable
+	for _, drive := range fcCfg.Drives {
+		drivePath := firecracker.StringValue(drive.PathOnHost)
+		f, err := os.OpenFile(drivePath, os.O_RDWR, 0666)
+		if err != nil {
+			panic(fmt.Errorf("Failed to drive: %v", err))
```
sipsma: `Failed to open drive for reading and writing`?

xibz: Fixed!

xibz:
> I'm assuming that you've run the example where you've replaced /path/to/rootfs with an actual value to verify it all works in practice? If so LGTM.

Yea, works when replacing the proper strings. I've went ahead and squashed the fix up commits

PR #96, opened 2019-05-14
Improves and fixes jailer

jeromegn:
*Description of changes:*

- Adds configurable Stdin on jailer
- Proper handling of the logs and metrics fifos
- Adds `machine.PID()` to get the process ID and possibly gather metrics on the process

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -115,6 +115,11 @@ var CreateLogFilesHandler = Handler{
 		logFifoPath := m.cfg.LogFifo
 		metricsFifoPath := m.cfg.MetricsFifo
 
+		if len(logFifoPath) == 0 || len(metricsFifoPath) == 0 {
+			// logging is disabled
+			return nil
+		}
+
 		if err := createFifos(logFifoPath, metricsFifoPath); err != nil {
 			m.logger.Errorf("Unable to set up logging: %s", err)
 			return err
--- a/handlers.go
+++ b/handlers.go
@@ -22,6 +22,7 @@ import (
 const (
 	StartVMMHandlerName                = "fcinit.StartVMM"
 	BootstrapLoggingHandlerName        = "fcinit.BootstrapLogging"
+	CreateLogFilesHandlerName          = "fcinit.CreateLogFilesHandler"
 	CreateMachineHandlerName           = "fcinit.CreateMachine"
 	CreateBootSourceHandlerName        = "fcinit.CreateBootSource"
 	AttachDrivesHandlerName            = "fcinit.AttachDrives"
@@ -107,6 +108,24 @@ var StartVMMHandler = Handler{
 	},
 }
 
+// CreateLogFilesHandler is a named handler that will create the fifo log files
+var CreateLogFilesHandler = Handler{
+	Name: CreateLogFilesHandlerName,
+	Fn: func(ctx context.Context, m *Machine) error {
+		logFifoPath := m.cfg.LogFifo
+		metricsFifoPath := m.cfg.MetricsFifo
+
+		if err := createFifos(logFifoPath, metricsFifoPath); err != nil {
+			m.logger.Errorf("Unable to set up logging: %s", err)
+			return err
+		}
+
+		m.logger.Debug("Created metrics and logging fifos.")
+
+		return nil
+	},
+}
+
 // BootstrapLoggingHandler is a named handler that will set up fifo logging of
 // firecracker process.
 var BootstrapLoggingHandler = Handler{
@@ -180,6 +199,7 @@ func NewSetMetadataHandler(metadata interface{}) Handler {
 
 var defaultFcInitHandlerList = HandlerList{}.Append(
 	StartVMMHandler,
+	CreateLogFilesHandler,
 	BootstrapLoggingHandler,
 	CreateMachineHandler,
 	CreateBootSourceHandler,
--- a/jailer.go
+++ b/jailer.go
@@ -309,12 +309,7 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 		stderr = os.Stderr
 	}
 
-	stdin := cfg.JailerCfg.Stdin
-	if stdin == nil {
-		stdin = os.Stdin
-	}
-
-	m.cmd = NewJailerCommandBuilder().
+	builder := NewJailerCommandBuilder().
 		WithID(cfg.JailerCfg.ID).
 		WithUID(*cfg.JailerCfg.UID).
 		WithGID(*cfg.JailerCfg.GID).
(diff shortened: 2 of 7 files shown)


jeromegn:
Some tests are failing, any pointers would help. I assume it's all tests that were calling setupLogging and expecting it to create fifos.

samuelkarp:
> Some tests are failing, any pointers would help. I assume it's all tests that were calling setupLogging and expecting it to create fifos.

I reran the CI job and there are a few things that are showing as failing:

```
--- FAIL: TestJail (0.00s)
    jailer_test.go:162: did not find link files handler
```
This looks like it's caused by your change [here](https://github.com/firecracker-microvm/firecracker-go-sdk/pull/96/files#diff-63049f1d12f4a6384f2b2405c2318c3eR426); the `AppendAfter` function expects the first argument to represent a handler already in the chain.

```
--- FAIL: TestStartVMMOnce (0.02s)
    machine_test.go:408: startVMM failed: Failed to create log fifo: no such file or directory
```
```
--- FAIL: TestLogFiles (0.00s)
    machine_test.go:709: unexpected error: Failed to create log fifo: no such file or directory
```
These two look like failures to find the log fifo, possibly because the `LinkFilesHandler` isn't being run.

```
--- FAIL: TestJailerMicroVMExecution (0.09s)
    machine_test.go:246: Failed to start VMM: link testdata/firecracker.log /tmp/jailer-test760018090/firecracker/b/root/firecracker.log: invalid cross-device link
```
On our CI host, `/tmp` is a bind-mount hard-links fail between /tmp and the directory where tests are being run.

```
*** Test killed with quit: ran too long (10m0s).
```
I don't know offhand what's taking too long to run.

jailer.go
```diff
@@ -367,6 +374,24 @@ func LinkFilesHandler(rootfs, kernelImageFileName string) Handler {
...
+				if err := linkFileToRootFS(
+					m.cfg.JailerCfg,
+					filepath.Join(rootfs, fileName),
+					fifoPath,
+				); err != nil {
+					return err
+				}
+				if err := os.Chown(filepath.Join(rootfs, fileName), *m.cfg.JailerCfg.UID, *m.cfg.JailerCfg.GID); err != nil {
+					return err
+				}
```
xibz: There needs to be an update to the `m.cfg.LogFifo` and `m.cfg.MetricsFifo` path to point to the root directory.

```
*fifoPath = fileName
```

handlers.go
```diff
@@ -107,6 +108,24 @@ var StartVMMHandler = Handler{
...
 }
 
+// CreateLogFilesHandler is a named handler that will create the fifo log files
+var CreateLogFilesHandler = Handler{
+	Name: CreateLogFilesHandlerName,
+	Fn: func(ctx context.Context, m *Machine) error {
+		logFifoPath := m.cfg.LogFifo
+		metricsFifoPath := m.cfg.MetricsFifo
+
+		if err := createFifos(logFifoPath, metricsFifoPath); err != nil {
```
xibz: We should check whether or not any of the paths are empty. If they are logging is disabled.

xibz:
Can you also add an `os.Stat` in the `TestJailerMicroVM` test to ensure that the fifos are actually linked in the Jailer's chroot?

In addition, an update to the `CHANGELOG.md` under the `unreleased` section with a description of this change?

jeromegn:
Might be a little while until I look at this again.

It works for us right now so I'm just using it as-is.

jeromegn:
Anything missing here, since @xibz's pushes?

jailer.go
```diff
@@ -395,18 +416,18 @@ func NewNaiveChrootStrategy(rootfs, kernelImagePath string) NaiveChrootStrategy
...
+// ErrJailerMissingHandler occurs when the CreateLogFilesHandlerName is not
 // present in FcInit.
-var ErrCreateMachineHandlerMissing = fmt.Errorf("%s is missing from FcInit's list", CreateMachineHandlerName)
+var ErrJailerMissingHandler = fmt.Errorf("%s is missing from FcInit's list", CreateLogFilesHandlerName)
 
 // AdaptHandlers will inject the LinkFilesHandler into the handler list.
 func (s NaiveChrootStrategy) AdaptHandlers(handlers *Handlers) error {
-	if !handlers.FcInit.Has(CreateMachineHandlerName) {
-		return ErrCreateMachineHandlerMissing
+	if !handlers.FcInit.Has(CreateLogFilesHandlerName) {
```
samuelkarp: What's the reason behind changing this function?

machine.go
```diff
@@ -152,6 +152,14 @@ func (m *Machine) Logger() *log.Entry {
 	return m.logger.WithField("subsystem", userAgent)
 }
 
+// PID returns the machine's process ID
+func (m *Machine) PID() (int, error) {
```
samuelkarp: I could be missing something, but I don't see this function getting called anywhere.

jailer.go
```diff
@@ -375,6 +374,28 @@ func LinkFilesHandler(rootfs, kernelImageFileName string) Handler {
...
+					*fifoPath,
+				); err != nil {
+					return err
+				}
+
+				if err := os.Chown(filepath.Join(rootfs, fileName), *m.cfg.JailerCfg.UID, *m.cfg.JailerCfg.GID); err != nil {
+					return err
+				}
+
+				*fifoPath = fileName
```
samuelkarp: Can you add a comment above this indicating why it's stored?  Something like:

```go
// update fifoPath as jailer works relative to the chroot dir
```

machine_test.go
```diff
@@ -141,8 +141,8 @@ func TestJailerMicroVMExecution(t *testing.T) {
 	os.MkdirAll(jailerTestPath, 0777)
 
 	socketPath := filepath.Join(jailerTestPath, "firecracker", "TestJailerMicroVMExecution.socket")
-	logFifo := filepath.Join(testDataPath, "firecracker.log")
-	metricsFifo := filepath.Join(testDataPath, "firecracker-metrics")
+	logFifo := filepath.Join(tmpDir, "firecracker.log")
```
samuelkarp: How come this was changed?

jailer.go
xibz: Before it would link after create machine, however since fifos are created after create machine, we need to either update where the handler get's injected or rearrange the default handler list. I opted for updating where it get's injected as the list is ordered pretty logically.

machine.go
xibz: Yea, I wasn't sure about this function, but I assumed @jeromegn was using it for something, so I didn't touch it.

jeromegn: I am using it, but it could be part of a separate PR. I need to get at the PID for the process for the scheduling software I'm using.

jailer.go
samuelkarp: It would be good to add some contextual information into the godoc for this function.  "LinkFilesHandler must be called after CreateLogFilesHandler in order to ensure the fifos are linked properly".

Note that this *is* a breaking change, as we've changed the name of the symbol for the error.  Since we're pre-1.0 it's not impossible to change, but we should make the name flexible enough that we can change the message for the error without having to change the symbol again.  Something like `ErrRequiredHandlerMissing` would make more sense.

machine.go
samuelkarp: @jeromegn Thanks! Can we move this to a separate PR?

machine_test.go
xibz: the `testDataPath` should only house read only files, as this will create issues across tests, more specifically this one due to linking failing since `firecracker.log` or `firecracker-metrics` potentially existing.

jailer.go
xibz: Fixed

jailer.go
xibz: Fixed

samuelkarp: @xibz Can you change this to `ErrRequiredHandlerMissing`?  `ErrRequiredMissingHandler` doesn't really seem grammatical to me.

xibz: Np! fixed.

xibz:
@samuelkarp - Did you mean wrapped body at 72 instead of 50?

samuelkarp:
@xibz oops, yes :sweat_smile:

Issue #100, opened 2019-06-26
Support firecracker 0.17.0 API


nmeyerhans:
The SDK currently is based on the Firecracker 0.15.x API. The current release is 0.17.0, and there have been [some breaking changes](https://github.com/firecracker-microvm/firecracker/compare/v0.15.2...v0.17.0#diff-269561b710bfd7d1718952e4c1893672) to the API specification.

nmeyerhans:
Merged #101

PR #101, opened 2019-06-26
V0.17.0

nmeyerhans:
Addresses #100 by updating the SDK for compatibility with Firecracker 0.17's API.

Certain attributes are now passed as pointers.

`GetMachineConfig` and related objects have been renamed to `GetMachineConfiguration` for consistency with `PutMachineConfiguration`.


nmeyerhans:
Tests will fail until firecracker 0.17.0 installed on the CI hosts.

xibz:
`version.go` file needs to be updated along with the `CHANGELOG.md`'s `Unreleased` section to reflect the correct version.

nmeyerhans:
> `version.go` file needs to be updated along with the `CHANGELOG.md`'s `Unreleased` section to reflect the correct version.

Done.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Fixes bug where context was not being used at all during startVM (#86)
 * Updates the jailer's socket path to point to the unix socket in the jailer's workspace (#86)
 * Fixes bug where default socketpath would always be used when not using jailer (#84).
+* Update for compatibility with Firecracker 0.17.x
```
xibz: The `# [ Unreleased ]` needs to be renamed to `# v0.17.0`

nmeyerhans: Not until we actually release a 0.17.0 version of the SDK, which is not what we're doing here.

nmeyerhans:
A number of fields in the `models.MachineConfiguration` structure generated by OpenAPI have changed from direct references to pointers. As a result, the zero-value of these fields has changed from something vaguely meaningful (or at least technically valid) to a null pointer. This will require changes to client code that is relying on the default values for things like `VcpuCount`, `MemSizeMib`, and `HtEnabled`. I'm updating the validation code and tests to better handle the current behavior...

nmeyerhans:
> Don't need the `nil` check since `Int64Value` will return `0` if the value is `nil`. Unless you did this intentionally for readability?

Yeah, it's kind of for readability. It's the nil value that we really care about. If the value isn't specified, then what we end up sending to firecracker isn't valid, and the API call to `PUT /machine-config` fails with `Missing mandatory fields.` The check for the validity of the numerical value is important, but a secondary consideration. (But maybe that's just a kind of recency bias on my part. I'm not opposed to changing the code if you care more than I do about it.)

xibz:
> I'm not opposed to changing the code if you care more than I do about it.

Im okay with it. Just wanted to make sure the changes were intentional and not ignorant that `Int64Value` returns 0 when nil.

nmeyerhans:
So, I've added code to fail `MachineConfiguration` validation if `VcpuCount`, `MemSizeMib`, or `HtEnabled` are unspecified. Another option would be to apply default values in that case, and proceed with validation. I'm inclined to leave the strict validation in place, since I don't really think it's a good idea for client code to be configuring VMs without making explicit choices about these settings. Does anybody disagree with me? @samuelkarp @sipsma @mxpv @xibz ?

xibz:
@nmeyerhans - Yea, stricter validation and explicit choices is +1 for me.

:shipit:

samuelkarp:
:+1: for strict validation. If we later decide to remove validation, we can do so.

machine.go
```diff
@@ -126,6 +126,17 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("socket %s already exists", cfg.SocketPath)
 	}
 
+	if cfg.MachineCfg.VcpuCount == nil ||
```
samuelkarp: I'd replace these conditionals with `assert.GreaterOrEqual(t, cfg.MachineCfg.VcpuCount, 1, "machine needs a nonzero VcpuCount")` and `assert.NotNil(t, cfg.MachineCfg.HtEnabled, "machine needs a setting for ht_enabled")`.

nmeyerhans: This is in the actually validation code, not in the test code. So the right behavior is to return a validation failure error to the calling code.

samuelkarp: Thanks, you are correct.

PR #102, opened 2019-07-10
Expose machine's PID

jeromegn:
*Description of changes:*

Access to the microvm's process PID is useful for getting process metrics as well as keeping state around. We're using scheduling software and knowing about the PID allows us to "reattach" the task.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -163,6 +163,14 @@ func (m *Machine) Logger() *log.Entry {
 	return m.logger.WithField("subsystem", userAgent)
 }
 
+// PID returns the machine's running process PID or an error if not running
+func (m *Machine) PID() (int, error) {
+	if m.cmd == nil || m.cmd.Process == nil {
+		return 0, fmt.Errorf("machine is not running")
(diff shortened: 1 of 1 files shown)


machine.go
```diff
@@ -163,6 +163,14 @@ func (m *Machine) Logger() *log.Entry {
 	return m.logger.WithField("subsystem", userAgent)
 }
 
+// PID returns the machine's running process PID or an error if not running
+func (m *Machine) PID() (int, error) {
+	if m.cmd == nil || m.cmd.Process == nil {
+		return 0, fmt.Errorf("machine is not running")
```
sipsma: I think we should also return an error if the VM has exited, something like
```
select{
case <-exitCh:
    return 0, fmt.Errorf("machine process has exited")
default:
}
```

Though it doesn't eliminate the possibility entirely, this will at least make it harder for users to accidentally write bugs related to pid reuse.

jeromegn: Good idea, added!

Issue #66, opened 2019-02-22
Add support for patching rate limiters


xibz:
The release of v0.15.0 of firecracker introduces a new API which allows the patching of rate limiters. Currently the SDK does not support this, but should.

PR #103, opened 2019-07-10
Add UpdateGuestNetworkInterface

kzys:
*Issue #, if available:*

#66

*Description of changes:*

While both Drive and NetworkInterface can have RateLimiters on creation,
updating RateLimiters is only supported by NetworkInterface
(via PartialNetworkInterface).

### DIFF
--- a/machine_test.go
+++ b/machine_test.go
@@ -499,7 +499,13 @@ func testUpdateGuestDrive(ctx context.Context, t *testing.T, m *Machine) {
 }
 
 func testUpdateGuestNetworkInterface(ctx context.Context, t *testing.T, m *Machine) {
-	if err := m.UpdateGuestNetworkInterface(ctx, "1", models.PartialNetworkInterface{IfaceID: String("1")}); err != nil {
+	rateLimitSet := RateLimiterSet{
+		InRateLimiter: NewRateLimiter(
+			TokenBucketBuilder{}.WithBucketSize(10).WithRefillDuration(10).Build(),
+			TokenBucketBuilder{}.WithBucketSize(10).WithRefillDuration(10).Build(),
+		),
+	}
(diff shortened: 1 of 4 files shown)


sipsma:
Change looks good!

If possible, I think it's worth additionally adding a test for your change to [`machine_test.go`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/a585107792f345aeb230e10bdc30844edb615e7a/machine_test.go#L326-L335). I don't think we need to necessarily test the rate limiting itself (as that is in all likelihood a huge undertaking out of scope of just adding this feature), but it would be good to even just have a test that calls the `Patch` api you added as a basic sanity check that it doesn't crash or something like that.

machine_test.go
```diff
@@ -489,6 +498,12 @@ func testUpdateGuestDrive(ctx context.Context, t *testing.T, m *Machine) {
 	}
 }
 
+func testUpdateGuestNetworkInterface(ctx context.Context, t *testing.T, m *Machine) {
+	if err := m.UpdateGuestNetworkInterface(ctx, "1", models.PartialNetworkInterface{IfaceID: String("1")}); err != nil {
```
kzys: This looks wrong. Not so sure why PartialNetworkInterface has iface_id, while the RESTful API's path also has iface_id.

xibz: Yea, I think this shows we may want to revisit the design on how we expose this to users. We may want to rename the method to be something like `UpdateGuestNetworkRates` or something along those lines.
```go
func (m *Machine) UpdateGuestNetworkRates(ctx context.Context, ifaceID string, rates RateLimitSet, opts ...UpdateGuestNetworkRatesOpts) {
     // create partial network iface here with the iface id and rate limits
}
````

The reason why there is two is due to how Firecracker handles API requests. I remember opening a github issue about this with `PatchGuestDrive`.

PR #105, opened 2019-07-12
Make logrus' log level configurable via FC_TEST_LOG_LEVEL

kzys:
*Issue #, if available:*

N/A

*Description of changes:*

It makes `go test` less noisy and helps us to find the usage of
the global default logger. Since Machine and Client take logrus.Entry,
they shouldn't use the default one.

### DIFF
--- a/fctesting/utils.go
+++ b/fctesting/utils.go
@@ -16,9 +16,12 @@ package fctesting
 import (
 	"os"
 	"testing"
+
+	log "github.com/sirupsen/logrus"
 )
 
 const rootDisableEnvName = "DISABLE_ROOT_TESTS"
+const logLevelEnvName = "FC_TEST_LOG_LEVEL"
 
 var rootDisabled bool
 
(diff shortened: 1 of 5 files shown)


fctesting/utils.go
```diff
@@ -43,3 +46,24 @@ func RequiresRoot(t testing.TB) {
 			rootDisableEnvName)
 	}
 }
+
+func newLogger() *log.Logger {
+	str := os.Getenv(logLevelEnvName)
+	if str != "" {
+		logLevel, err := log.ParseLevel(str)
+		if err != nil {
+			panic(err)
```
xibz: Would suggest passing in `testing.TB` and calling `Fatalf` instead. This allows users to know that this is only to be used in testing as well.

fctesting/utils.go
```diff
@@ -43,3 +46,24 @@ func RequiresRoot(t testing.TB) {
...
+		}
+		return &log.Logger{
+			Out:   os.Stdout,
+			Level: logLevel,
+		}
+	} else {
```
xibz: Don't need else can just return on parent's scope.

fctesting/utils.go
```diff
@@ -43,3 +46,24 @@ func RequiresRoot(t testing.TB) {
...
+		return log.New()
+	}
+}
+
+// NewLogEntry creates log.Entry. The level is specified by "FC_TEST_LOG_LEVEL" environment variable
+func NewLogEntry() *log.Entry {
```
xibz: This should probably take a `testing.TB` and pass that to `newLogger`. This will get rid of the panic and instead call `t.Fatalf`

fctesting/utils.go
kzys: Right and golint also pointed out that.

fctesting/utils.go
kzys: Thanks. Will do.

fctesting/utils.go
```diff
@@ -43,3 +46,24 @@ func RequiresRoot(t testing.TB) {
...
+
+func newLogger(t testing.TB) *log.Logger {
+	str := os.Getenv(logLevelEnvName)
+	if str == "" {
+		return log.New()
+	}
+
+	logLevel, err := log.ParseLevel(str)
+	if err != nil {
+		t.Fatalf("Failed to parse '%s' as Log Level: %v", str, err)
```
xibz: nit: you can use `%q` to single quote wrap a string.

kzys: Didn't know that. Will fix!

PR #110, opened 2019-07-31
Removes JailerEnable and changes JailerCfg to be a pointer

xibz:
This fix allows for a more intuitive design in letting users enable the
jailer.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -90,12 +90,8 @@ type Config struct {
 	// validation of configuration performed by the SDK.
 	DisableValidation bool
 
-	// EnableJailer will enable the jailer. By enabling the jailer, root level
-	// permissions are required.
-	EnableJailer bool
-
 	// JailerCfg is configuration specific for the jailer process.
-	JailerCfg JailerConfig
+	JailerCfg *JailerConfig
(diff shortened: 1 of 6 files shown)


machine.go
```diff
@@ -98,6 +98,33 @@ type Config struct {
 	JailerCfg JailerConfig
 }
 
+// WithJailerConfig will set the given JailerConfig to the JailerCfg field. In
+// addition this will also set EnableJailer to true
```
samuelkarp: Do we strictly need to have the `EnableJailer` field or can we enable the jailer when `JailerCfg != nil`?

xibz: Yea, that is an option, using nil as an indicator. I wanted to prevent doing that since whenever you want to grab values from the jailer config, you'd need to nil check. But I am okay with using nil as a check if we think that is a better experience.

sipsma: I agree it seems simpler to just have our code do nil checks internally and enable the jailer when JailerCfg is not nil. I think I'd be confused as a user if I saw there was `EnableJailer`, `JailerCfg` and `WithJailerConfig`.

PR #112, opened 2019-08-08
Expose config to allow for advanced handlers

xibz:
Exposing the config allows for modification of the machine configuration
during the 'Start' method. An example usage of this is using a custom
jailer

Also fixes a linting issue for `ErrMissingJailerConfig`

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -46,7 +46,7 @@ var ConfigValidationHandler = Handler{
 	Name: ValidateCfgHandlerName,
 	Fn: func(ctx context.Context, m *Machine) error {
 		// ensure that the configuration is valid for the FcInit handlers.
-		return m.cfg.Validate()
+		return m.Cfg.Validate()
 	},
 }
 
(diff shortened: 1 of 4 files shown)


kzys:
How about renaming that to `Config` since we are going to expose the member to end-customers? I know that Go is known for shorter variable names, but even `golang/go` uses `Config` over `Cfg`.

* https://github.com/golang/go/blob/1dc0110bf725640a9b912e3d31e6654ed1c4da9d/src/cmd/compile/internal/ssa/func.go#L27
* https://github.com/golang/go/blob/1dc0110bf725640a9b912e3d31e6654ed1c4da9d/src/go/printer/printer.go#L54
* https://github.com/golang/go/blob/1dc0110bf725640a9b912e3d31e6654ed1c4da9d/src/image/gif/reader.go#L593

```
% git grep '\.Config ='
src/cmd/compile/internal/gc/ssa.go:     s.f.Config = ssaConfig
src/cmd/compile/internal/ssa/func_test.go:      f.Config = c.config
src/go/printer/printer.go:      p.Config = *cfg
src/image/gif/writer.go:        if e.g.Config == (image.Config{}) {
src/image/gif/writer_test.go:           g0.Config = image.Config{
% git grep '\.Cfg = '
%
```

xibz:
@kzys - While I would agree, I think we should be consistent on in the SDK on how we name things. We already have fields called `MachineCfg` and `JailerCfg`.

kzys:
Ah right. Then we should use `Cfg` here.

PR #114, opened 2019-08-22
Added GetMetadata function to Machine

### LINKED ISSUE #109: Support retrieval of MMDS from firecracker
There currently is no method on the Machine struct for retrieving metadata from Firecracker. 

There's a method to set the metadata and it looks like the client is setup to make the GET call for MMDS, I think things just need to be hooked up between the client and the Machine.

cpheps:
*Issue #, if available:*
Resolves #109

*Description of changes:*
Added a function `GetMetadata` function to the Machine to wrap the `GetMmds` of client. I had the function return a `map[string]interface{}`, I figured this was the path of least resistance to return the meta data and allow the caller to parse it.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -15,6 +15,7 @@ package firecracker
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


machine.go
```diff
@@ -660,6 +660,24 @@ func (m *Machine) SetMetadata(ctx context.Context, metadata interface{}) error {
...
 
+// GetMetadata gets the machine's metadata from MDDS
+func (m *Machine) GetMetadata(ctx context.Context) (map[string]interface{}, error) {
+	resp, err := m.client.GetMmds(ctx)
+	if err != nil {
+		m.logger.Errorf("Getting metadata: %s", err)
+		return nil, err
+	}
+
+	payload, ok := resp.Payload.(map[string]interface{})
```
xibz: I don't know if I like defaulting to `map[string]interface{}` as `[]interface{}` seems just as feasible. I think we need to think about this a little more, maybe use reflection to unmarshal into a structure may be nice.

```go
func(m *Machine) GetMetadata(ctx context.Context, v interface{}) error {
    // TODO: use reflection here
}
```

How do you feel about just updating the client for now and removing the machine `GetMetadata` and adding an issue asking for a higher level abstraction?

cpheps: I thought about reflection too, I wasn't sure if you guys wanted all that code inside of the sdk. I don't think it'll be too bad since the MMDs only allows strings, objects, and arrays of strings. 

I assume for the reflection we would just unmarshal off of the json struct tags?

xibz: @cpheps - Yea, using json tags is the way to go, and I think this creates a much better interface for users to be able to have the SDK do the unmarshaling for them.

cpheps: Agreed, I'll go down that path. I also noticed the Patch MMDs was not hooked up to the machine, that seems fairly straight forward and I need that functionality so I'm going to throw that onto this PR.

cpheps:
@xibz Added reflection unmarshaling to GetMmds and hooked up the PATCH MMDs to the machine.

cpheps:
@samuelkarp Do you know if you guys are waiting for a resolution on firecracker-microvm/firecracker#1229 before merging? We're currently using my fork in production but would like to get back to the actual library just wondering on the timeline.

samuelkarp:
Merged as https://github.com/firecracker-microvm/firecracker-go-sdk/commit/b000873be8091832e6abd02d7e745b2617963145, commit sha is different as the commit message was rewritten https://github.com/firecracker-microvm/firecracker-go-sdk/commit/c3a3e5caf56c6b97c41c827013ebd55a30310e6b.

PR #117, opened 2019-08-28
Add tc-redirect-tap plugin with support for ADD command.

sipsma:
As of this commit, this is not yet integrated with the go-sdk itself (that will be done in follow-ups). In the long-term, this plugin will likely live in its own separate repository but is
being developed in-tree here to ease development for the time being.


cni/cmd/tc-redirect-tap/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"testing"
+
```
samuelkarp: nit: blank line can be removed

cni/cmd/tc-redirect-tap/main_test.go
```diff
@@ -0,0 +1,296 @@
...
+		CNIVersion: baseResult.CNIVersion,
+		Interfaces: append([]*current.Interface{}, baseResult.Interfaces...),
+		IPs:        append([]*current.IPConfig{}, baseResult.IPs...),
+	}
+
+	err = p.add(newResult)
+	require.NoError(t, err,
+		"failed to add tap device")
+
+	require.Len(t, newResult.Interfaces,
```
kzys: For checking the value which you can access without any additional function calls, I'm inclined to use `assert` than `require`, since it would be easier to read.

I think assert/require (or Fail/FailNow on testing package) are coming from googletest, which explicitly recommends non-fatal versions over fatal-versions;

https://github.com/google/googletest/blob/master/googletest/docs/primer.md

> The assertions come in pairs that test the same thing but have different effects on the current function. ASSERT_* versions generate fatal failures when they fail, and abort the current function. EXPECT_* versions generate nonfatal failures, which don't abort the current function. Usually EXPECT_* are preferred, as they allow more than one failure to be reported in a test. However, you should use ASSERT_* if it doesn't make sense to continue when the assertion in question fails.

cni/cmd/tc-redirect-tap/main_test.go
sipsma: fixed

cni/cmd/tc-redirect-tap/main_test.go
sipsma: Yeah, agree that's useful, updated all the test files

Issue #115, opened 2019-08-27
Add support for describeInstance


haikuoliu:
Hi all,

Firecracker-containerd `GetVMInfo` API doesn't really call the VM API, we want to periodically call some VM API to check if instance is still responsive, we'd like to use [describeInstance](https://github.com/firecracker-microvm/firecracker/blob/29f81f1e72361329aa09c9db0d87e68d3d091fcf/api_server/swagger/firecracker.yaml#L29) API for this purpose, but  it doesn't seem it's implemented, do you have plan to implement it? Apologize ahead if you already done this.

Thanks,
Haikuo

kzys:
Currently our Machine interface doesn't expose the underlying client, but we would do that. Please see the discussion on #116.

kzys:
Another health check use-case. https://firecracker-microvm.slack.com/archives/CDL3FUR8B/p1572309553045800

kzys:
https://github.com/firecracker-microvm/firecracker-go-sdk/pull/116#issuecomment-642244716

> I see that this work was scrapped in favor of providing a getter method for the inner client, I would be willing to implement if the change isn't already in progress.
> 
> I'm looking for a way to run a health check against a running VM and know that it is in the running state. Are there any existing solutions?

Hello  @olibearo. I think that the definition of "healthy" depends on applications. For example, firecracracker-containerd is launching a small agent process inside a VM. So having a connection to the agent means "healthy" for firecracker-containerd. But other applications would have different definitions.

mazzy89:
I'm interested to have a way to check the healthiness of the VM. In my case, I just run firecracker VMs inside a container. Similar to how Ignite does by spawning inside a Docker container a Firecracker VM so in my case I would need to implement a liveness/readiness endpoint that checks the status of the running VM.

PR #119, opened 2019-08-29
Add health checker

mxpv:
Signed-off-by: Maksym Pavlenko <makpav@amazon.com>

*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-go-sdk/issues/115

*Description of changes:*
Adds `IsAlive`

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -27,8 +27,9 @@ import (
 	"syscall"
 	"time"
 
-	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 	log "github.com/sirupsen/logrus"
+
+	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 )
 
 const (
@@ -710,8 +711,18 @@ func (m *Machine) UpdateGuestDrive(ctx context.Context, driveID, pathOnHost stri
 // IsAlive makes sure that a Firecracker instance is running and responds via API by
 // calling `DescribeInstance` endpoint behind.
 func (m *Machine) IsAlive(ctx context.Context) error {
-	_, err := m.client.DescribeInstance(ctx)
-	return err
+	instance, err := m.client.DescribeInstance(ctx)
+	if err != nil {
+		return err
+	}
+
+	state := StringValue(instance.Payload.State)
+	switch state {
+	case models.InstanceInfoStateStarting, models.InstanceInfoStateRunning:
+		return nil
+	default:
+		return fmt.Errorf("invalid instance state: %q", state)
+	}
 }
 
 // refreshMachineConfiguration synchronizes our cached representation of the machine configuration
--- a/machine.go
+++ b/machine.go
@@ -708,9 +708,9 @@ func (m *Machine) UpdateGuestDrive(ctx context.Context, driveID, pathOnHost stri
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


machine.go
```diff
@@ -707,6 +707,13 @@ func (m *Machine) UpdateGuestDrive(ctx context.Context, driveID, pathOnHost stri
 	return nil
 }
 
+// IsAlive makes sure that a Firecracker instance is running and responds via API by
+// calling `DescribeInstance` endpoint behind.
+func (m *Machine) IsAlive(ctx context.Context) error {
```
sipsma: It looks like one of the valid states that can be returned by `DescribeInstance` is `Halted`: https://github.com/firecracker-microvm/firecracker/blob/35b73437641122865b1603bd74c144035e1790c9/api_server/swagger/firecracker.yaml#L453

There's also `Uninitialized`. That makes me feel like `IsAlive` is not really what one would expect. It seems more like it's checking whether the VM manager is alive rather than what we'd consider the `Machine`.

If we want to keep the behavior as is, I can't think of any great names, but maybe something like `Exists` would be better than `IsAlive`?

On the other hand, would it make sense to change the behavior and return an error here if the State is `Halted` or `Uninitialized`? What's the actual motivation behind this change?

sipsma: (I accidentally clicked the approve button, I think this should be addressed before merging)

samuelkarp: (@sipsma - I've "dismissed" your review per your comment here.)

samuelkarp: Some suggestions for names: `IsVMMAlive`, `IsVMMResponsive`, `Reachable`

machine.go
```diff
@@ -707,6 +708,23 @@ func (m *Machine) UpdateGuestDrive(ctx context.Context, driveID, pathOnHost stri
...
+	instance, err := m.client.DescribeInstance(ctx)
+	if err != nil {
+		return err
+	}
+
+	state := StringValue(instance.Payload.State)
```
samuelkarp: I don't think it makes sense to verify the state if the purpose of this method is to see whether Firecracker responds to an API.  I think getting any non-error response would meet the success criteria.

sipsma: If the motivation behind the change is to test the responsiveness of the API, then agreed and also think `IsVMMResponsive` is a good name.

If the motivation is instead to test that the Machine is in the running state, then can we just add a `State` function that returns the state value, whatever it may be?

mxpv: @haikuoliu which behavior would suite better here?

mxpv: Right now I've renamed it to `IsRunning` which tells whether a VM is in `Starting` or `Running` state.

kzys: It is a bit confusing to see that we have `Running` as a state, and `IsRunning` actually means `IsRunningOrStarting`. Naming is hard...

I'm inclined to expose Firecracker's state as is instead of introducing our own notion of the state (running, responsive, alive, ...).

kzys: The status type could provide rich methods later, like https://godoc.org/inet.af/http#Status.IsClientError (inet.af is bradfitz's net/http rework proposal, which is referenced from https://github.com/golang/go/issues/29652)

mxpv:
Closing until we figure out requirements.

PR #122, opened 2019-09-04
Download publicly-available artifacts when possible

kzys:
*Issue #, if available:*

NA

*Description of changes:*

While GitHub's firecracker 0.17.0 binary cannot be used due to vsock,
0.18.0 would be able to use with the Go SDK off-the-shelf.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -35,18 +35,16 @@ steps:
 
   - label: ':hammer: tests'
     commands:
-      - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
+      - ./download-testdata.sh
       - 'ln -s /usr/local/bin/firecracker-v0.17.0 testdata/firecracker'
-      - 'ln -s /usr/local/bin/jailer-v0.17.0 testdata/jailer'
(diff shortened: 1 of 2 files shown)


kzys:
I agree that keeping the artifacts on the instance is more efficient and have a fewer error cases, but the instance preparation script is really not visible from the collaborators of the project who are not working at AWS.

Other options would be;
- Caching the artifacts on /tmp or somewhere. More error cases though.
- Open-source the instance preparation script.
- Document what we do on the script.

Thoughts?

samuelkarp:
> Document what we do on the script.

I think this is the easiest answer for now.  The script is pretty straightforward though:

* Launch a *.metal instance from the official Debian AMI
* RAID0 the local NVMe storage, configure LVM, and divide the big PV into a linear LV at `/local` and a thin pool for use with our snapshotter
* Download a bunch of things from our S3 bucket (firecracker, jailer, containerd, runc, disk images)
* Install the buildkite-agent and Docker
* Set up a healthcheck to let us know the host is alive
* Recycle the host at a regular cadence

kzys:
I see. I've opened #123 instead. The NVMe, containerd and runc are currently not in use in this repository. So I've excluded from from the pull request.

PR #124, opened 2019-09-04
Add DEL and CHECK support to tc-redirect plugin

sipsma:

There are unit tests and have also verified with e2e tests that will be included in upcoming GoSDK+Firecracker-containerd PRs (which are too large to include with this change).

### DIFF
--- a/cni/cmd/tc-redirect-tap/main.go
+++ b/cni/cmd/tc-redirect-tap/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containernetworking/cni/pkg/version"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containernetworking/plugins/pkg/utils/buildversion"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 
 	"github.com/firecracker-microvm/firecracker-go-sdk/cni/internal"
@@ -42,23 +43,32 @@ func add(args *skel.CmdArgs) error {
 		return err
 	}
 
-	currentResult, err := getCurrentResult(args)
+	if p.netNS == nil {
+		return errors.Errorf("netns path %q does not exist", args.Netns)
+	}
+
+	if p.currentResult == nil {
+		return &NoPreviousResultError{}
+	}
+
+	err = p.add()
 	if err != nil {
 		return err
 	}
 
-	err = p.add(currentResult)
-	if err != nil {
-		return err
-	}
-
-	return types.PrintResult(currentResult, currentResult.CNIVersion)
+	return types.PrintResult(p.currentResult, p.currentResult.CNIVersion)
 }
 
 func del(args *skel.CmdArgs) error {
 	p, err := newPlugin(args)
 	if err != nil {
 		return err
+	}
+
+	if p.netNS == nil {
+		// the network namespace is already gone and everything we create
+		// is inside the netns, so nothing to do here.
+		return nil
 	}
 
 	return p.del()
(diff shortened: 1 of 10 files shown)


cni/cmd/tc-redirect-tap/main.go
```diff
@@ -229,9 +274,106 @@ func (p plugin) add(currentResult *current.Result) error {
...
-	panic("del is currently unimplemented")
+	return p.netNS.Do(func(_ ns.NetNS) error {
+		var multiErr *multierror.Error
+
+		redirectLink, err := p.GetLink(p.redirectInterfaceName)
+		switch err.(type) {
+		case *internal.LinkNotFoundError:
+			// if the link doesn't exist, there's nothing to do
+		default:
+			if err != nil {
```
kzys: `case nil:`? Maybe personal preference though.

sipsma: Oh nice I didn't know you could have a `case nil` statement, updated with that

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -229,9 +274,105 @@ func (p plugin) add(currentResult *current.Result) error {
...
 
 func (p plugin) del() error {
-	panic("del is currently unimplemented")
+	return p.netNS.Do(func(_ ns.NetNS) error {
+		var multiErr *multierror.Error
+
+		// try to remove the qdisc we added from the redirect interface
+		redirectLink, err := p.GetLink(p.redirectInterfaceName)
+		switch err.(type) {
+
```
samuelkarp: nit: I think we can get rid of the blank lines before the first `case`.

cni/internal/cniutil.go
```diff
@@ -61,3 +64,41 @@ func IfacesWithName(name string, ifaces ...*current.Interface) []*current.Interf
 
 	return foundIfaces
 }
+
+// VMTapPair takes a CNI result and returns the tap iface and vm iface corresponding
+// to the provided vmID. See the vmconf package docs for details on the expected
+// vm and tap iface configurations.
+func VMTapPair(
```
samuelkarp: This function declaration looks super weird.  Would this look better?

```go
func VMTapPair(
	result *current.Result,
	vmID string,
)(
	vmIface *current.Interface,
	tapIface *current.Interface,
	err error
)
```

Also: can we avoid the named returns?

cni/internal/cniutil.go
```diff
@@ -61,3 +64,41 @@ func IfacesWithName(name string, ifaces ...*current.Interface) []*current.Interf
...
+	result *current.Result, vmID string,
+) (vmIface *current.Interface, tapIface *current.Interface, err error) {
+	vmIfaces, otherIfaces := FilterBySandbox(vmID, result.Interfaces...)
+	if len(vmIfaces) > 1 {
+		return nil, nil, errors.Errorf(
+			"expected to find at most 1 interface in sandbox %q, but instead found %d",
+			vmID, len(vmIfaces))
+	}
+	if len(vmIfaces) == 0 {
```
samuelkarp: nit: this can be an else if

go.mod
```diff
@@ -8,9 +8,12 @@ require (
...
 	github.com/go-openapi/swag v0.17.1
 	github.com/go-openapi/validate v0.17.1
+	github.com/google/go-cmp v0.3.1 // indirect
+	github.com/hashicorp/go-multierror v1.0.0
 	github.com/pkg/errors v0.8.1
 	github.com/sirupsen/logrus v1.1.1
 	github.com/stretchr/testify v1.3.0
 	github.com/vishvananda/netlink v1.0.0
 	golang.org/x/sys v0.0.0-20190616124812-15dcb6c0061f
+	gotest.tools v2.2.0+incompatible
```
samuelkarp: Is this a transitive dependency?

go.mod
```diff
@@ -8,9 +8,12 @@ require (
 	github.com/go-openapi/strfmt v0.17.1
 	github.com/go-openapi/swag v0.17.1
 	github.com/go-openapi/validate v0.17.1
+	github.com/google/go-cmp v0.3.1 // indirect
```
samuelkarp: Is this a transitive dependency?

cni/cmd/tc-redirect-tap/main.go
sipsma: I put it there because I think it's easier to read/navigate when each case is a separate block. If there's disagreement I can change it.

cni/internal/cniutil.go
sipsma: Sure I don't have a preference either way for both.

cni/internal/cniutil.go
sipsma: Sure, I don't have a preference either way.

go.mod
sipsma: These and the other one were removed when I ran `go mod tidy`, so updating without them

cni/internal/cniutil.go
sipsma: Updated the format of the declaration, but as I was editing I remembered I used the named returns because I think it makes it clear which arg is which since the first two are both of the same type.

Given that, I think this is an okay use of named returns because there are no naked return statements (all of the args are always fully specified in each `return`) and there's no weirdness happening with defers modifying the return values or something like that. Let me know if you disagree.

cni/cmd/tc-redirect-tap/main.go
samuelkarp: I'm not a fan, but it's not a big deal.  Feel free to leave it.

cni/internal/cniutil.go
samuelkarp: I tend to dislike named returns and use them only when necessary (defers modifying the returns is one use-case).  You did avoid naked returns; naked returns make it much harder to reason about what is actually returned.  But the use of named returns make it possible for future introduction of a naked return by someone else.

I don't feel strongly enough here to want to force you to change it, but I avoid named returns as much as possible.

PR #125, opened 2019-09-09
Add support for CNI-configured network interfaces.

### LINKED ISSUE #126: Support CNI-configured network interfaces
(This code resolving this is already out, retroactively creating to have an issue to reference)

As part of the design for adding CNI support to Firecracker-containerd, it was decided that most of the functionality should be added to the Go SDK so it can benefit both Firecracker-containerd and potentially other consumers. The relevant design doc can be found here: https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/networking.md

sipsma:

Closes #126 

See the README updates for an initial overview of the changes.

For those who have reviewed previous iterations of this code, it's largely the same with a few notable exceptions:
1. Docs in `README.md`
1. Tests (unit and an e2e test) plus a new install makefile target in `tc-redirect-tap` that's used to initialize the tests (though is also generally useful)
1. `DeleteNetworkList` is now called, both as a cleanup handler and before network creation in order to better handle the case where there was a crash while a previous VM using that exact network configuration was running (CNI does not allow two networks with the exact same configuration, including `ContainerID`, to run at the same time)
1. Tweaks to the process for obtaining a unique ID from the VM's socket path. The canonicalized path often easily exceeds `SUN_PATH` max len, so instead the path is left untouched and the canonicalization only affects the unique id. It also more carefully evaluates symlinks in the path to ensure that paths not yet existing are handled gracefully


.buildkite/pipeline.yml
```diff
@@ -34,13 +34,16 @@ steps:
   - wait
 
   - label: 'build'
-    command: 'make'
+    commands:
+      - 'make'
+      - 'make -C cni'
```
kzys: How about invoking `make -C cni` from the top-level make like other targets?

machine.go
```diff
@@ -264,10 +284,44 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
...
 }
 
+func uniqueIDFromSocketPath(socketPath string) (string, error) {
+	absPath, err := filepath.Abs(socketPath)
+	if err != nil {
+		return "", errors.Wrapf(err,
+			"failed to get absolute path of socket path %q", socketPath)
+	}
+
+	// we don't assume all parent directories for the socket yet exist, so we evaluate
```
kzys: Not so sure what you want to accomplish by doing so.

Let's say the sockerPath is `/foo/bar/baz` and `/foo` is liked to `/realfoo` which is missing, can't you check the existence of `/foo/bar/baz` directly to confirm that the path is valid?

network_test.go
```diff
@@ -0,0 +1,392 @@
...
+    },
+    {
+      "type": "tc-redirect-tap"
+    }
+  ]
+}`, networkName, ipamDataDir)
+
+	cniConfPath := filepath.Join(cniConfDir, fmt.Sprintf("%s.conflist", networkName))
+	require.NoError(t,
+		ioutil.WriteFile(cniConfPath, []byte(cniConf), 0644),
```
kzys: Does it have to be writable from the group and others, but not readable?

network_test.go
```diff
@@ -0,0 +1,392 @@
...
+		"failed to write cni conf file")
+
+	numVMs := 10
+	vmIPs := make(chan string, numVMs)
+
+	var vmWg sync.WaitGroup
+	for i := 0; i < numVMs; i++ {
+		vmWg.Add(1)
+		go func(vmID string) {
```
kzys: Extracting as a function?

.buildkite/pipeline.yml
sipsma: I think I prefer them being separate since `tc-redirect-tap` is what's built by `make -C cni` and it's more of an optional part of the Go SDK (and eventually will probably not be in this repo at all).

machine.go
sipsma: The problem occurs when:
1. `/foo/bar` doesn't exist yet
1. `/foo` does exist and is a symlink to `/realfoo`

In that case trying to resolve `/foo/bar` with just `EvalSymlinks` will fail with `ENOENT` whereas what we actually want is to obtain `/realfoo/bar` even though `bar` doesn't yet exist. Let me know if that makes sense

network_test.go
sipsma: `0644` should result in `-rw-r--r--`, so it's just readable by others here, which is harmless for this configuration file in this test

network_test.go
sipsma: Overall this function is very specific to this test. However, I will try to extract some individual parts of it that could be generalized and re-used, which I agree would be good.

network_test.go
kzys: Oh, sorry. Right.

network_test.go
kzys: Yeah, I agree that the function on the goroutine won't be used by others, but just thought that the whole function is long.

sipsma: For sure, I extracted a few decent size chunks out now, let me know what you think.

kernelArgs.go
```diff
@@ -0,0 +1,57 @@
...
+// deserialize the provided string to a kernelArgs map
+func parseKernelArgs(rawString string) kernelArgs {
+	argMap := make(map[string]*string)
+	for _, kv := range strings.Fields(rawString) {
+		kvSplit := strings.SplitN(kv, "=", 2)
+
+		key := kvSplit[0]
+
+		var value *string
+		if len(kvSplit) == 2 {
```
xibz: it is possible for this value `key=v1=v2`, however you'd only would get `key` with this logic. Can you also add a test to capture this?

You probably want to `strings.Join` on `kvSplit` being greater than 1

machine.go
```diff
@@ -264,10 +284,44 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
...
+	// symlinks in the path one component at a time to handle any ENOENT errors
+	parentDir := "/"
+	for _, pathPart := range strings.Split(filepath.Dir(absPath), string(os.PathSeparator)) {
+		parentDir = filepath.Join(parentDir, pathPart)
+
+		resolvedDir, err := filepath.EvalSymlinks(parentDir)
+		if err != nil && !os.IsNotExist(err) {
+			return "", errors.Wrapf(err,
+				"failed while attempting to resolve any symlinks in socket path %q", socketPath)
+		} else {
```
xibz: nit: can lose the else

README.md
```diff
@@ -35,8 +35,107 @@ Network configuration
...
 
 Firecracker, by design, only supports Linux tap devices. The SDK
-provides facilities to attach a tap device to the Firecracker VM, but
-the client is responsible for further configuration.
+provides facilities to:
+* Attach a pre-created tap device, optionally with static IP configuration, to
+  the VM. This is referred to as a "static network interface".
+* Create a tap device via [CNI](https://github.com/containernetworking/cni) plugins, 
+  which will then be attached to the VM automatically by the SDK. This is referred 
+  to as a "CNI network interface"
```
samuelkarp: Considering CNI stands for "Container Network Interface", maybe we can refer to this as a "CNI-configured interface" instead?

network.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	return nil
+}
+
+// return the network interface that has static IP configuration, or nil if there is no such interface
+func (networkInterfaces NetworkInterfaces) staticIPInterface() *NetworkInterface {
+	// Validation that there is at most one interface with StaticIPConfiguration
+	// is done as part of the NetworkConfigValidationHandler, can safely just use
+	// the first result here and assume it's the only one.
```
xibz: And if there are more than one? Should we return an error or do some more stricter validation up top?

README.md
```diff
@@ -35,8 +35,107 @@ Network configuration
...
+}
+```
+
+and the 
+[`ptp`](https://github.com/containernetworking/plugins/tree/master/plugins/main/ptp), 
+[`host-local`](https://github.com/containernetworking/plugins/tree/master/plugins/ipam/host-local) 
+and [`tc-redirect-tap`](cni/Makefile)
+CNI plugin binaries installed under `/opt/cni/bin`, you can specify, in the Go SDK API, 
+a `Machine` with the following `NetworkInterface`:
+```
```
samuelkarp: You can do this to have a block with syntax highlighting:

    ```go

kernelArgs.go
sipsma: That's not possible because we use `strings.SplitN` with `n=2` so there can only be up to 2. I'll add a comment about this though.

machine.go
sipsma: I've gotten conflicting opinions on similar situations (whether there should be an `else` after an if statement with a `return`). I'm personally slightly biased towards using `else` just because it's a little more resilient to changes in the future (if the `return` is moved or something like that), but don't have a strong opinion. Let me know what you think.

README.md
```diff
@@ -35,8 +35,107 @@ Network configuration
...
+    CNIConfiguration: &firecracker.CNIConfiguration{
+      NetworkName: "fcnet",
+      IfName: "veth0",
+    },
+  }}
+}
+```
+
+Note that `NetworkName` in the `CNIConfiguration` of the API matches the `name` field 
+specified inside the `/etc/cni/conf.d/fcnet.conflist` file.
```
samuelkarp: Are you required to have the `/etc/cni/conf.d/fcnet.conflist` file?  Is it possible to specify all of these options via an API call instead of requiring an on-disk file?

README.md
```diff
@@ -35,8 +35,107 @@ Network configuration
...
+CNI and place the final VM inside the resultant network namespace. The end result being:
+* Outside the network namespace, a single veth endpoint created by the `ptp` plugin will
+  exist with a static IP from the `host-local` plugin (i.e. `192.168.127.1`)
+* Inside the VM's network namespace:
+    * The other side of the veth device will exist with name `veth0`, as specified by the
+      `IfName` parameter above, and a different IP (i.e. `192.168.127.2`)
+    * The tap device created by `tc-redirect-tap`, which will not have an IP but will have
+      all of its traffic mirrored with the `veth0` device
+* Inside the actual Firecracker VM guest:
+    * A network interface with the same IP as that of `veth0` (i.e. `192.168.127.2`)
```
samuelkarp: What's the interface name inside the VM guest?

network.go
sipsma: The code as is returns an error in the case that there is an interface with IP configuration when multiple interfaces were provided (see `NetworkInterfaces.validate`, currently line 360 of this file). The comment here was just meant to explain that validation already occurs there so we can thus operate on simpler assumptions rather than re-validate here.

README.md
```diff
@@ -35,8 +35,107 @@ Network configuration
...
+  * Users can specify multiple static interfaces as long as none of them 
+    include IP configuration.
+* DNS nameserver settings will only be effective if the VM's rootfs makes
+  `/etc/resolv.conf` be a symlink to `/proc/net/pnp`.
+* Only up to 2 DNS nameservers can be configured within the VM internally.
+  * If a static network interface specifies more than 2, an error will be 
+    returned.
+  * If a CNI network interface receives more than 2 nameservers from the CNI 
+    invocation result, the nameservers after the second will be ignored without 
+    error (in order to be compatible with pre-existing CNI plugins/configuration).
```
samuelkarp: Can we emit a warning into the log?

kernelArgs.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
samuelkarp: nit: I think it's a bit odd to have `kernelArgs.go` as the file name.  I'd vote for either `kernelargs.go`, `kernel.go`, or `kernel_args.go`.

kernelArgs_test.go
```diff
@@ -0,0 +1,48 @@
...
+		"foo":  &fooVal,
+		"blah": nil,
+		"huh":  &emptyVal,
+		"bleh": nil,
+		"duh":  &emptyVal,
+		"boo":  &booVal,
+	})
+
+	actualParsedArgs := parseKernelArgs(argsString)
+	require.Equal(t, expectedParsedArgs, actualParsedArgs)
```
samuelkarp: Can you add a comment at the end like this?

```go
require.Equal(t, expectedParsedArgs, actualParsedArgs, "arguments parsed correctly")
```

Adding different comments for different `require`/`assert` calls will make it easier to disambiguate which one failed.

Also, I think these should both be `assert` instead of `require`.  Then you can see both fail and see how they failed.

README.md
sipsma: Right now, yes you have to make a file. I thought about what you're suggesting too, but specifying these settings in a file is the most typical way of using CNI and what other comparable frameworks ask for, so I think asking for a filepath is the most basic support we should offer. We can very easily add support for specifying all of the parameters directly via the API on top of this, but that should be a next iteration IMO. I'll open an issue for doing so if there's no disagreement.

README.md
sipsma: In practice it's often `eth0`, but it's gonna be up to the VM guest's OS AFAIK. It's not something we configure or need to know at the moment. We just supply the `ip=...` boot params and then the kernel applies those to the first non-loopback interface it finds (which is part of why we validate that if IP configuration is supplied, only up to one network interface can be provided).

In the future, if we move away from the `ip=...` implementation, we may need to start caring about this and/or having the user provide values for guest interface names, but until that time I don't think it's worth mentioning.

README.md
sipsma: Yeah good idea, will do

kernelArgs.go
sipsma: Sure, I have no preference either way

kernelArgs_test.go
sipsma: Yeah I will add the comments to the assertions.

I made this require because if the assertion fails the second assertion is going to be using input that just failed a test, so running the second assertion doesn't seem to provide much value in that case.

kernelArgs.go
xibz: Ah, totally forgot about that line.

machine.go
xibz: `lint` will complain about it

network.go
xibz: Yea, but this doesn't check if there aren't multiple `StaticConfiguration`s, which is where the issues lies for me. I don't know I like assuming that only one was provided.

machine.go
samuelkarp: I'd lean toward not having the else.

network.go
sipsma: We aren't assuming there is only one provided, we are just doing the validation elsewhere. It's just up to the caller (in this case, `Machine`) to validate `NetworkInterfaces` before trying to use it. I considered making a `NewNetworkInterfaces` func that does the creation of `NetworkInterfaces` and validation all in one go, but that seemed to go against the pattern already in the SDK to use validation handlers, which is why I went with what I have.

machine.go
sipsma: There doesn't appear to be any lint rules defined for the SDK at the moment so I'll leave it as is for now since I kind of disagree with that linting rule. If anyone else feels differently let me know.

network.go
```diff
@@ -0,0 +1,505 @@
...
+	// Note that this does NOT necessarily correspond to the name of the
+	// tap device the Firecracker VM will use as the tap device may be
+	// created by a chained plugin that adapts the tap to a pre-existing
+	// network device (which will by the one with "IfName").
+	IfName string
+
+	// Args (optional) corresponds to the CNI_ARGS parameter as specified in
+	// the CNI spec. It allows custom args to be passed to CNI plugins during
+	// invocation.
+	Args [][2]string
```
samuelkarp: Did we decide that changing this to `map[string]string` wasn't feasible?

machine.go
sipsma: > I'd lean toward not having the else.

Okay, I don't have a strong opinion so will update to not use it if I'm the only one who wants the else.

network_test.go
```diff
@@ -0,0 +1,410 @@
...
+			Mask: net.IPv4Mask(255, 255, 255, 0),
+		},
+		Gateway:     net.IPv4(10, 0, 0, 1),
+		Nameservers: []string{"1.1.1.1", "8.8.8.8"},
+	}
+
+	// IPv6 is currently invalid
+	invalidIPConfiguration = &IPConfiguration{
+		IPAddr: net.IPNet{
+			IP:   net.ParseIP("2001:db8:a0b:12f0::2"),
```
samuelkarp: It would be good to indicate that the 2001:DB8::/32 prefix is [reserved for documentation](https://tools.ietf.org/html/rfc3849).

network_test.go
```diff
@@ -0,0 +1,410 @@
...
+	kernelArgsNoIP   = parseKernelArgs("foo=bar this=phony")
+	kernelArgsWithIP = parseKernelArgs("foo=bar this=phony ip=whatevz")
+
+	validIPConfiguration = &IPConfiguration{
+		IPAddr: net.IPNet{
+			IP:   net.IPv4(10, 0, 0, 2),
+			Mask: net.IPv4Mask(255, 255, 255, 0),
+		},
+		Gateway:     net.IPv4(10, 0, 0, 1),
+		Nameservers: []string{"1.1.1.1", "8.8.8.8"},
```
samuelkarp: Is there a particular reason you're using public nameservers as opposed to other [RFC 1918](https://tools.ietf.org/html/rfc1918) addresses?

README.md
samuelkarp: Sounds good, but can you make it clear that the file is required for now?

README.md
samuelkarp: It would be good to address that point in the bullets here

kernelArgs_test.go
samuelkarp: You might be able to reduce dev-test cycles by having both failures shown, but I don't have strong opinions about this.

sipsma:
I saw a previously unencountered failure in CI where machines only responded to 2/3 pings. I tried just adding a sleep assuming they were just taking a second to boot but I believe the issue was actually that I pushed multiple fixup commits close together, resulting in tests running concurrently (which will give some VMs the same IP because each test suite is using its own IPAM cache).

The most ideal possible fix for this IMO is to run each test in an isolated container similar to firecracker-containerd, but that's a rather large undertaking, so for now I will just have the IPAM cache be shared between tests. None of our code uses IPAM directly, it's only used by the third party `ptp`+`host-local` plugin which is pre-installed on the host, so I think it's alright to have that be shared between tests. Let me know if there's any other thoughts

network.go
sipsma: The problem is that the CNI spec doesn't say to my knowledge whether the order of the args matters or not. If we use a map we are assuming order doesn't matter, which may or may not be true for arbitrary CNI plugins.

machine.go
kzys: I see. And the socket file won't be there at the time NewMachine is calling this function.

README.md
sipsma: I'll add `the name of the interface is determined by the Guest OS, not the SDK`

README.md
sipsma: Sure, in the section above I mention where the SDK looks for the CNI configuration files, I'll add an explicit sentence about it there

network_test.go
kzys: Or [IPv4 Address Blocks Reserved for Documentation](https://tools.ietf.org/html/rfc5737)?

In the meantime, I can quickly recognize 8.8.8.8 and 1.1.1.1 as DNS servers, which is good, while `Nameservers:` pretty much explain that though.

sipsma: These values are not used in e2e tests or any time a real machine is spun up, they are just used for validation unit tests.

However, I see the point that it's possible for someone to inadvertently use them in actual e2e tests in the future, so it's probably best to change them. I'll update to IPs from the range @kzys linked to, which seems safest.

network_test.go
sipsma: Will add a comment

network.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	// Try deleting the network in case it already exists, which can happen if there is
+	// a crash before cleanup from a previous invocations. If it doesn't exist,
+	// well-behaved CNI plugins should treat this as a no-op without returning an error.
+	// We can be reasonably sure any previous VM that was using this network is gone due
+	// to earlier validation that the VM's socket path does not already exist.
+	err = delNetworkFunc()
+	if err != nil {
+		logger.Errorf("failed to delete any pre-existing CNI network %+v: %v", cniConf, err)
```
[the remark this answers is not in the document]
sipsma: I may need to clarify the comment and update this.

Well-behaved plugins will not return any error from their delete method if the network doesn't exist, in which case `err` here will be nil. What's happening here is that they can encounter some other actual problem in the process of deleting things and actually fail to delete something.

Now that I've given this a second thought, I actually don't think this should be a log at all, it should return the error since an error may mean we are trying to create a network that was only partially deleted. I will update to do that.

machine.go
sipsma: So, I was going over everything again and just realized this logic was subtly wrong anyways. What we wanted was to set the parent path using the output of `ResolveSymlinks` only if err was nil, but because of the if statement it was still using the output of `ResolveSymlinks` when the path didn't exist.

I fixed it now and it invalidated this concern.

sipsma:
I started giving some more thought to the whole `uniqueID` thing after realizing there was a subtle bug with the logic. I fixed the bug but ended up concluding that even though it worked the whole idea was just fundamentally too complicated for little benefit and switched to a slightly different approach. Now, the user can optionally provide a VMID with their Machine configuration. If they don't, a uuid4 random ID will be generated for it.

The whole point of the unique ID was to be able to provide an ID for the `ContainerID` in CNI and to use for a network namespace path in the case where we create one for the user. My thinking at the time was that using the socket path would be a good way of uniquely identifying the VM, but obviously that ended up just resulting in a lot of tricky cases to handle. Just generating a random ID is a lot simpler and gives the same end result. Let me know what you think, the change didn't really impact anything else, it's in this fixup commit: https://github.com/firecracker-microvm/firecracker-go-sdk/pull/125/commits/b460583e076d4876e0f7fd79b82a5a1a92447733

sipsma:
> We also have a some ID generation logic in somewhere (maybe in firecracker-contained side). We probably can unify both of them in future.

Yep! I've updated my firecracker-containerd branch to set it's `vmID` as the Machine's `vmID` when creating it w/ the Go SDK, so they should be the same now when using firecracker-containerd.

network_test.go
```diff
@@ -0,0 +1,409 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
samuelkarp: 2019

samuelkarp: Typically everything that we modify in 2019 should have "-2019" to extend the range or have 2019 included.  I just noticed this one since it was a new file.

kernelargs.go
```diff
@@ -0,0 +1,58 @@
...
+		if value != nil {
+			field += "=" + *value
+		}
+		fields = append(fields, field)
+	}
+	return strings.Join(fields, " ")
+}
+
+// deserialize the provided string to a kernelArgs map
+func parseKernelArgs(rawString string) kernelArgs {
```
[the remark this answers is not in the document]
samuelkarp: Fair, and I did say it wasn't required :smile:

Issue #132, opened 2019-09-19
Support customizing tap device uid/gid in tc-redirect-tap plugin


sipsma:
[Right now it's not possible for users of tc-redirect-tap to set a specific tap uid/gid to use; it will always be created with the uid/gid of the caller](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/e4d1f825a38ade07bd1e9dc2bfd64abb7f0331f2/cni/cmd/tc-redirect-tap/main.go#L126-L128).

We should add support for customizing the uid/gid. We need to decide if it is best added as a CNI_ARG or a parameter in the CNI configuration file.

Zyqsempai:
I will work on this one.

sipsma:
Great, this will be fairly similar to what I described in [the issue for supporting tap name](https://github.com/firecracker-microvm/firecracker-go-sdk/issues/131#issuecomment-540245061), but instead adding support for overriding [the uid/gid fields here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/5fb4eeae40c2d36188185e5fda3eb7d306e51b96/cni/cmd/tc-redirect-tap/main.go#L126-L128).

This again will require defining keys to use for these fields, something like `TC_REDIRECT_TAP_UID` and `TC_REDIRECT_TAP_GID` should be good I think.

kzys:
Thanks @Zyqsempai. Resolving this one too 💯

Issue #131, opened 2019-09-19
Support customizing tap device name in tc-redirect-tap plugin


sipsma:
[Right now it's not possible for users of `tc-redirect-tap` to set a specific tap device name to use; it will always be created with the name the kernel decides (usually `tap0`)](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/e4d1f825a38ade07bd1e9dc2bfd64abb7f0331f2/cni/cmd/tc-redirect-tap/main.go#L124).

We should add support for that. We need to decide if it is best added as a `CNI_ARG` or a parameter in the CNI configuration file.

Zyqsempai:
I will take this one!

sipsma:
Sounds great @Zyqsempai ! I took a look at this again, my thinking right now is that it would be best to start with support for configuring the tap name through the CNI args rather than the CNI configuration file, which is probably more flexible for users.

So what that means is checking if the CNI args specified a tap name and, if so, setting it in [the plugin configuration here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/cni/cmd/tc-redirect-tap/main.go#L124). The CNI args are accessible in the `args.Args` field of [the object provided to us here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/5fb4eeae40c2d36188185e5fda3eb7d306e51b96/cni/cmd/tc-redirect-tap/main.go#L94). We will need to define a key name for specifying the tap name, something simple but namespaced like `TC_REDIRECT_TAP_NAME` is probably sufficient.

Let me know if that makes sense! Thanks again.

kzys:
Thanks @Zyqsempai! Resolving.

PR #142, opened 2019-10-15
Added possibility to set tapName, tapUID and tapGID from CNI cmd args

Zyqsempai:
Signed-off-by: bpopovschi <zyqsempai@mail.ru>

*Issue #131 , #132 *

*Description of changes:*
Added possibility to set `tapName`, `tapUID` and `tapGID` from CNI cmd args
Added test for `newPlugin` function

### DIFF
--- a/cni/cmd/tc-redirect-tap/main.go
+++ b/cni/cmd/tc-redirect-tap/main.go
@@ -15,7 +15,10 @@ package main
 
 import (
 	"encoding/json"
+	pluginargs "github.com/firecracker-microvm/firecracker-go-sdk/cni/cmd/tc-redirect-tap/args"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
@@ -101,7 +104,7 @@ func newPlugin(args *skel.CmdArgs) (*plugin, error) {
 		// It's valid for the netns to no longer exist during DEL commands (in which case DEL is
 		// a noop). Thus, we leave validating that netNS is not nil to command implementations.
 		switch err.(type) {
-		case *ns.NSPathNotExistErr:
+		case ns.NSPathNotExistErr:
 			netNS = nil
 		default:
 			return nil, errors.Wrapf(err, "failed to open netns at path %q", args.Netns)
@@ -118,14 +121,10 @@ func newPlugin(args *skel.CmdArgs) (*plugin, error) {
 		}
 	}
 
-	return &plugin{
+	plugin := &plugin{
 		NetlinkOps: internal.DefaultNetlinkOps(),
-
-		// TODO(sipsma) support customizing tap name through args
-
-		// TODO(sipsma) support customizing tap uid/gid through args
-		tapUID: os.Geteuid(),
-		tapGID: os.Getegid(),
+		tapUID:     os.Geteuid(),
+		tapGID:     os.Getegid(),
 
 		// given the use case of supporting VMs, we call the "containerID" a "vmID"
 		vmID: args.ContainerID,
@@ -135,7 +134,33 @@ func newPlugin(args *skel.CmdArgs) (*plugin, error) {
 		netNS: netNS,
 
 		currentResult: currentResult,
-	}, nil
+	}
+	parsedArgs, err := extractArgs(args.Args)
+	if err != nil {
+		return nil, err
+	}
+
+	if tapName, wasDefined := parsedArgs[pluginargs.TCRedirectTapName]; wasDefined {
+		plugin.tapName = tapName
+	}
+
+	if tapUIDVal, wasDefined := parsedArgs[pluginargs.TCRedirectTapUID]; wasDefined {
+		tapUID, err := strconv.Atoi(tapUIDVal)
+		if err != nil {
+			return nil, errors.Wrapf(err, "tapUID should be numeric convertible, got %q", tapUIDVal)
+		}
+		plugin.tapUID = tapUID
+	}
+
+	if tapGIDVal, wasDefined := parsedArgs[pluginargs.TCRedirectTapGUID]; wasDefined {
+		tapGID, err := strconv.Atoi(tapGIDVal)
+		if err != nil {
+			return nil, errors.Wrapf(err, "tapGID should be numeric convertible, got %q", tapGIDVal)
+		}
+		plugin.tapGID = tapGID
+	}
+
+	return plugin, nil
 }
 
 func getCurrentResult(args *skel.CmdArgs) (*current.Result, error) {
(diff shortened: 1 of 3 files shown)


Zyqsempai:
@sipsma PTAL, i decided to add a test for `newPlugin` but I struggling with `netns` check, do you have any idea how we can trick it?

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -376,3 +392,15 @@ type NoPreviousResultError struct{}
 func (e NoPreviousResultError) Error() string {
 	return "no previous result was found, was this plugin chained with a previous one?"
 }
+
+func extractArgs(args, argName string) string {
```
xibz: Can we add a little documentation to `extractArgs` as well as what format is expected?

xibz: Can we also add a test for this?

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -376,3 +392,15 @@ type NoPreviousResultError struct{}
...
+	argumentsPairs := strings.Split(args, ";")
```
xibz: Do we care about multiple pairs?

Zyqsempai:
Hey guys, thanks for your replies, I am going into vacation for a week, so this PR will be frozen for some time, but don't worry, I will finish it as soon as possible after my return.

xibz:
@Zyqsempai - Thank you for keeping us informed! And we really appreciate the time and effort spent on the PR. Please let us know if you have any question or comments about any of our feedback.

sipsma:
> Hey guys, thanks for your replies, I am going into vacation for a week, so this PR will be frozen for some time, but don't worry, I will finish it as soon as possible after my return.

+1 to @xibz's comment, no worries and appreciate all your help on this!

For the issue of testing `newPlugin`, I think the simplest thing we can do is take advantage of the fact that it's not an error to provide `newPlugin` a path to a netns that doesn't exist. If that happens, it should just ignore the error and move on, letting the rest of the function be tested normally. So if we just provide it a netns path like `filepath.Join("/tmp", strconv.Itoa(int(time.Now().UnixNano())))` which for sure doesn't exist, then we should be able to run the unit test as non-root in theory.

So maybe give that a try and if it turns out to not work we can try something else like breaking `newPlugin` up into multiple functions and just unit test that parts that can be unit tested w/out becoming root.

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -376,3 +390,18 @@ type NoPreviousResultError struct{}
...
 	return "no previous result was found, was this plugin chained with a previous one?"
 }
+
+// extractArgs returns desired parameter value from cli args string
+// args string - cli args string("key1=val1;key2=val2)
+// argName string - desired args key name(key1)
+func extractArgs(args, argName string) string {
+	argumentsPairs := strings.Split(args, ";")
+	for _, pairStr := range argumentsPairs {
+		pair := strings.Split(pairStr, "=")
```
nmeyerhans: Is the simple string splitting we're doing here sufficiently robust against embedded delimiter characters? In other words, if we're looking for "key1=val1;key2=val2", what happens if val1 contains a semicolon or an equal sign?

I'm perfectly happy to be told "This is not possible" or something to that effect. Just want to make sure that something is actually enforcing that, though, if it is the case.

Zyqsempai:
@sipsma @xibz Hey guys, finally I am back, and the good news, I finally fixed that test!

cni/cmd/tc-redirect-tap/main.go
Zyqsempai: Actually it's a good question, do you have any idea how we can be sure that THIS will not happen?

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -135,7 +137,21 @@ func newPlugin(args *skel.CmdArgs) (*plugin, error) {
...
 
 		currentResult: currentResult,
-	}, nil
+	}
+
+	tapName := extractArgs(args.Args, TC_REDIRECT_TAP_NAME)
+	if tapName != "" {
+		plugin.tapName = tapName
+	}
+	tapUID, _ := strconv.Atoi(extractArgs(args.Args, TC_REDIRECT_TAP_UID))
```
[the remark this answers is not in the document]
Zyqsempai: Actually, in current implementation, it's not possible without any intervention in behavior, since those params are not mandatory, we cannot  return errors in case they just wasn't provided, but this will happen coz by default extractArgs returns "" and this will not be properly converted in int with atoi function.

cni/cmd/tc-redirect-tap/main.go
sipsma: This string is provided to us by the CNI caller as defined in [the spec](CNI_ARGS) (search for `CNI_ARGS`).

`;` is defined as the delimiter there, so I guess the spec is also implicitly saying that `;` would be invalid in an actual key/value. So I don't think we can do much more in our code if the caller provides a key/val that contains `;`, it would just be user error but not detectable-in-general by us. The best we can do is validate that the arguments we are looking for are in the format expected.

It might be good to mention this in the comment for this function and link to the CNI spec.

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -15,7 +15,10 @@ package main
 
 import (
 	"encoding/json"
+	args2 "github.com/firecracker-microvm/firecracker-go-sdk/cni/cmd/tc-redirect-tap/args"
```
sipsma: nit: i'd prefer we rename this import to something like "pluginargs" which is a little more descriptive

cni/cmd/tc-redirect-tap/main.go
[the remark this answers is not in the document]
sipsma: Right, so we need to distinguish between
1. the arg being provided and valid
2. the arg being provided but invalid
3. the arg not being provided at all

Maybe a good way to implement this would be to generalize the approach and have `extractArgs` just convert the whole `args` string into a `map[string]string` of the key/val pairs and return that map. Then we can use the map's builtin ability to test whether a key is present, something like:
```
parsedArgs := extractArgs(args.Args)

var err error
var tapUID int
if tapUIDVal, wasDefined := parsedArgs[TCRedirectTapName]; wasDefined {
    tapUID, err = strconv.Atoi(tapUIDVal)
    if err != nil {
       // return error saying that tapUID was an invalid format 
    }
}
```

Zyqsempai:
@sipsma PTAL

cni/cmd/tc-redirect-tap/main.go
```diff
@@ -376,3 +398,18 @@ type NoPreviousResultError struct{}
...
+// extractArgs returns cli args in form of map of strings
+// args string - cli args string("key1=val1;key2=val2)
+func extractArgs(args string) map[string]string {
+	result := make(map[string]string)
+	argumentsPairs := strings.Split(args, ";")
+	for _, pairStr := range argumentsPairs {
+		pair := strings.Split(pairStr, "=")
+		if len(pair) > 1 {
+			result[pair[0]] = pair[1]
+		}
```
sipsma: Just some very minor additional corner cases to handle:
* If `extractArgs` was provided `"KEY=foo=blah"`, then we'd get a map where `KEY` is set to `foo` and the rest of the arg val is ignored. I just tried, it apparently is allowed to use an `=` character in a tap device name, so this is something we should handle correctly even though it's extremely obscure.
* If we were provided something like `"TC_REDIRECT_TAP_UID;TC_REDIRECT_TAP_GID"`, we'd just ignore the args and use defaults, which is a bit surprising for users because `"TC_REDIRECT_TAP_UID=;TC_REDIRECT_TAP_GID="` would result in an error.

I think both of these issues can be addressed by using `strings.SplitN(pairStr, "=", 2)` and then returning an error from this function if `len(pair) != 2` (just saying that the arg string is in an invalid format).

Zyqsempai:
@sipsma One more review please!

Zyqsempai:
@sipsma Done!

PR #144, opened 2019-10-15
Prepare files under testdata/ from Makefile

kzys:
*Issue #, if available:*

NA

*Description of changes:*

The testdata/ directory must have vmlinux, root-drive.img and firecracker
binary, in order to run Go tests.

This change prepares the files by just running "make".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,18 +13,42 @@
 
 # Set this to pass additional commandline flags to the go compiler, e.g. "make test EXTRAGOARGS=-v"
 EXTRAGOARGS:=
+DISABLE_ROOT_TESTS?=1
+
+# The below files are needed and can be downloaded from the internet
+testdata_objects = testdata/vmlinux testdata/root-drive.img testdata/firecracker
+
+# --location is needed to follow redirects on github.com
+curl = curl --location
 
 all: build
 
 test: all-tests
 
-unit-tests:
-	go test -short ./... $(EXTRAGOARGS)
+unit-tests: $(testdata_objects) check-kvm
+	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test -short ./... $(EXTRAGOARGS)
 
-all-tests:
-	go test ./... $(EXTRAGOARGS)
+all-tests: $(testdata_objects) check-kvm
+	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test ./... $(EXTRAGOARGS)
+
+check-kvm:
+	@test -w /dev/kvm || \
+		(echo "In order to run firecracker, $(shell whoami) must have write permission to /dev/kvm"; false)
 
 generate build clean:
 	go $@ $(EXTRAGOARGS)
 
-.PHONY: all generate clean build test unit-tests all-tests
+distclean: clean
+	rm -rf $(testdata_objects)
+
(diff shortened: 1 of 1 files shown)


prepare-testdata.sh
```diff
@@ -0,0 +1,29 @@
...
+    local from to
+    from="$1"
+    to="$2"
+
+    if [ ! -e "$to" ]; then
+        curl -fsSL -o "$to" "$from"
+    fi
+}
+
+error() {
```
mxpv: How this fn used?

kzys: /dev/kvm check was there, and I forgot to remove the error function when I moved that from here to Makefile. Let me remove...

kzys: https://github.com/firecracker-microvm/firecracker-go-sdk/pull/144/commits/3721c6919236d1c695c712e1bb082cf79bedb404 doesn't have the function.

Makefile
```diff
@@ -13,18 +13,24 @@
...
-	go test -short ./... $(EXTRAGOARGS)
+unit-tests: test-setup
+	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test -short ./... $(EXTRAGOARGS)
 
-all-tests:
-	go test ./... $(EXTRAGOARGS)
+all-tests: test-setup
+	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test ./... $(EXTRAGOARGS)
+
+test-setup:
```
samuelkarp: You can have make track the files so it knows whether or not to run this target each time.  Something like this:

```make
testdata_objects = testdata/vmlinux testdata/root-drive.img testdata/firecracker
test-setup: $(testdata_objects)
	@test -w /dev/kvm || \
		(echo "In order to run firecracker, $(shell whoami) must have write permission to /dev/kvm"; false)

$(testdata_objects):
	./prepare-testdata.sh
```

kzys: If we use make to track the files, we need to pass the target to the shell script to only download the missing file.

```
$(testdata_objects):
	./prepare-testdata.sh $@
```

Another path we could take is having each target separately.

```
testdata/vmlinux:
        curl -fsSL -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin

```

kzys: Removed the shell script and used `make` for tracking existing files on https://github.com/firecracker-microvm/firecracker-go-sdk/pull/144/commits/c210e9bf1f8a2da501c21fe240463ab8b23bb3c8

Makefile
```diff
@@ -13,18 +13,42 @@
...
+	$(curl) -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin
+
+testdata/firecracker:
+	$(curl) -o $@ https://github.com/firecracker-microvm/firecracker/releases/download/v0.18.0/firecracker-v0.18.0
+	chmod +x $@
+
+testdata/root-drive.img:
+	$(curl) -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/fsfiles/hello-rootfs.ext4
+
+.PHONY: all generate clean build test unit-tests all-tests check-kvm
```
kzys: distclean should be there...

PR #145, opened 2019-10-19
Fixes issue where FifoLogWriter would hang

### LINKED ISSUE #143: Startup Hangs when using LogWriter
This pertains to PR #134 I worked on and it introduced the issue. 

When it was suggested to move `captureFifoToFile` to the handler, I forgot about the fact that if Firecracker is not started before we start listening on the log.fifo then `os.Open` in [captureFifoToFile](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/machine.go#L582) will block. Since we moved `captureFifoToFile` to a handler, this blocks the handlers which execute before we start Firecracker so it'll never unblock., shown [here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/machine.go#L329-L334).

Mostly creating this issue to spark a discussion on the best place to put captureFifoToFile. The implementation before the above PR did not work with Jailer as the sdk jailer setup would change the path to the `LogFifo` in the config before we tried to listen on it.

I've outlined workflows from what I understand needs to happen in the case of jailer and plain firecracker.

Firecracker:
- Create `log.fifo` and `metric.fifo`
- Start Firecracker
- Configure logging and metrics via API
- Call `captureFifoToFile`

Jailer:
- Create `log.fifo` and `metric.fifo`
- Jailer chroot for vm setup
- Hard link `log.fifo` and `metric.fifo` into jailer chroot for VM
- Jailer start Firecracker
- Configure logging and metrics via API
- Call `captureFifoToFile` on original `log.fifo` path (not hard linked)

xibz:
Prior to this fix the FifoLogWriter would hang the process due to
waiting for the fifo to be connected to firecracker when attempting to
open the fifo. This fix wraps the capture fifo method in a goroutine.

Fixes #143 

### DIFF
--- a/handlers_test.go
+++ b/handlers_test.go
@@ -1,19 +1,16 @@
 package firecracker
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 	"reflect"
 	"testing"
-	"time"
 
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 	ops "github.com/firecracker-microvm/firecracker-go-sdk/client/operations"
 	"github.com/firecracker-microvm/firecracker-go-sdk/fctesting"
-	"github.com/pkg/errors"
 )
 
 func TestHandlerListAppend(t *testing.T) {
@@ -663,76 +660,6 @@ func TestHandlers(t *testing.T) {
 	}
 }
 
-func TestCreateLogFilesHandler(t *testing.T) {
-	logWriterBuf := &bytes.Buffer{}
-	config := Config{
-		LogFifo:       filepath.Join(testDataPath, "firecracker-log.fifo"),
-		MetricsFifo:   filepath.Join(testDataPath, "firecracker-metrics.fifo"),
-		FifoLogWriter: logWriterBuf,
-	}
-
-	defer func() {
-		os.Remove(config.LogFifo)
-		os.Remove(config.MetricsFifo)
-	}()
-
-	ctx := context.Background()
-	m, err := NewMachine(ctx, config, WithLogger(fctesting.NewLogEntry(t)))
-	if err != nil {
-		t.Fatalf("failed to create machine: %v", err)
-	}
-
-	// spin off goroutine to write to Log fifo so we don't block
-	doneChan := make(chan struct{}, 1)
-	go func() {
-		defer close(doneChan)
-
-		// try to open file
-		fifoPipe, err := openFileRetry(config.LogFifo)
-		if err != nil {
-			t.Error(err)
-		}
-
-		if _, err := fifoPipe.WriteString("data\n"); err != nil {
-			t.Errorf("Failed to write to fifo %v", err)
-		}
-
-		fifoPipe.Close()
-		return
-	}()
-
-	// Execute Handler
-	if err := CreateLogFilesHandler.Fn(ctx, m); err != nil {
-		t.Errorf("failed to call CreateLogFilesHandler function: %v", err)
-		return
-	}
-
-	// Block until writing go routine is done to check data that was written
-	<-doneChan
-
-	// Poll for verifying logs were written as we need to allow time
-	// for copying from the log fifo into the FifoLogWriter
-	timer := time.NewTimer(1 * time.Second)
-	for {
-		select {
-		case <-timer.C:
-			t.Fatal("timed out reading from log writer")
-		default:
-			logData, err := logWriterBuf.ReadString('\n')
-			if err != nil {
-				time.Sleep(10 * time.Millisecond)
-				continue
-			}
-
-			if logData != "data\n" {
-				t.Errorf("expected 'data' written to log got '%s'", logData)
-			}
-			return
-		}
-	}
-
-}
-
 func compareHandlerLists(l1, l2 HandlerList) bool {
 	if l1.Len() != l2.Len() {
 		return false
@@ -753,22 +680,4 @@ func compareHandlerLists(l1, l2 HandlerList) bool {
 	}
 
 	return true
-}
-
-func openFileRetry(filePath string) (file *os.File, err error) {
-	timer := time.NewTimer(1 * time.Second)
-	for {
-		select {
-		case <-timer.C:
-			err = errors.New("timed out waiting for file")
-			return
-		default:
-			file, err = os.OpenFile(filePath, os.O_WRONLY, os.ModePerm)
-			if err == nil {
-				timer.Stop()
-				return
-			}
-			time.Sleep(10 * time.Millisecond)
-		}
-	}
 }
--- a/machine.go
+++ b/machine.go
@@ -576,28 +576,37 @@ func (m *Machine) setupLogging(ctx context.Context) error {
 	return nil
 }
 
-func captureFifoToFile(logger *log.Entry, fifoPath string, fifo io.Writer) error {
-	// create the fifo pipe which will be used
+func captureFifoToFile(logger *log.Entry, fifoPath string, w io.Writer) error {
+	// open the fifo pipe which will be used
 	// to write its contents to a file.
-	fifoPipe, err := os.OpenFile(fifoPath, os.O_RDONLY, 0600)
+	fd, err := syscall.Open(fifoPath, syscall.O_RDONLY|syscall.O_NONBLOCK, 0600)
 	if err != nil {
 		return fmt.Errorf("Failed to open fifo path at %q: %v", fifoPath, err)
 	}
 
-	if err := syscall.Unlink(fifoPath); err != nil {
-		logger.Warnf("Failed to unlink %s", fifoPath)
+	fifoPipe := os.NewFile(uintptr(fd), fifoPath)
+	if fifoPipe == nil {
+		return fmt.Errorf("Invalid file descriptor")
 	}
 
 	logger.Debugf("Capturing %q to writer", fifoPath)
 
-	// Uses a go routine to do a non-blocking io.Copy. The fifo
-	// file should be closed when the appication has finished, since
-	// the forked firecracker application will be closed resulting
-	// in the pipe to return an io.EOF
+	// Uses a goroutine to copy the contents of the fifo pipe to the io.Writer.
+	// In the event that the goroutine finishes, which is caused by either the
+	// context being closed or the application being closed, we will close the
+	// pipe and unlink the fifo path.
 	go func() {
-		defer fifoPipe.Close()
+		defer func() {
+			if err := fifoPipe.Close(); err != nil {
+				logger.Warnf("Failed to close fifo pipe: %v", err)
+			}
 
-		if _, err := io.Copy(fifo, fifoPipe); err != nil {
+			if err := syscall.Unlink(fifoPath); err != nil {
+				logger.Warnf("Failed to unlink %s: %v", fifoPath, err)
+			}
+		}()
+
+		if _, err := io.Copy(w, fifoPipe); err != nil {
 			logger.Warnf("io.Copy failed to copy contents of fifo pipe: %v", err)
 		}
 	}()
--- a/machine_test.go
+++ b/machine_test.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
@@ -146,7 +147,12 @@ func TestJailerMicroVMExecution(t *testing.T) {
 	socketPath := filepath.Join(jailerTestPath, "firecracker", "TestJailerMicroVMExecution.socket")
 	logFifo := filepath.Join(tmpDir, "firecracker.log")
 	metricsFifo := filepath.Join(tmpDir, "firecracker-metrics")
+	capturedLog := filepath.Join(tmpDir, "writer.fifo")
+	fw, err := os.OpenFile(capturedLog, os.O_CREATE|os.O_RDWR, 0600)
+	require.NoError(t, err, "failed to open fifo writer file")
 	defer func() {
+		fw.Close()
+		os.Remove(capturedLog)
 		os.Remove(socketPath)
 		os.Remove(logFifo)
 		os.Remove(metricsFifo)
@@ -183,6 +189,7 @@ func TestJailerMicroVMExecution(t *testing.T) {
 			ExecFile:       getFirecrackerBinaryPath(),
 			ChrootStrategy: NewNaiveChrootStrategy(jailerFullRootPath, vmlinuxPath),
 		},
+		FifoLogWriter: fw,
 	}
 
 	if _, err := os.Stat(vmlinuxPath); err != nil {
@@ -250,6 +257,10 @@ func TestJailerMicroVMExecution(t *testing.T) {
 	}
 
 	m.StopVMM()
+
+	info, err := os.Stat(capturedLog)
+	assert.NoError(t, err, "failed to stat captured log file")
+	assert.NotEqual(t, 0, info.Size())
 }
 
 func TestMicroVMExecution(t *testing.T) {
@@ -263,7 +274,12 @@ func TestMicroVMExecution(t *testing.T) {
 	socketPath := filepath.Join(testDataPath, "TestMicroVMExecution.sock")
 	logFifo := filepath.Join(testDataPath, "firecracker.log")
 	metricsFifo := filepath.Join(testDataPath, "firecracker-metrics")
+	capturedLog := filepath.Join(testDataPath, "writer.fifo")
+	fw, err := os.OpenFile(capturedLog, os.O_CREATE|os.O_RDWR, 0600)
+	require.NoError(t, err, "failed to open fifo writer file")
 	defer func() {
+		fw.Close()
+		os.Remove(capturedLog)
 		os.Remove(socketPath)
 		os.Remove(logFifo)
 		os.Remove(metricsFifo)
@@ -292,6 +308,7 @@ func TestMicroVMExecution(t *testing.T) {
 		Debug:             true,
 		DisableValidation: true,
 		NetworkInterfaces: networkIfaces,
+		FifoLogWriter:     fw,
 	}
 
 	ctx := context.Background()
@@ -353,6 +370,10 @@ func TestMicroVMExecution(t *testing.T) {
 	// didn't for some reason, we still need to terminate it:
 	m.StopVMM()
 	m.Wait(vmmCtx)
+
+	info, err := os.Stat(capturedLog)
+	assert.NoError(t, err, "failed to stat captured log file")
+	assert.NotEqual(t, 0, info.Size())
 }
 
 func TestStartVMM(t *testing.T) {
@@ -774,7 +795,7 @@ func TestLogFiles(t *testing.T) {
 }
 
 func TestCaptureFifoToFile(t *testing.T) {
-	fifoPath := filepath.Join(testDataPath, "fifo")
+	fifoPath := filepath.Join(testDataPath, "TestCaptureFifoToFile")
 
 	if err := syscall.Mkfifo(fifoPath, 0700); err != nil {
 		t.Fatalf("Unexpected error during syscall.Mkfifo call: %v", err)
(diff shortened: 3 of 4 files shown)


handlers.go
```diff
@@ -152,9 +152,13 @@ var CreateLogFilesHandler = Handler{
 		)
 
 		if m.Cfg.FifoLogWriter != nil {
-			if err := captureFifoToFile(m.logger, logFifoPath, m.Cfg.FifoLogWriter); err != nil {
-				m.logger.Warnf("captureFifoToFile() returned %s. Continuing anyway.", err)
-			}
+			// We wrap this in a goroutine since opening the fifo queue will block
+			// until the fifo has been connected when starting firecracker
+			go func() {
+				if err := captureFifoToFile(m.logger, logFifoPath, m.Cfg.FifoLogWriter); err != nil {
```
xibz: There is a goroutine already in `captureFifoToFile`. It may be best to just remove the one in `captureFifoToFile` and make `captureFifoToFile` to be expected to be ran in a goroutine.

handlers.go
```diff
@@ -152,9 +152,13 @@ var CreateLogFilesHandler = Handler{
 		)
 
 		if m.Cfg.FifoLogWriter != nil {
-			if err := captureFifoToFile(m.logger, logFifoPath, m.Cfg.FifoLogWriter); err != nil {
-				m.logger.Warnf("captureFifoToFile() returned %s. Continuing anyway.", err)
-			}
+			// We wrap this in a goroutine since opening the fifo queue will block
+			// until the fifo has been connected when starting firecracker
```
sipsma: I just did an experiment to see what happens when:
1. A fifo is created
2. The fifo is attempted to be opened w/ `O_RDONLY` (causing the open call to block, as expected)
3. The fifo is then unlinked before a writer opens the other side

The behavior I saw, sadly, was that the open `O_RDONLY` did not get unblocked when the fifo was unlinked.

I think this is a problem because this could actually happen with the current code if `captureFifoToFile` is blocked on the open call, something else goes wrong with VM creation causing the create to fail and the fifos to be unlinked before the VMM opens them for reading on the other side. In this case, I think we would end up having a goroutine eternally blocked waiting on the open call.

Here's the code I tested with:
```
package main

import (
        "os"
        "syscall"
        "time"
)

func main() {
        err := syscall.Mkfifo("fifo", 0700)
        if err != nil {
                panic(err)
        }

        go func() {
                time.Sleep(3*time.Second)
                err := syscall.Unlink("fifo")
                if err != nil {
                        panic(err)
                }
        }()

        _, err = os.OpenFile("fifo", os.O_RDONLY, 0600)
        if err != nil {
                panic(err)
        }
}
```

sipsma: As far as possible fixes, maybe one thing to look into is containerd's standalone `fifo` package: https://github.com/containerd/fifo

I believe it allows you to use a `context` object to control the opening of FIFOs, which could be helpful here.

xibz: Bah, that sucks. I do have another solution but it requires a little bit more work, but should be more preferable imo

nmeyerhans: > The behavior I saw, sadly, was that the open `O_RDONLY` did not get unblocked when the fifo was unlinked.
> 

This is not surprising. The named FIFO behaves just like any other file when it's unlinked: the directory entry is removed, and the link count decremented, but processes with open file descriptors are not impacted.

I think this problem comes down to inadequate management of filesystem resources in two different contexts: inside the jail and outside the jail. AIUI, the logic before the jailer change was correct, we just lost some state when we overwrote the configuration with the jailed paths for the FIFOs. I think we might want to provide a context-aware abstraction around file paths within the jail, allowing us to identify the appropriate path from either context.

sipsma: > This is not surprising. The named FIFO behaves just like any other file when it's unlinked: the directory entry is removed, and the link count decremented, but processes with open file descriptors are not impacted.

The part that surprises me is that AFAIK no process actually has the file descriptor open when you run that example I pasted. There is a process trying to open the fifo, but the fifo is unlinked and not opened by anything else I'm aware of. That seems bizarre but maybe it's the expected behavior.

jailer.go
```diff
@@ -358,6 +358,7 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 
 	m.cmd = builder.Build(ctx)
 
+	m.Handlers.FcInit = m.Handlers.FcInit.Swap(createBootstrapLoggingHandler(m.Cfg.LogFifo))
```
sipsma: I'm not super familiar with this aspect of the gosdk, but would it make more sense to put this swap in the call to `ChrootStrategy.AdaptHandlers` call below? On the surface `AdaptHandlers` seems like the right place to update the bootstrapLoggingHandler's log fifo path.

handlers.go
```diff
@@ -174,10 +168,37 @@ var BootstrapLoggingHandler = Handler{
...
+			if err := captureFifoToFile(m.logger, m.Cfg.LogFifo, m.Cfg.FifoLogWriter); err != nil {
+				m.logger.Warnf("captureFifoToFile() returned %s. Continuing anyway.", err)
+			}
+		}
+
 		return nil
 	},
 }
 
+func createBootstrapLoggingHandler(logFifo string) Handler {
```
sipsma: Would it make sense to above this function do:
```
var BootstrapLoggingHandler = Handler{
  Name: BootstrapLoggingHandlerName,
  Fn: func(ctx context.Context, m *Machine) error {
    return createBootstrapLoggingHandler(m.Cfg.LogFifo)
  },
}
``` 

in order to dupe less code?

handlers.go
nmeyerhans: Yeah, POSIX specifies that open() with O_RDONLY will block forever if there isn't a writer. Once the syscall has started, there isn't really a way to short-circuit it if the file disappears. The check for existence has already succeeded and doesn't run again.

If you change your test code to add a second call to OpenFile() with os.O_WRONLY in the goroutine after the sleep, your program will run to completion.

One possibility to avoid blocking would be to add `syscall.O_NONBLOCK` to the open flags, and then retry the operation several times in case open fails with `ENXIO`.

jailer.go
xibz: Changed to just how default handlers are constructed

handlers.go
xibz: Also changed

handlers.go
```diff
@@ -151,31 +151,33 @@ var CreateLogFilesHandler = Handler{
...
-			m.logger.Warnf("setupLogging() returned %s. Continuing anyway.", err)
-		} else {
-			m.logger.Debugf("setup logging: success")
-		}
+func createBootstrapLoggingHandler(logFifo string) Handler {
+	return Handler{
+		Name: BootstrapLoggingHandlerName,
+		Fn: func(ctx context.Context, m *Machine) error {
+			if err := m.setupLogging(ctx); err != nil {
+				m.logger.Warnf("setupLogging() returned %s. Continuing anyway.", err)
```
sipsma: If `setupLogging` returns an error, we should probably return from the function (even if we only log the error and return nil). Otherwise, we can continue on from here to the `captureFifoToFile` call below and get blocked again due to `setupLogging` failing.

Is there any expected reason for `setupLogging` to fail? Would it just be simpler to return any error from it?

xibz: Fixed! Discussed offline but what was decided was to return nil if an error occurs, that we avoid the file potentially being blocked by open.

handlers_test.go
```diff
@@ -663,76 +660,6 @@ func TestHandlers(t *testing.T) {
 	}
 }
 
-func TestCreateLogFilesHandler(t *testing.T) {
```
xibz: This test provides inconsistent results due to the handler now being asynchronous on opening the fifo. However, I realized that this test is testing the fifo writer, which we already have a test for. With that I decided to delete it.

cpheps:
I was wondering if there is a timeline on this PR? We would very much like to get the fixes in this PR and #147 into or project.

xibz:
Hey @cpheps, thank you for your patience! I am just waiting on one additional approval before I merge this in. I will go ahead and ping a couple folks to take a look at this, but we will look at this as soon as we can.

machine.go
```diff
@@ -579,11 +579,16 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
 	// to write its contents to a file.
-	fifoPipe, err := os.OpenFile(fifoPath, os.O_RDONLY, 0600)
+	fd, err := syscall.Open(fifoPath, syscall.O_RDONLY|syscall.O_NONBLOCK, 0600)
 	if err != nil {
 		return fmt.Errorf("Failed to open fifo path at %q: %v", fifoPath, err)
 	}
 
+	fifoPipe := os.NewFile(uintptr(fd), fifoPath)
+	if fifoPipe == nil {
+		return fmt.Errorf("Invalid file descriptor: %v", err)
```
kzys: `os.NewFile()` doesn't set `err`. So we shouldn't include err in the error message since it would be always nil.

machine_test.go
```diff
@@ -250,6 +256,10 @@ func TestJailerMicroVMExecution(t *testing.T) {
 	}
 
 	m.StopVMM()
+
+	info, err := os.Stat(fifoWriterPath)
```
samuelkarp: Why are you stating the writer here?

machine.go
xibz: Fixed!

handlers.go
samuelkarp: @xibz Are you planning to add the retries that @nmeyerhans suggested?

machine_test.go
xibz: just to ensure that the fifo was actually created. Just want to ensure that that logic is correct.

xibz:
> Can you add a test for captureFifoToFile (just by itself, not with anything else) testing that it can open a fifo that hasn't been opened for write yet, and then start actually capturing the content once the fifo is opened for write? I don't see that covered in the tests here.

There is already a test that does that, `TestCaptureFifoToFile`.

machine_test.go
samuelkarp: You're creating the fifo in the test (not the code under test) on line 150.  Why do you then need to assert that it's created?

(Stat doesn't tell you anything about the contents, either.)

samuelkarp:
> There is already a test that does that, `TestCaptureFifoToFile`.

`TestCaptureFifoToFile` opens the file for write (line 784) and writes to it (line 789) prior to calling `captureFifoToFile` (line 806).  I'm asking you to call `captureFifoToFile` before opening the file for write and writing to it (to test that we're using `O_NONBLOCK` correctly).

handlers.go
xibz: @samuelkarp - Not needed as we are opening with `O_RDONLY`.

From the man page
```
A process can open a FIFO in nonblocking mode.  In this case, opening
       for read-only succeeds even if no one has opened on the write side
       yet and opening for write-only fails with ENXIO (no such device or
       address) unless the other end has already been opened.
```

machine_test.go
xibz: Not created, just that something has been written to it

samuelkarp: Okay, the name confused me.  This isn't a fifo, but the destination file where the data passed through a fifo ends up being written (in `captureFifoToFile`).  Can you change the name of this to reflect what it actually is?  Something like capturedLog?

xibz: Done!

machine_test.go
```diff
@@ -781,6 +801,18 @@ func TestCaptureFifoToFile(t *testing.T) {
 	}
 	defer os.Remove(fifoPath)
 
+	fifoLogPath := fifoPath + ".log"
```
samuelkarp: nit: can you call this `logPath` or `destinationPath`?

machine_test.go
```diff
@@ -781,6 +801,18 @@ func TestCaptureFifoToFile(t *testing.T) {
...
+	fifo, err := os.OpenFile(fifoLogPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
```
samuelkarp: This isn't a fifo and shouldn't be called `fifo`.  This is your destination file, so call it `destination` or something like that.

machine_test.go
```diff
@@ -781,6 +801,18 @@ func TestCaptureFifoToFile(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create fifo file: %v", err)
```
samuelkarp: This isn't a fifo.

machine_test.go
```diff
@@ -781,6 +801,18 @@ func TestCaptureFifoToFile(t *testing.T) {
 	}
 	defer os.Remove(fifoPath)
 
+	fifoLogPath := fifoPath + ".log"
+	fifo, err := os.OpenFile(fifoLogPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
```
samuelkarp: nit: You've got `assert` in this file already, so you might as well do `assert.NoError(t, err, "failed to create destination file")`.

machine_test.go
```diff
@@ -781,6 +801,18 @@ func TestCaptureFifoToFile(t *testing.T) {
 	}
 	defer os.Remove(fifoPath)
```
samuelkarp: You've changed the behavior of this test.  I think it'd be good to have tests for both behaviors rather than picking one over the other.

machine.go
```diff
@@ -576,26 +576,35 @@ func (m *Machine) setupLogging(ctx context.Context) error {
 	return nil
 }
 
-func captureFifoToFile(logger *log.Entry, fifoPath string, fifo io.Writer) error {
+func captureFifoToFile(logger *log.Entry, captureFifoPath string, fifo io.Writer) error {
```
samuelkarp: I think you actually do want `fifoPath`; it's the path of the fifo that we're reading.  You could change the `fifo io.Writer` to something like `w io.Writer` or `destination io.Writer`, which I think would be a lot more clear; the writer isn't a fifo and shouldn't be called one (we're reading a fifo, not writing one).

machine.go
```diff
@@ -576,26 +576,35 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
 	}
 
-	logger.Debugf("Capturing %q to writer", fifoPath)
+	logger.Debugf("Capturing %q to writer", captureFifoPath)
 
-	// Uses a go routine to do a non-blocking io.Copy. The fifo
-	// file should be closed when the appication has finished, since
-	// the forked firecracker application will be closed resulting
-	// in the pipe to return an io.EOF
+	// Uses a goroutine to copy the contents of the fifo pipe to the captured log
```
samuelkarp: "io.Writer"

xibz: Fixed!

machine.go
```diff
@@ -576,28 +576,37 @@ func (m *Machine) setupLogging(ctx context.Context) error {
 	return nil
 }
 
-func captureFifoToFile(logger *log.Entry, fifoPath string, fifo io.Writer) error {
+func captureFifoToFile(logger *log.Entry, fifoPath string, w io.Writer) error {
 	// create the fifo pipe which will be used
```
samuelkarp: nit: "open the fifo"

machine.go
xibz: Fixed!

machine_test.go
```diff
@@ -814,6 +834,45 @@ func TestCaptureFifoToFile(t *testing.T) {
...
+
+	go func() {
+		t := time.NewTicker(250 * time.Millisecond)
+		select {
+		case <-t.C:
+			f.Close()
+		}
+	}()
+
+	assert.NoError(t, err, "failed to stat file")
```
samuelkarp: You're missing the actual stat here I think.  You could also use `ioutil.ReadFile` to get the file content and compare it to what you expect ("Hello world!").

machine.go
xibz: Fixed!

machine_test.go
xibz: doh, fixed.

xibz: Also, to test the expected bytes, we'd need to write as soon as `captureFifoToFile` opens the file for reading, otherwise the bytes get lost.

xibz: Alright, added the assert for expected bytes.

machine.go
```diff
@@ -576,28 +576,37 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
 	go func() {
-		defer fifoPipe.Close()
+		defer func() {
+			if err := fifoPipe.Close(); err != nil {
+				logger.Warnf("Failed to close fifo pipe: %v", err)
+			}
+
+			if err := syscall.Unlink(fifoPath); err != nil {
+				logger.Warnf("Failed to unlink %s", fifoPath)
+			}
```
kzys: - syscall.Unlink's error should be logged here.
- How about making both of them `Errorf`? Let's say the FIFO is being removed before reaching this line, wouldn't it indicate that we have a logic error somewhere, like cleaning up files twice?

xibz: Sure, but I think warning is fine, since this isn't really a detrimental piece of code.

machine_test.go
```diff
@@ -797,21 +819,90 @@ func TestCaptureFifoToFile(t *testing.T) {
...
-		t.Errorf("Failed to stat file: %v", err)
+	f, err := os.OpenFile(fifoPath, os.O_RDWR, 0600)
+	if err != nil {
+		t.Fatalf("Failed to open file, %q: %v", fifoPath, err)
 	}
+	expectedBytes := []byte("Hello world!")
+	f.Write(expectedBytes)
+	defer f.Close()
+
+	go func() {
```
samuelkarp: nit: Can you change this to: `time.AfterFunc(250*time.Millisecond, func() { f.Close() })`?  https://golang.org/pkg/time/#AfterFunc

machine_test.go
samuelkarp: @xibz Can you fix this error message?

machine_test.go
xibz: Fixed

machine_test.go
```diff
@@ -812,17 +812,14 @@ func TestCaptureFifoToFile(t *testing.T) {
 	defer f.Close()
 
 	go func() {
-		t := time.NewTicker(250 * time.Millisecond)
-		select {
-		case <-t.C:
-			f.Close()
-		}
+		t := time.AfterFunc(250*time.Millisecond, func() { f.Close() })
+		<-t.C
```
samuelkarp: You don't need this.  You can just do `time.AfterFunc()` without the goroutine and without looking at `t.C`.

machine_test.go
```diff
@@ -890,11 +887,8 @@ func TestCaptureFifoToFile_nonblock(t *testing.T) {
 	defer f.Close()
 
 	go func() {
-		t := time.NewTicker(250 * time.Millisecond)
-		select {
-		case <-t.C:
-			f.Close()
-		}
+		t := time.AfterFunc(250*time.Millisecond, func() { f.Close() })
```
samuelkarp: Same here.

machine_test.go
samuelkarp: You don't need the goroutine and you don't need to drain `t.C`:

> AfterFunc waits for the duration to elapse and then calls f in its own goroutine. It returns a Timer that can be used to cancel the call using its Stop method.

xibz: Sorry, thought you just wanted to swap the lines. Fixed.

samuelkarp:
Can you fix your commit history?  You've got some issues in it:

* https://github.com/firecracker-microvm/firecracker-go-sdk/pull/145/commits/542e20c16bfa0c9361cd725be8e009e29f2a6fc6 is marked as a fixup of https://github.com/firecracker-microvm/firecracker-go-sdk/pull/145/commits/445c8758fd419ad9b36ec55723f88c1e6d4ec04c but should really be a fixup of https://github.com/firecracker-microvm/firecracker-go-sdk/pull/145/commits/08f6e0887b33db17ecc9bc55627ea9be9df9fff9
* https://github.com/firecracker-microvm/firecracker-go-sdk/pull/145/commits/61b460be8768fccd3ec629cc1bb1aaefa6ebf79d is marked as a fixup of https://github.com/firecracker-microvm/firecracker-go-sdk/pull/145/commits/445c8758fd419ad9b36ec55723f88c1e6d4ec04c but should really be a fixup of https://github.com/firecracker-microvm/firecracker-go-sdk/pull/145/commits/08f6e0887b33db17ecc9bc55627ea9be9df9fff9

xibz:
> Can you fix your commit history? You've got some issues in it:

Done!

PR #147, opened 2019-10-23
Fixes leak of open file during capture fifo

### LINKED ISSUE #146: goroutine within captureFifoToFile may leak
I am under the suspicion that this [goroutine](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/5fb4eeae40c2d36188185e5fda3eb7d306e51b96/machine.go#L597-L603) may be leaking if the file has been deleted or unlinked. More testing may be needed to verify this.

xibz:
This change fixes a leak that occurs when using context cancel or
stopping the VM. The goroutine in `captureFifoToFile` spins up a
goroutine to copy the contents of the file, but will continue reading
from the file even if the VM shuts down. To prevent this, we now look at
the context to see if it has finished or if the machine has closed its
exit channel

Fixes #146 

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -576,7 +576,7 @@ func (m *Machine) setupLogging(ctx context.Context) error {
 	return nil
 }
 
-func captureFifoToFile(logger *log.Entry, fifoPath string, w io.Writer) error {
+func (m *Machine) captureFifoToFile(logger *log.Entry, fifoPath string, w io.Writer) error {
 	// open the fifo pipe which will be used
 	// to write its contents to a file.
 	fd, err := syscall.Open(fifoPath, syscall.O_RDONLY|syscall.O_NONBLOCK, 0600)
@@ -590,6 +590,16 @@ func captureFifoToFile(logger *log.Entry, fifoPath string, w io.Writer) error {
 	}
 
 	logger.Debugf("Capturing %q to writer", fifoPath)
+
+	// this goroutine is to track the life of the application along with whether
+	// or not the context has been cancelled which is signified by the exitCh. In
+	// the event that the exitCh has been closed, we will close the fifo file.
+	go func() {
+		<-m.exitCh
+		if err := fifoPipe.Close(); err != nil {
+			logger.WithError(err).Debug("failed to close fifo")
+		}
+	}()
 
 	// Uses a goroutine to copy the contents of the fifo pipe to the io.Writer.
 	// In the event that the goroutine finishes, which is caused by either the
--- a/machine_test.go
+++ b/machine_test.go
@@ -14,6 +14,7 @@
 package firecracker
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"flag"
@@ -30,6 +31,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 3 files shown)


machine_test.go
```diff
@@ -939,3 +944,50 @@ func TestPID(t *testing.T) {
...
+	logger.Logger.Level = logrus.WarnLevel
+	logger.Logger.Out = loggerBuffer
+	err = m.captureFifoToFile(ctx, logger, fifoPath, buf)
+	assert.NoError(t, err, "failed to capture fifo to file")
+	select {
+	case <-m.exitCh:
+	case <-ctx.Done():
+	}
+
+	// wait sometime for the logs to populate
```
kzys: Waiting because of goroutines?

machine_test.go
```diff
@@ -939,3 +944,50 @@ func TestPID(t *testing.T) {
...
+	loggerBuffer := bytes.NewBuffer(nil)
+	logger := fctesting.NewLogEntry(t)
+	logger.Logger.Level = logrus.WarnLevel
+	logger.Logger.Out = loggerBuffer
+	err = m.captureFifoToFile(ctx, logger, fifoPath, buf)
+	assert.NoError(t, err, "failed to capture fifo to file")
+	select {
+	case <-m.exitCh:
+	case <-ctx.Done():
+	}
```
kzys: Doesn't it always go to ctx.Done() branch after 1 second? How about explicitly calling cancel() or put something on the channel instead of waiting?

machine_test.go
```diff
@@ -939,3 +944,50 @@ func TestPID(t *testing.T) {
...
+
+	// wait sometime for the logs to populate
+	time.Sleep(250 * time.Millisecond)
+	expected := `io.Copy failed to copy contents of fifo pipe: read testdata/TestCaptureFifoToFileLeak.fifo: file already closed`
+	assert.True(t, strings.Contains(loggerBuffer.String(), expected))
```
kzys: Nit: assert has `Contains`

machine_test.go
xibz: Yea

machine_test.go
xibz: Yea, I can call cancel. Thanks!

machine.go
```diff
@@ -590,6 +590,22 @@ func captureFifoToFile(logger *log.Entry, fifoPath string, fifo io.Writer) error
 
 	logger.Debugf("Capturing %q to writer", fifoPath)
 
+	// this goroutine is to track the life of the application along with whether
+	// or not the context has been cancelled. In the event that the context has
+	// been cancel or the application has closed, we will close the fifo file
+	go func() {
+		select {
```
samuelkarp: This can be written more simply:

```go
select {
	case <-ctx.Done():
	case <-m.exitCh:
}
if err := fifoPipe.Close(); err != nil {
	logger.WithError(err).Debug("failed to close file")
}
```

machine_test.go
xibz: Fixed!

machine.go
xibz: Fixed.

machine.go
```diff
@@ -590,6 +590,20 @@ func captureFifoToFile(logger *log.Entry, fifoPath string, fifo io.Writer) error
...
+		case <-ctx.Done():
```
sipsma: Is there a reason to also use the `ctx` here instead of just using `m.exitCh` exclusively? IIUC the `ctx` passed here will ultimately [be passed in via `machine.Start` here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/437f90243b6d40db06333829a6dc2c38e7570af4/machine.go#L308), which is also the same [`ctx` passed to `startVMM`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/437f90243b6d40db06333829a6dc2c38e7570af4/handlers.go#L118), which in turn is [a `ctx` used to determine when to close `exitCh`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/437f90243b6d40db06333829a6dc2c38e7570af4/machine.go#L503-L514).

It doesn't break anything to use `ctx` here but it does increase the complexity for readers to go through to determine when this select will get unblocked, so all else being equal I'd prefer we only use `exitCh`.

xibz: Sure, fixed. They have to go through that logic anyway to determine that `exitCh` is closed when the context is closed. So, I don't think it adds anymore complexity then what they already have to do, however.

PR #153, opened 2019-10-31
Report its exit status from Wait() correctly

kzys:
*Issue #, if available:*

NA

*Description of changes:*

`err` was actually the variable outside of the goroutine.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -173,8 +173,8 @@ type Machine struct {
 	startOnce sync.Once
 	// exitCh is a channel which gets closed when the VMM exits
 	exitCh chan struct{}
-	// err records any error executing the VMM
-	err error
+	// fatalErr records an error that either stops or prevent starting the VMM
+	fatalErr error
 
 	// callbacks that should be run when the machine is being torn down
 	cleanupOnce  sync.Once
@@ -349,7 +349,7 @@ func (m *Machine) Wait(ctx context.Context) error {
 	case <-ctx.Done():
 		return ctx.Err()
 	case <-m.exitCh:
-		return m.err
+		return m.fatalErr
 	}
 }
 
@@ -444,7 +444,7 @@ func (m *Machine) startVMM(ctx context.Context) error {
 	if err != nil {
 		m.logger.Errorf("Failed to start VMM: %s", err)
 
-		m.err = err
+		m.fatalErr = err
 		close(m.exitCh)
 
 		return err
@@ -501,7 +501,7 @@ func (m *Machine) startVMM(ctx context.Context) error {
 	err = m.waitForSocket(3*time.Second, errCh)
 	if err != nil {
 		err = errors.Wrapf(err, "Firecracker did not create API socket %s", m.Cfg.SocketPath)
-		m.err = err
+		m.fatalErr = err
 		close(m.exitCh)
 
 		return err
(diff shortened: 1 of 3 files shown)


machine.go
```diff
@@ -459,17 +459,19 @@ func (m *Machine) startVMM(ctx context.Context) error {
...
 
-		if err := m.doCleanup(); err != nil {
-			m.logger.Errorf("failed to cleanup after VM exit: %v", err)
+		cleanupErr := m.doCleanup()
+		if cleanupErr != nil {
+			m.logger.Errorf("failed to cleanup after VM exit: %v", cleanupErr)
 		}
 
-		errCh <- err
+		errCh <- multierror.Append(waitErr, cleanupErr).ErrorOrNil()
```
xibz: This change loses the error that was returned during the VM process. See line 431

kzys: startVMM() returns the err and it is propagated to Start(). Do we want to receive the error from Wait()?

network_test.go
```diff
@@ -316,7 +316,10 @@ func TestNetworkMachineCNI(t *testing.T) {
 
 			require.NoError(t, m.StopVMM(), "failed to stop machine")
 			waitCtx, waitCancel := context.WithTimeout(ctx, 3*time.Second)
-			assert.NoError(t, m.Wait(waitCtx), "failed waiting for machine stop")
+
+			// Since Shutdown() needs a different kernel, we simply killing Firecracker from StopVMM()
```
kzys: s/we simply killing/it simply kills/

machine.go
sipsma: The only difference I can see is that previously when `err` gets sets on line 498 (`err = m.waitForSocket`), that value would get picked up and put on `errCh` on line 474. However I think that behavior was somewhat buggy anyways in that if the machine starts and hangs without creating the socket file but also without the vmm process actually exiting, the call to `cmd.Wait` on line 462 could just hang indefinitely and, even if `cmd.Wait` did get unblocked, the goroutine on line 505 was never started and thus the error would never get set to `m.err`.

Given all that I think I am good with the change in this PR as it simplifies the code a bit, but it may be a good idea to also change line 501 to set `m.err = errors.New(msg)` instead of just `err`, which partially addresses the above issue by keeping the invariant that if there is an error during start, it will be set to `m.err` before `m.exitCh` is closed.

kzys: Then maybe L444 as well? We can say that all errors in startVMM() must be accessible through exitCh

sipsma: Yeah good call, it should be set to `m.err` there too for consistency

xibz: I think we should rename the err field or document that the error is only from the Wait call, if we are to do this. As it sits err seems like any error that was returned

kzys: How about `startErr`? I agree that "err records any error executing the VMM" is too broad.

kzys: Ah, but we are reusing the field for "any errors from Wait()"

sipsma: I think a reasonable definition for `m.err` would be "any error that occurs during or after machine start". We can write the doc for `Wait()` as "returns any error that occurs during or after machine start once the vmm process has exited". That seems to be what the intention of it was, right?

kzys: I've renamed `m.err` to `m.fatalErr` on https://github.com/firecracker-microvm/firecracker-go-sdk/pull/153/commits/c32330d0013580fd07adc64145a7e4baeb28dfaf.

kzys: I think I have addressed all comments on this thread. PTAL!

PR #155, opened 2019-11-04
Moves NetNS to Config from JailerConfig 

xibz:
This change moves the NetNS field from the JailerConfig to the Config.
This makes the most sense since net namespaces are not subject to only
jailers, but can be used generally as well.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -108,6 +108,10 @@ type Config struct {
 	// set the CNI ContainerID and create a network namespace path if
 	// CNI configuration is provided as part of NetworkInterfaces
 	VMID string
+
+	// NetNS represents the path to a network namespace handle. If present, the
+	// application will use this to join the associated network namespace
+	NetNS string
 }
 
 // Validate will ensure that the required fields are set and that
@@ -297,6 +301,10 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) 
 	m.machineConfig = cfg.MachineCfg
 	m.Cfg = cfg
 
+	if cfg.NetNS == "" && cfg.NetworkInterfaces.cniInterface() != nil {
+		m.Cfg.NetNS = m.defaultNetNSPath()
+	}
+
 	m.logger.Debug("Called NewMachine()")
 	return m, nil
 }
@@ -354,24 +362,8 @@ func (m *Machine) Wait(ctx context.Context) error {
 	}
 }
 
-func (m *Machine) netNSPath() string {
-	// If the jailer specifies a netns, use that
-	if jailerNetNS := m.Cfg.JailerCfg.netNSPath(); jailerNetNS != "" {
-		return jailerNetNS
-	}
-
-	// If there isn't a jailer netns but there is a network
-	// interface with CNI configuration, use a default netns path
-	if m.Cfg.NetworkInterfaces.cniInterface() != nil {
-		return filepath.Join(defaultNetNSDir, m.Cfg.VMID)
-	}
-
-	// else, just don't use a netns for the VM
-	return ""
-}
-
 func (m *Machine) setupNetwork(ctx context.Context) error {
-	err, cleanupFuncs := m.Cfg.NetworkInterfaces.setupNetwork(ctx, m.Cfg.VMID, m.netNSPath(), m.logger)
+	err, cleanupFuncs := m.Cfg.NetworkInterfaces.setupNetwork(ctx, m.Cfg.VMID, m.Cfg.NetNS, m.logger)
 	m.cleanupFuncs = append(m.cleanupFuncs, cleanupFuncs...)
 	return err
 }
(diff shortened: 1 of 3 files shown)


machine.go
```diff
@@ -296,6 +291,12 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
 	m.machineConfig = cfg.MachineCfg
 	m.Cfg = cfg
 
+	m.Handlers.FcInit = m.Handlers.FcInit.Swap(NewSetupNetworkHandler(m.netNSPath()))
```
samuelkarp: This should probably only happen if `m.netNSPath() != ""`

machine.go
```diff
@@ -296,6 +291,12 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
...
+
+	for _, opt := range opts {
```
samuelkarp: Why'd you want to move the opts down here?

xibz: I moved them down because that's generally where they belong. They usually are towards the end of a function, but I found that we were adding things to the `NewMachine` function which could stomp over any option that was provided. So I moved it down for safety purposes to avoid stomping over user overrides.

samuelkarp: Can you do that in a separate commit?  It's unrelated to the subject of this commit (and PR).

sipsma:
I'm not sure this addresses the original issue in https://github.com/firecracker-microvm/firecracker-containerd/pull/305

Even if you override the handler to use a different netns path than what is returned by `m.netNSPath()`, if your machine has a CNI-configured interface then when you get to [these lines](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/6292bbe547ca12dcd4b00ed071b2e682888a32c9/machine.go#L432) `m.netNSPath()` will return [the default path](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/6292bbe547ca12dcd4b00ed071b2e682888a32c9/machine.go#L365) used when a path isn't specified via the go-sdk's `JailerConfig`.

So the only way this will actually work is if the handler gets overridden with the exact default value that is returned in `m.netNSPath()`. We could expose that default value in order to make it possible to do this, but honestly this just seems incredibly confusing. There's then 3 possible sources for the netns path can be set and 2 of them have some degree of correspondence required.

What's the downside of the other suggestion I mentioned in the other PR to just set the `NetNS` path of the machine's JailerCfg? That would leave the current logic for getting the netns path in the Go SDK the same and seems straightforward on the surface.

xibz:
> What's the downside of the other suggestion I mentioned in the other PR to just set the NetNS path of the machine's JailerCfg? That would leave the current logic for getting the netns path in the Go SDK the same and seems straightforward on the surface.

Because we use the `nil` and present value of jailer config to mean something. If we have a jailer config, that means we should enable jailing. So that means that logic would need to be rewritten and introduce a new flag to distinguish between running firecracker's jailer versus a custom one.

sipsma:
> Because we use the `nil` and present value of jailer config to mean something. If we have a jailer config, that means we should enable jailing. So that means that logic would need to be rewritten and introduce a new flag to distinguish between running firecracker's jailer versus a custom one.

Okay, I can see there being value in finding a way to unite the JailerCfg of the machine w/ firecracker-containerd's jailer instead of having firecracker-containerd mangle the handler setup of the machine entirely by itself, but I get that's kind of a monumental effort right now.

What I think matters for this change is that there is only one place for users of the SDK to externally specify the netns path (which keeps the complication for users to the possible minimum). What do you think about moving the `NetNS` field outside of `JailerCfg` into a field directly in the machine's `Cfg`? We can then internally pass the netns path to the jailer (as a private field) but also just use it elsewhere in the machine too. This is a breaking change, but I think it's pretty reasonable since the `NetNS` path is now a field that can apply to any machine, even one that is not jailed.

xibz:
@sipsma - I messaged you a little bit ago with that same suggestion :p

sipsma:
> @sipsma - I messaged you a little bit ago with that same suggestion :p

Sorry, I'm on plane wifi right now; github pages load after a minute or two but Chime hangs indefinitely at "Connecting..." :-) This is also what I suggested as the fallback in the original comment on the other PR, so are there any concerns with it? If not then sounds good to me!

machine.go
```diff
@@ -151,6 +155,7 @@ func (cfg *Config) Validate() error {
 	return nil
 }
 
+// ValidateNetwork .
```
kzys: Nit: please write the doc comment.

machine.go
```diff
@@ -420,19 +413,22 @@ func (m *Machine) attachDrives(ctx context.Context, drives ...models.Drive) erro
...
-	hasNetNS := m.netNSPath() != ""
-	jailerProvidedNetNS := m.Cfg.JailerCfg.netNSPath() != ""
+	isDefaultNetNSPath := m.Cfg.NetNS == m.defaultNetNSPath()
 	startCmd := m.cmd.Start
 
 	var err error
-	if hasNetNS && !jailerProvidedNetNS {
+	if isDefaultNetNSPath {
 		// If the VM needs to be started in a netns but no jailer netns was configured,
 		// start the vmm child process in the netns directly here.
```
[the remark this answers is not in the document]
xibz: Done!

machine.go
xibz: Removed. Shouldn't have added that :p

machine.go
xibz: Refactored. No longer uses handler to enter netns

machine.go
xibz: Just removed altogether. I'll go ahead and create a separate PR for this.

PR #157, opened 2019-11-05
Fixes race with io.Copy and O_NONBLOCK file

### LINKED ISSUE #156: Race condition when using LogWriter
After #145 was merged there now exists a race condition when setting up a LogWriter

In `captureFifoToFile`  this goroutine

```Go
	go func() {
		defer func() {
			if err := fifoPipe.Close(); err != nil {
				logger.Warnf("Failed to close fifo pipe: %v", err)
			}

			if err := syscall.Unlink(fifoPath); err != nil {
				logger.Warnf("Failed to unlink %s: %v", fifoPath, err)
			}
		}()

		if _, err := io.Copy(w, fifoPipe); err != nil {
			logger.Warnf("io.Copy failed to copy contents of fifo pipe: %v", err)
		}
	}()
```

will not block on `io.Copy`, I think this is due to opening with `syscall.O_NONBLOCK` unsure though. This causes the defer to trigger and `syscall.Unlink` to remove the originally configured `log.fifo`.  When using jailer if this happens before the `LinkFilesHandler` handler executes you can get a missing file error when trying to hardlink the `log.fifo` shown [here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/jailer.go#L418-L424).

If not using jailer I think that `io.Copy` could just exit and `BootstrapLoggingHandler` could have an error if it sent it's logging setup request after the `syscall.Unlink` occurs.

xibz:
This fix adds a retry around io.Copy to retry the fifo until it is
connected on the other end. Prior to this fix, the io.Copy would
immediately exit without error.

Fixes #156 

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -27,6 +27,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containerd/fifo"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/gofrs/uuid"
 	"github.com/hashicorp/go-multierror"
@@ -576,17 +577,12 @@ func (m *Machine) setupLogging(ctx context.Context) error {
 	return nil
 }
 
-func (m *Machine) captureFifoToFile(logger *log.Entry, fifoPath string, w io.Writer) error {
+func (m *Machine) captureFifoToFile(ctx context.Context, logger *log.Entry, fifoPath string, w io.Writer) error {
 	// open the fifo pipe which will be used
 	// to write its contents to a file.
-	fd, err := syscall.Open(fifoPath, syscall.O_RDONLY|syscall.O_NONBLOCK, 0600)
+	fifoPipe, err := fifo.OpenFifo(ctx, fifoPath, syscall.O_RDONLY|syscall.O_NONBLOCK, 0600)
 	if err != nil {
 		return fmt.Errorf("Failed to open fifo path at %q: %v", fifoPath, err)
-	}
-
-	fifoPipe := os.NewFile(uintptr(fd), fifoPath)
-	if fifoPipe == nil {
-		return fmt.Errorf("Invalid file descriptor")
 	}
 
 	logger.Debugf("Capturing %q to writer", fifoPath)
@@ -617,7 +613,7 @@ func (m *Machine) captureFifoToFile(logger *log.Entry, fifoPath string, w io.Wri
 		}()
 
 		if _, err := io.Copy(w, fifoPipe); err != nil {
-			logger.Warnf("io.Copy failed to copy contents of fifo pipe: %v", err)
+			logger.WithError(err).Warn("io.Copy failed to copy contents of fifo pipe")
 		}
 	}()
 
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/firecracker-microvm/firecracker-go-sdk
 go 1.11
 
 require (
+	github.com/containerd/fifo v0.0.0-20190816180239-bda0ff6ed73c
 	github.com/containernetworking/cni v0.7.2-0.20190807151350-8c6c47d1c7fc
 	github.com/containernetworking/plugins v0.8.2
 	github.com/go-openapi/errors v0.17.1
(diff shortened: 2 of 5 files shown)


machine.go
```diff
@@ -616,8 +616,17 @@ func (m *Machine) captureFifoToFile(logger *log.Entry, fifoPath string, w io.Wri
...
-			logger.Warnf("io.Copy failed to copy contents of fifo pipe: %v", err)
+		var err error
+		// since we are opening the fifo with O_NONBLOCK, any reads on a not
+		// connected fifo will return immediately with no error. To prevent nothing
+		// being read to the io.Writer, we will retry until an error has occurred
+		// which happens when the application has exited or the context has been
+		// closed.
+		for err == nil {
+			if _, err = io.Copy(w, fifoPipe); err != nil {
+				logger.Warnf("io.Copy failed to copy contents of fifo pipe: %v", err)
```
samuelkarp: If we're retrying past EOF and we expect an error, we're going to start getting a warning every time.  Maybe we don't want to log a warning when we're expecting an error?

xibz:
> I don't know if a retry loop is the right answer here. It feels like a bandaid over us not fully understanding how FIFOs work on Linux.

Instead of using a retry around `io.Copy` we can use `syscall.Select` to see if there is data on the fifo. That's when we `io.Copy`. That too would be a retry and is generally used when dealing with `O_NONBLOCK`. I'll go ahead and make the updates!

machine.go
xibz: @samuelkarp - I reworked this to use `syscall.Select`. This is a much better solution. Let me know what you think!

machine.go
```diff
@@ -616,8 +616,36 @@ func (m *Machine) captureFifoToFile(logger *log.Entry, fifoPath string, w io.Wri
...
+		// FD_SET taken from select.h of the linux kernel.
+		rfds.Bits[fd/strconv.IntSize] |= (1 << uint(fd%strconv.IntSize))
+
+		timeout := &syscall.Timeval{}
+		timeout.Usec = 10000000 // timeout in 10 microseconds
+		const maxRetries = 10
+		// Retry until there is data available on the file descriptor or that we've
+		// reached our retry limit.
+		for i := 0; i < maxRetries; i++ {
+			n, err := syscall.Select(fd+1, rfds, nil, nil, timeout)
```
sipsma: I'd prefer if we can find a different way to do this that doesn't involve `select`. The code written is already fairly complicated and it doesn't to my knowledge yet deal with issues mentioned in the `select(2)` docs like `EINTR` being returned if a signal interrupts the call, needing to reset the `Timeval` struct every time, or the mysterious phrase [`There may be other circumstances in which a file descriptor is spuriously reported as ready`](http://man7.org/linux/man-pages/man2/select.2.html#BUGS) (which implies that we could for unknown reasons still try to do `io.Copy` on a FIFO that hasn't been opened for writing yet).

In firecracker-containerd we just use the [`containerd/fifo`](https://github.com/containerd/fifo) package to open a FIFO and then later just call `io.Copy` on it without needing to deal with any sort of retries ourselves. Have you given that a try? It seems like there's potential it could simplify this code quite a bit.

xibz: I don’t know if that package will work by looking at the code. I am pretty sure we will run into the same issue. I can try it, but I have a feeling it will be for naught

sipsma: Try opening it w/ just `syscall.O_RDWR` using `fifo.OpenFifo` on line 582. My understanding of the implementation in the `fifo` package is that when you do that, you won't block on the open call waiting for the other side (which is what we want) but then once you actually do the call to `io.Copy`, [the reader will internally wait for the other side to be opened or the object returned by `OpenFifo` to be closed (i.e. via line 599).](https://github.com/containerd/fifo/blob/bda0ff6ed73c67bfb5e62bc9c697f146b7fd7f13/fifo.go#L152-L162)

xibz: @sipsma - Yea, it totally works! I was skeptical but didn't see that they implemented the `syscall.Conn` interface in raw which handles all the states for the fifo. Very nice implementation!

sipsma: Awesome! Yeah I'm very glad the containerd folks made that package...

Issue #164, opened 2019-12-11
Make hardcoded timeouts configurable


sipsma:
When attempting to reproduce some rare errors in firecracker-containerd, I've at times increased load quite a bit which has caused me to hit various hardcoded timeouts in the go sdk. When I increase the timeouts in my fork the operations succeed, so the timeouts are just set too low for the tests I'm running. The current values seem fine as defaults, but they should be configurable.

Ones I've encountered:
* https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/machine.go#L495
* https://github.com/firecracker-microvm/firecracker-go-sdk/blob/9e2ff62839b228665976be79191e33cffce5f760/firecracker.go#L29
* https://github.com/firecracker-microvm/firecracker-go-sdk/blob/9e2ff62839b228665976be79191e33cffce5f760/firecracker.go#L173

Zyqsempai:
@sipsma Totally make sense to me, but how do you plan to configure it?

nmeyerhans:
It's probably reasonable for this sort of thing to be settable via environment variables.

xibz:
I think I much prefer these to live in the config, rather than as environment variables. We can support both, but I think having them inside the config at the bare minimum is preferable. And honestly, [this](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/9e2ff62839b228665976be79191e33cffce5f760/firecracker.go#L173) can probably just use the request timeout.

nmeyerhans:
What config?

xibz:
@nmeyerhans - The config structure, https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/machine.go#L52

nmeyerhans:
Right. That's exactly why I think we should use environment variables. That structure comes from the client application, meaning that if we require these settings to come from there, then they can only be adjusted by a client application that knows about them. Since this is specifically useful for debugging, and affects the internal behavior of the SDK, there's no need for the client to know anything about them. Environment variables let us manipulate the SDK settings directly, independent of whatever client is in use. This makes SDK debugging much more practical.

Unless there are use-cases for adjusting these settings in production deployments, we should not. require clients to handle this.

xibz:
Hm, how do you feel about supporting both then? I mean that definitely adds a level of complexity of figuring out what value to use when both are set, but I can definitely see users wanting to adjust the timeouts in the config.

nmeyerhans:
[...]

PR #165, opened 2019-12-18
Added env's to be able to configure timouts

Zyqsempai:
Signed-off-by: Boris Popovschi <zyqsempai@mail.ru>

Issue #164 

*Description of changes:*
Added `FIRECRACKER_INIT_TIMEOUT_SECONDS` env to configure init timeouts and `FIRECRACKER_REQUEST_TIMEOUT_MILLISECONDS` for request timeouts.

### DIFF
--- a/firecracker.go
+++ b/firecracker.go
@@ -26,7 +26,12 @@ import (
 	ops "github.com/firecracker-microvm/firecracker-go-sdk/client/operations"
 )
 
-const firecrackerRequestTimeout = 500 * time.Millisecond
+const (
+	// env name to make firecracker request timeout configurable
+	firecrackerRequestTimeoutEnv = "FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS"
+
+	defaultFirecrackerRequestTimeout = 500
+)
 
 // newFirecrackerClient creates a FirecrackerClient
 func newFirecrackerClient(socketPath string, logger *logrus.Entry, debug bool) *client.Firecracker {
@@ -51,13 +56,18 @@ func WithOpsClient(opsClient ops.ClientIface) ClientOpt {
 
 // Client is a client for interacting with the Firecracker API
 type Client struct {
-	client *client.Firecracker
+	client                    *client.Firecracker
+	firecrackerRequestTimeout int
+	firecrackerInitTimeout    int
 }
 
 // NewClient creates a Client
 func NewClient(socketPath string, logger *logrus.Entry, debug bool, opts ...ClientOpt) *Client {
 	httpClient := newFirecrackerClient(socketPath, logger, debug)
 	c := &Client{client: httpClient}
+	c.firecrackerRequestTimeout = envValueOrDefaultInt(firecrackerRequestTimeoutEnv, defaultFirecrackerRequestTimeout)
+	c.firecrackerInitTimeout = envValueOrDefaultInt(firecrackerInitTimeoutEnv, defaultFirecrackerInitTimeoutSeconds)
+
 	for _, opt := range opts {
 		opt(c)
 	}
@@ -72,7 +82,7 @@ type PutLoggerOpt func(*ops.PutLoggerParams)
 // PutLogger is a wrapper for the swagger generated client to make calling of
 // the API easier.
 func (f *Client) PutLogger(ctx context.Context, logger *models.Logger, opts ...PutLoggerOpt) (*ops.PutLoggerNoContent, error) {
-	timeout, cancel := context.WithTimeout(ctx, firecrackerRequestTimeout)
+	timeout, cancel := context.WithTimeout(ctx, time.Duration(f.firecrackerRequestTimeout)*time.Millisecond)
 	defer cancel()
 
 	loggerParams := ops.NewPutLoggerParamsWithContext(timeout)
@@ -91,7 +101,7 @@ type PutMachineConfigurationOpt func(*ops.PutMachineConfigurationParams)
 // PutMachineConfiguration is a wrapper for the swagger generated client to
 // make calling of the API easier.
 func (f *Client) PutMachineConfiguration(ctx context.Context, cfg *models.MachineConfiguration, opts ...PutMachineConfigurationOpt) (*ops.PutMachineConfigurationNoContent, error) {
-	timeout, cancel := context.WithTimeout(ctx, firecrackerRequestTimeout)
+	timeout, cancel := context.WithTimeout(ctx, time.Duration(f.firecrackerRequestTimeout)*time.Millisecond)
 	defer cancel()
 
 	mc := ops.NewPutMachineConfigurationParamsWithContext(timeout)
@@ -110,7 +120,7 @@ type PutGuestBootSourceOpt func(*ops.PutGuestBootSourceParams)
 // PutGuestBootSource is a wrapper for the swagger generated client to make
 // calling of the API easier.
 func (f *Client) PutGuestBootSource(ctx context.Context, source *models.BootSource, opts ...PutGuestBootSourceOpt) (*ops.PutGuestBootSourceNoContent, error) {
-	timeout, cancel := context.WithTimeout(ctx, firecrackerRequestTimeout)
+	timeout, cancel := context.WithTimeout(ctx, time.Duration(f.firecrackerRequestTimeout)*time.Millisecond)
 	defer cancel()
 
 	bootSource := ops.NewPutGuestBootSourceParamsWithContext(timeout)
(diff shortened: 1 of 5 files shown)


Zyqsempai:
@nmeyerhans @xibz @sipsma PTAL

firecracker.go
```diff
@@ -26,7 +26,12 @@ import (
 	ops "github.com/firecracker-microvm/firecracker-go-sdk/client/operations"
 )
 
-const firecrackerRequestTimeout = 500 * time.Millisecond
+const (
+	// env name to make firecracker request timeout configurable
+	firecrackerRequestTimeoutEnv = "FIRECRACKER_REQUEST_TIMEOUT_MILLISECONDS"
```
samuelkarp: For environment variables that are part of the SDK and not an application built on top of the SDK, I think it would be better to be explicit about that in the naming of the variable.  Something like "FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS".

machine.go
```diff
@@ -42,6 +42,11 @@ const (
 
 	// as specified in http://man7.org/linux/man-pages/man8/ip-netns.8.html
 	defaultNetNSDir = "/var/run/netns"
+
+	// env name to make firecracker init timeout configurable
+	firecrackerInitTimeoutEnv = "FIRECRACKER_INIT_TIMEOUT_SECONDS"
```
samuelkarp: Same here, could be better as "FIRECRACKER_GO_SDK_INIT_TIMEOUT_SECONDS".

utils.go
```diff
@@ -27,3 +29,13 @@ func waitForAliveVMM(ctx context.Context, client *Client) error {
 		}
 	}
 }
+
+// envValueOrDefaultTime check if env value exists and returns it or returns default value
+// provided as a second param to this function
+func envValueOrDefaultTime(envName string, def int) time.Duration {
```
samuelkarp: I think I'd prefer to see this just return an `int` (and named `envValueOrDefaultInt`) and let the conversion to a `time.Duration` occur elsewhere where it can be scaled (by `time.Millisecond` or `time.Second` as is happening for the two new environment variables).  Letting this return a `time.Duration` doesn't really make it clear as to the scale of the duration.

xibz: I wonder if it makes sense to do this during `NewMachine` and set that value somewhere on the machine so we don't have to do this multiple times. That way defaults are all happening in the same place. What do you think?

Zyqsempai: Hmm, I can't say that I like that idea, to have a global state, especially few of them like in this case, it's not so expensive to make this call on each request.

Zyqsempai:
@samuelkarp Done.

Makefile
```diff
@@ -14,6 +14,8 @@
 # Set this to pass additional commandline flags to the go compiler, e.g. "make test EXTRAGOARGS=-v"
 EXTRAGOARGS:=
 DISABLE_ROOT_TESTS?=1
+FIRECRACKER_INIT_TIMEOUT_SECONDS?=
+FIRECRACKER_REQUEST_TIMEOUT_MILLISECONDS?=
```
nmeyerhans: We shouldn't need these settings here, since environment variables evaluate to the empty string by default.

utils.go
```diff
@@ -27,3 +29,13 @@ func waitForAliveVMM(ctx context.Context, client *Client) error {
 		}
 	}
 }
+
+// envValueOrDefaultInt check if env value exists and returns it or returns default value
+// provided as a second param to this function
+func envValueOrDefaultInt(envName string, def int) int {
```
nmeyerhans: I don't really love this function name, but I guess it's descriptive enough and I can't really think of anything better right now.

utils.go
xibz: I think it is less about it being more efficient, but a preference of having defaults happening in the same place, `NewClient`. That way in the future we can change the default in a single place. Currently if we rename the function, `envValueOrDefaultInt`, to something else, we'd have to change it in several places. So, I think I prefer that it was done in `NewClient` and have a private field that houses the values.

utils.go
xibz: Can we also add a test for this function?

utils.go
```diff
@@ -27,3 +29,13 @@ func waitForAliveVMM(ctx context.Context, client *Client) error {
...
+	envVal, err := strconv.Atoi(os.Getenv(envName))
+	if envVal == 0 || err != nil {
+		envVal = def
```
xibz: I don't know if I like the error being silently switch to the default here. I think we, at minimum, would want to log this back to the user.

.buildkite/pipeline.yml
```diff
@@ -47,6 +47,8 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
 
   - label: ':hammer: tests'
+    FIRECRACKER_INIT_TIMEOUT_SECONDS: 3
```
samuelkarp: Can you update the environment variable names here?

utils.go
samuelkarp: This function shouldn't produce logs, but if we do want to expose the error it can return `(int, error)`.

nmeyerhans: This is an internal API, so we can change it at will if we need to. We don't have a logger object in most of the contexts in which we call this function, so I'm not sure there's much value in returning the error.

Since this functionality is intended as a convenience for helping developers debug the SDK, it won't be used often and will almost certainly be used in conjunction with inspection of the code, consultation of the docs, etc. So I think we can simply document that the values provided by the environment will be parsed by Atoi() and that errors will be silently swallowed. No need to overthink this.

xibz: I don't think it is a matter of overthinking but a concern of someone fat fingering a value and it being ignored. I can see some frustration here if a user intended for a value to be used and it turns out we were ignoring it.

I would rather have the error be apparent than the user need to go through code or docs.

Zyqsempai:
Guys, PTAL

.buildkite/pipeline.yml
```diff
@@ -47,6 +47,8 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
 
   - label: ':hammer: tests'
+    FIRECRACKER_GO_SDK_INIT_TIMEOUT_SECONDS: 3
+    FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS: 500
```
nmeyerhans: I'd actually prefer that we not set these here, even if we're just explicitly setting the default value. Using the default value during our automated testing is the right thing to do, but explicitly setting the values introduces an opportunity for configuration drift.

.buildkite/pipeline.yml
```diff
@@ -57,6 +59,8 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
 
   - label: ':hammer: root tests'
+    FIRECRACKER_GO_SDK_INIT_TIMEOUT_SECONDS: 3
+    FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS: 500
```
nmeyerhans: Same as above.

.buildkite/pipeline.yml
Zyqsempai: Hmm, then how do you suggest to set it?

Zyqsempai: I mean how should regular users set this env's?

nmeyerhans: What you're doing is exactly right, and I think this is how users should set these variables when they do. I just don't think we should hardcode these settings here. When we run the tests in buildkite, we should rely on the defaults from the go code. If somebody needs to debug something and tweak these values, they can set environment variables, either here in the buildkite config (temporarily, during debugging) or in their runtime environment.

Zyqsempai: Then we definitely should point it in some kind of documentation or somewhere else, since anyone except us, don't know a thing about those env's.

firecracker.go
```diff
@@ -26,7 +26,16 @@ import (
...
-const firecrackerRequestTimeout = 500 * time.Millisecond
+const (
+	// env name to make firecracker request timeout configurable
+	firecrackerRequestTimeoutEnv = "FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS"
+
+	defaultFirecrackerRequestTimeout = 500
+)
+
+var (
+	firecrackerRequestTimeout, firecrackerInitTimeout int
```
samuelkarp: I'd much prefer to not have global/package-level variables when we can avoid it.  Can you move these into the `Client` struct?

.buildkite/pipeline.yml
nmeyerhans: Yup, maybe add something to [HACKING.md](/firecracker-microvm/firecracker-go-sdk/HACKING.md)?

Zyqsempai:
Take a final look guys;)

Issue #85, opened 2019-03-07
Rethink signal handling


samuelkarp:
Default signal handling in Linux could lead the `firecracker` process to become orphaned.  To prevent this, the SDK sets up some default signal handlers that forward signals along to the `firecracker` process.  However, this is pretty implicit and not super discoverable unless you read the code, and it may interfere with signal handling that the calling application intends to perform.  We should think about:

* what we want signal handling to be
* whether the SDK should register signal handlers by default
* how this can be made more explicit and less implicit

rstat1:
If I may add my 2 cents as a user of firecracker and this sdk (in some personal experiments), my biggest issue with the SDK's signal handling is that it doesn't unset the handlers after the firecracker process goes away. This ended up interfering with some of my own signal handlers.

evantorrie:
In my case, I want to control signal handling from my own process - I would rather have some option to disable the auto-signal handler creation inside the firecracker SDK.

kzys:
How about having 

```
func (m *Machine) Close() error {
	if err := m.StopVMM(); err != nil {
		return err
	}
	return m.Wait(nil)
}
```

and let customers kill `firecracker` by closing the machine?

It is a bit cumbersome than our current signal handling, but more explicit and won't interfere with the host application's signal handling.

We currently try to free resources we made by assuming the application's signal handling policy, but we really don't know about. In Go, providing explicit `Close()` is more common way of doing that.

jeromegn:
For what it's worth, we have a use-case where we do not want signals to be handled by firecracker-go-sdk since we sometimes want to orphan firecrackers and pick them back up later (we only really care about their PID and use meta data to recognize them).

It could be an option in the machine config.

kzys:
I think #166 is good enough to close this issue.

Regarding the idea I proposed above, StopVMM() is killing a VM forcibly and we are trying to avoid using the method on firecracker-containerd. So calling this method `Close()` is bit misleading, since it is unsafe.

samuelkarp:
@kzys I agree.  Closing this.

PR #166, opened 2019-12-19
Configurable signal handling, fix signal cleanup

michaeldwan:
*Issue #, if available:* #85 

*Description of changes:*
Adds a `DisableSignals` option to `machine.Config` for host apps that need to manage firecracker signals themselves. This also fixes an issue where signals weren't cleaned if the firecracker socket failed to open.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -894,7 +894,7 @@ func (m *Machine) setupSignals() {
 	}
 
 	m.logger.Debugf("Setting up signal handler: %v", signals)
-	sigchan := make(chan os.Signal)
+	sigchan := make(chan os.Signal, len(signals))
 	signal.Notify(sigchan, signals...)
 
 	go func() {
--- a/machine.go
+++ b/machine.go
@@ -898,11 +898,14 @@ func (m *Machine) setupSignals() {
 	signal.Notify(sigchan, signals...)
 
 	go func() {
-		select {
-		case sig := <-sigchan:
-			m.logger.Printf("Caught signal %s", sig)
-			m.cmd.Process.Signal(sig)
-		case <-m.exitCh:
+		for {
+			select {
+			case sig := <-sigchan:
+				m.logger.Printf("Caught signal %s", sig)
+				m.cmd.Process.Signal(sig)
+			case <-m.exitCh:
+				break
+			}
 		}
 
 		signal.Stop(sigchan)
--- a/machine.go
+++ b/machine.go
@@ -117,6 +117,10 @@ type Config struct {
 	// NetNS represents the path to a network namespace handle. If present, the
 	// application will use this to join the associated network namespace
 	NetNS string
+
+	// ForwardSignals is an optional list of signals to catch and forward to
+	// firecracker. If not provided, the default signals will be used.
+	ForwardSignals []os.Signal
 }
 
 // Validate will ensure that the required fields are set and that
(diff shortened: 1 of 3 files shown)


machine.go
```diff
@@ -880,3 +869,30 @@ func (m *Machine) waitForSocket(timeout time.Duration, exitchan chan error) erro
 		}
 	}
 }
+
+// Set up a signal handler and pass INT, QUIT, and TERM through to firecracker
+func (m *Machine) setupSignals() {
+	if m.Cfg.DisableSignals {
```
xibz: How about instead of a `DisableSignals` flag, we have a strategy that allows for setting signals up. This strategy could be something like

```go
type BootstrapSignals func() error

type Machine struct {
    // machine field here
    BootstrapSignals BootstrapSignals
}
```

Then the default would be the `setupSignals` method you already have, but you can overwrite it with whatever you want.

xibz:
Also, please sign your commits with `git commit -s`. That seems to be the only reason why the CI is failing

machine.go
michaeldwan: It seems host apps need a way to control the list of signals to forward rather than inject new behavior. A bootstrap function is heavy handed for apps that just need to change which signals are forwarded since Machine doesn't export the command, so everything needs to be re-implemented. And for apps that don't want forwarding, an empty list seems clearer than a noop bootstrap function. 

I think exposing an optional list of signals to handle (which defaults to the current list) would make a cleaner API than injecting a function. What do you think?

michaeldwan:
@xibz I rebased with signed commits and updated the API to something explicit. How does this look?

Forward default signals:
```go
firecracker.NewMachine(ctx, firecracker.Config{})
```

Forward provided signals:
```go
firecracker.NewMachine(ctx, firecracker.Config{
    ForwardSignals: []os.Signal{os.Interrupt},
})
```

Disable all forwarding:
```go
firecracker.NewMachine(ctx, firecracker.Config{
    ForwardSignals: []os.Signal{},
})
```

michaeldwan:
FYI, we’ve been running this PR in production for a few hours through our custom nomad driver and disabling signals is working great for our use case.

machine.go
```diff
@@ -880,3 +869,38 @@ func (m *Machine) waitForSocket(timeout time.Duration, exitchan chan error) erro
 		}
 	}
 }
+
+// Set up a signal handler to pass through to firecracker
+func (m *Machine) setupSignals() {
+	signals := m.Cfg.ForwardSignals
+
+	if signals == nil {
```
xibz: Can we move this to `NewMachine`? That's where all defaults are occurring and I'd like to keep them in the same area.

michaeldwan: Good call, fixed.

machine.go
```diff
@@ -880,3 +879,28 @@ func (m *Machine) waitForSocket(timeout time.Duration, exitchan chan error) erro
...
+// Set up a signal handler to pass through to firecracker
+func (m *Machine) setupSignals() {
+	signals := m.Cfg.ForwardSignals
+
+	if len(signals) == 0 {
+		return
+	}
+
+	m.logger.Debugf("Setting up signal handler: %v", signals)
+	sigchan := make(chan os.Signal)
```
samuelkarp: Can you make `sigchan` have length greater than or equal to `len(signals)`?  From [the `signal.Notify` documentation](https://golang.org/pkg/os/signal/#Notify):

> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

michaeldwan:
The signal channel is now buffered. I'm going to try and get a test for this. I can't get the tests to run correctly on my mac so I'll be debugging on BuildKite... ignore failure spam for a bit.

michaeldwan:
I've made the requested changes and have a test ensuring signals are forwarded correctly. Root tests are failing before they start, but I don't think it's related to this change. Let me know how it looks.

samuelkarp:
@michaeldwan Thank you for being so persistent here!  You are correct that the root test issue is not related to your change.  I opened https://github.com/firecracker-microvm/firecracker-go-sdk/issues/168 to track the root cause and https://github.com/firecracker-microvm/firecracker-go-sdk/pull/169 should address it for now.  Can you rebase on top of `master` to pick up that change?  That should fix the issue you had and allow the root tests to run properly for your PR.

samuelkarp:
The test looks good to me too.  Once we have this rebased (or, if you don't have the time, I can do this for you) we'll be good to go.

michaeldwan:
@samuelkarp I rebased and tests are passing, thanks for the fix

xibz:
@michaeldwan - Thank you for the contribution and persistence. These changes look great. Going to go ahead and merge.

PR #167, opened 2019-12-27
Update Swagger API spec to include new option, InitrdPath

smira:
Signed-off-by: Andrey Smirnov <smirnov.andrey@gmail.com>

*Issue #, if available:*

*Description of changes:*

`master` version of `firecracker` adds support for `initrd_path`, this PR adds support to Go SDK

### DIFF
--- a/jailer.go
+++ b/jailer.go
@@ -381,6 +381,19 @@ func LinkFilesHandler(rootfs, kernelImageFileName string) Handler {
 				return err
 			}
 
+			initrdFilename := ""
+			if m.Cfg.InitrdPath != "" {
+				initrdFilename := filepath.Base(m.Cfg.InitrdPath)
+				// copy initrd to root fs
+				if err := linkFileToRootFS(
+					m.Cfg.JailerCfg,
+					filepath.Join(rootfs, initrdFilename),
+					m.Cfg.InitrdPath,
+				); err != nil {
+					return err
+				}
+			}
+
 			// copy all drives to the root fs
 			for i, drive := range m.Cfg.Drives {
 				hostPath := StringValue(drive.PathOnHost)
@@ -398,6 +411,9 @@ func LinkFilesHandler(rootfs, kernelImageFileName string) Handler {
 			}
 
 			m.Cfg.KernelImagePath = kernelImageFileName
+			if m.Cfg.InitrdPath != "" {
+				m.Cfg.InitrdPath = initrdFilename
+			}
 
 			for _, fifoPath := range []*string{&m.Cfg.LogFifo, &m.Cfg.MetricsFifo} {
 				if fifoPath == nil || *fifoPath == "" {
--- a/client/models/boot_source.go
+++ b/client/models/boot_source.go
@@ -33,6 +33,9 @@ type BootSource struct {
 	// Kernel boot arguments
 	BootArgs string `json:"boot_args,omitempty"`
 
+	// Host level path to the initrd image used to boot the guest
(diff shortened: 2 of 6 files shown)


jailer.go
```diff
@@ -364,7 +364,7 @@ func linkFileToRootFS(cfg *JailerConfig, dst, src string) error {
 
 // LinkFilesHandler creates a new link files handler that will link files to
 // the rootfs
-func LinkFilesHandler(rootfs, kernelImageFileName string) Handler {
+func LinkFilesHandler(rootfs, kernelImageFileName, initrdFilename string) Handler {
```
xibz: Do we really need the `initrdFilename` parameter?

We could instead just use `filepath.Base` to get the filename from the path.

jailer.go
```diff
@@ -431,13 +445,15 @@ func LinkFilesHandler(rootfs, kernelImageFileName string) Handler {
 type NaiveChrootStrategy struct {
 	Rootfs          string
 	KernelImagePath string
+	InitrdPath      string
 }
 
 // NewNaiveChrootStrategy returns a new NaivceChrootStrategy
-func NewNaiveChrootStrategy(rootfs, kernelImagePath string) NaiveChrootStrategy {
+func NewNaiveChrootStrategy(rootfs, kernelImagePath, initrdPath string) NaiveChrootStrategy {
```
xibz: since this is an optional field, I would much prefer functional options here.

```go
type NaiveChrootStrategyOption func(*NaiveChrootStrategy)

func NewNaiveChrootStrategy(rootfs, kernelImagePath, options ...NaiveChrootStrategyOption) NaiveChrootStrategy
```

kzys:
Note that this change will be in 0.20.1 release since Firecracker 0.20.0 doesn't have this option yet.

jailer.go
smira: removed that parameter, and reverted `NewNaiveChrootStrategy` changes below

smira:
Rebased without changes to resolve conflicts, @xibz does this version look better?

kzys:
@xibz Can you take a look?

smira:
> @smira - These changes look good, but I'd like to see at least a test to ensure `initrd` is actually being called. I think a simple way to test this is to have the init script add an entry via [mmds](https://github.com/firecracker-microvm/firecracker/blob/master/docs/mmds.md) and verify it on the SDK side by checking for a specific key. Please let us know if any questions with the test!

@xibz it sounds like a good but complicated test. Do you have any examples for similar flows or tests?

I feel that it tests more than just Firecracker Go SDK, but rather firecracker itself, as I think SDK work finishes when `initrd_path` is passed via API call to firecracker, and loading & executing `initrd` is beyond SDK control.

xibz:
@smira - Yea, this is just to ensure that the SDK is properly passing the initrd path via the handler. Unfortunately, we do not have any similar tests with that type of workflow. I would be okay with adding the test myself since it is a rather complicated thing to setup, as long as others, @kzys and @mxpv, are okay with that.

kzys:
I think having some tests around the functionality is good to make sure that Firecracker is not having any regressions.

That being said, making initrd from our tests is hard. Other binary blobs - Linux kernel (vmlinux) and rootfs are coming from Firecracker team's S3 bucket. Running mkinitrd or whatever is technically possible, but we don't have that on our EC2 host yet.

So I think merging this change is fine, since it is required to support the new feature.

xibz:
@smira - Thank you again for taking the time to submit this PR. Much appreciated!

Issue #175, opened 2020-01-21
jailer: flag seccomp-level is not supported anymore 


muhamadazmy:
The jailer on the master branch has dropped the `--seccomp-level` flag [here](https://github.com/firecracker-microvm/firecracker/compare/v0.20.0...master#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR19).

The sdk still append this flag to the `jailer` binary [here](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/jailer.go#L158)

Which causes the jailer to fail with error:
```
Arguments parsing error: Found argument 'seccomp-level' which wasn't expected, or isn't valid in this context.
```

xibz:
Hello @muhamadazmy, thank you for reaching out to us and notifying us of the issue. I've went ahead and added a PR with the fix, #176, but we will need to update our CI to build against master. If you have any questions about the PR and other issue, please let us know

muhamadazmy:
@xibz thank you so much for such quick response :)

xibz:
Hey @muhamadazmy - Sorry for the delay there. We managed to get the change merged in and verified that it does work with HEAD of Firecracker. We know have a better story on how we want to deal with supporting HEAD. If you have any other issues or questions, please don't hesitate to ask.

muhamadazmy:
Thank you so much @xibz . I really appreciate it 😊

PR #176, opened 2020-01-21
Remove SeccompLevel from Jailer

xibz:
Firecracker has since removed specifying the seccomp level in the jailer
and now is specified in Firecracker instead. This change removes the
seccomp level from the jailer and adds it to machine instead.

TODO:

- [x] Update CI system to build against master as well

#175 

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -101,6 +101,9 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    # TODO: Remove this once v0.21.0 has been released
+    soft_fail: # we softfail here since v0.20.0 jailer tests will be broken.
+      - exit_status: "*"
 
   - label: ':hammer: test against firecracker master'
     env:
@@ -120,8 +123,7 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
-    soft_fail:
-      - exit_status: "*"
+    # TODO: move soft_fail here once v0.21.0 of firecracker has been released
 
   - label: 'go mod tidy'
     commands:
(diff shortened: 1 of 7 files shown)


xibz:
managed to fix the seccomp issue and add the updated CI test to verify. However, Firecracker is panicking for some odd reason. Here is the issue, https://github.com/firecracker-microvm/firecracker/issues/1557

xibz:
root tests will fail due to Firecracker moving where the api socket location is.

IRCody:
> root tests will fail due to Firecracker moving where the api socket location is.

Is there a reason we can't look in the new location?

xibz:
@IRCody -We are using the new location. Firecracker has changed the default location, which is a breaking change. Which is why it succeeds against master but fails the root tests, since it is looking for the new location

IRCody:
So those tests will pass if we update the firecracker in CI?

.buildkite/pipeline.yml
```diff
@@ -101,6 +101,9 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    # TODO: Remove this once v0.21.0 has been released
```
kzys: Can we have build flags or something to remove `soft_fail`? We are not supporting multiple Firecracker versions, but keeping soft_fail will mask new incompatibilities.

Alternatively we could skip some tests.

xibz: > keeping soft_fail will mask new incompatibilities

Build flags would also do this?

xibz:
> So those tests will pass if we update the firecracker in CI?

Once Firecracker releases v0.21.0, then yes. However, they haven't done so, but customers are using HEAD of Firecracker and we shouldn't limit ourselves to the most released version, due to this.

kzys:
Here is what we used to have. Probably not really intentional, but due to the fact we didn't have Firecracker master on our build host.

|            | Firecraker v0.X.0 | Firecracker master |
|------------|-------------------|--------------------|
| SDK v0.X.0 | Must Work         | We Don't Know      |
| SDK master | Must Work         | We Don't know      |

And here is what we are going to have.

|            | Firecraker v0.X.0 | Firecracker master |
|------------|-------------------|--------------------|
| SDK v0.X.0 | Must Work         | We Don't Know      |
| SDK master | May Work          | Must Work          |

I agree that making master compatible with Firecracker's latest release is nice to have, but may require some efforts and we don't have customers who specifically ask about.

So I think merging this change is fine. We can always create a branch and cherry-pick if we want to have some fixes for Firecracker 0.20.0 customers.

PR #179, opened 2020-01-24
Add test step to test against HEAD of Firecracker

xibz:
This change adds a test step in BuildKite to test against HEAD of
Firecracker. The test will utilize Docker to run against Firecracker's
master branch.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -32,6 +32,17 @@ steps:
     commands:
       - 'sudo ip tuntap add fc-test-tap${BUILDKITE_BUILD_NUMBER} mode tap user $(sudo id -u buildkite-agent)'
       - 'sudo ip tuntap add fc-root-tap${BUILDKITE_BUILD_NUMBER} mode tap user $(sudo id -u buildkite-agent)'
+      - 'sudo ip tuntap add fc-mst-tap${BUILDKITE_BUILD_NUMBER} mode tap user $(sudo id -u buildkite-agent)'
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+
+  - label: ':linux: build docker images'
+    commands:
+      - 'make test-images'
+      - 'buildkite-agent artifact upload testdata/firecracker-master'
+      - 'buildkite-agent artifact upload testdata/jailer-master'
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
@@ -91,6 +102,27 @@ steps:
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
 
+  - label: ':hammer: test against firecracker master'
+    env:
+      FC_TEST_BIN: "testdata/firecracker-master"
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+      FC_TEST_JAILER_BIN: "testdata/jailer-master"
+    commands:
+      - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
+      - 'ln -s /var/lib/fc-ci/rootfs.ext4 testdata/root-drive.img'
+      - 'make -C cni install CNI_BIN_ROOT=$(pwd)/testdata/bin'
+      - 'buildkite-agent artifact download testdata/firecracker-master .'
+      - 'chmod +x testdata/firecracker-master'
+      - 'buildkite-agent artifact download testdata/jailer-master .'
+      - 'chmod +x testdata/jailer-master'
+      - "sudo -E FC_TEST_TAP=fc-mst-tap${BUILDKITE_BUILD_NUMBER} make test EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    soft_fail:
+      - exit_status: "*"
+
 
   # This allows the cleanup step to always run, regardless of test failure
   - wait: ~
@@ -100,6 +132,7 @@ steps:
     commands:
       - 'sudo ip tuntap del fc-test-tap${BUILDKITE_BUILD_NUMBER} mode tap'
       - 'sudo ip tuntap del fc-root-tap${BUILDKITE_BUILD_NUMBER} mode tap'
+      - 'sudo ip tuntap del fc-mst-tap${BUILDKITE_BUILD_NUMBER} mode tap'
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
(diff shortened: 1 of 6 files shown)


.buildkite/pipeline.yml
```diff
@@ -91,6 +99,21 @@ steps:
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
 
+  - label: ':hammer: test against firecracker master'
+    commands:
+      - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
+      - 'ln -s /var/lib/fc-ci/rootfs.ext4 testdata/root-drive.img'
+      - 'cp /usr/local/bin/firecracker-v0.20.0 testdata/firecracker'
+      - 'cp /usr/local/bin/jailer-v0.20.0 testdata/jailer'
```
kzys: It is confusing to see cp 0.20.0 under master test. Do we need them?

.buildkite/pipeline.yml
```diff
@@ -91,6 +99,21 @@ steps:
...
+      - 'make -C cni install CNI_BIN_ROOT=$(pwd)/testdata/bin'
+      - "sudo FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} make test-in-docker EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
```
kzys: Building Firecracker on Docker makes sense, but testing Go SDK on Docker doesn't (for me).

It would be better to either 1) use Docker for 0.20.0 tests as well or 2) don't use Docker for running tests. We are introducing unnecessary differences between 0.20.0 tests and master tests.

xibz: The reason why this was done is we'd need to swap out the binaries for the tests, one for master and another for the versioned. I'd rather avoid all of that and have it run in Docker.

.buildkite/pipeline.yml
xibz: Fixed

.buildkite/pipeline.yml
kzys: How about passing FC_TEST_BIN?

https://github.com/firecracker-microvm/firecracker-go-sdk/blob/57d08e35c3f176d2b9e8c43c220a3a6548c611ee/HACKING.md#L17

xibz: Good call! Totally forgot about that. Will update.

.buildkite/pipeline.yml
```diff
@@ -91,6 +99,22 @@ steps:
...
+      - 'ln -s /var/lib/fc-ci/vmlinux.bin testdata/vmlinux'
+      - 'ln -s /var/lib/fc-ci/rootfs.ext4 testdata/root-drive.img'
+      - 'make -C cni install CNI_BIN_ROOT=$(pwd)/testdata/bin'
+      - "sudo FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} make test EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    soft_fail:
+      - exit_status: 2
```
kzys: Why?

Makefile
```diff
@@ -12,8 +12,16 @@
...
+CARGO_CACHE_VOLUME_NAME?=cargocache
 DISABLE_ROOT_TESTS?=1
+DOCKER_IMAGE_TAG=latest
+EXTRAGOARGS:=
+FIRECRACKER_BUILDER_NAME=firecracker-builder
+GOLANG_IMAGE?=golang:1.13-stretch
+FIRECRACKER_BIN=testdata/firecracker-master
+JAILER_BIN=testdata/jailer-master
+FIRECRACKER_TARGET?=x86_64-unknown-linux-musl
+export GO_CACHE_VOLUME_NAME?=gocache
```
kzys: I think we don't need Go-related stuff inside Docker

Makefile
```diff
@@ -51,4 +60,33 @@ testdata/firecracker:
 testdata/root-drive.img:
 	$(curl) -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/fsfiles/hello-rootfs.ext4
 
+tools/firecracker-builder-stamp: tools/docker/Dockerfile
+	docker build \
+		-t localhost/$(FIRECRACKER_BUILDER_NAME):$(DOCKER_IMAGE_TAG) \
```
kzys: This `DOCKER_IMAGE_TAG` variable must be referenced and used from BuildKite to run multiple PR builds correctly.

tools/docker/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
kzys: firecracker-containerd needs entrypoint.sh to run firecracker-containerd as a background process and have "go test" as a foreground process. But SDK doesn't have to follow this pattern.

xibz: The reason why I did this is the `cp` was failing when I was doing it all on one command. So I decided to throw it in an entrypoint script and everything ended up working there.

.buildkite/pipeline.yml
xibz: We don't want failures to block PRs if master is broken. It should just let us know that something is broken in master.

Makefile
xibz: I'm not following. How does this prevent multiple builds?

kzys: Docker image `localhost/firecracker-builder:latest` is shared by multiple builds and not immutable. We should have something like https://github.com/firecracker-microvm/firecracker-containerd/blob/master/.buildkite/pipeline.yml#L22

tools/docker/entrypoint.sh
kzys: What was the error you got?

xibz: Basically, it couldn't stat the file, firecracker and jailer, even though there were there o.o.

Makefile
xibz: Done

tools/docker/entrypoint.sh
kzys: That's odd. I'm fine merging this PR but may try to remove the entrypoint.sh later.

.buildkite/pipeline.yml
kzys: I see. Thanks.

Makefile
```diff
@@ -12,8 +12,15 @@
...
-EXTRAGOARGS:=
+CARGO_CACHE_VOLUME_NAME?=cargocache
 DISABLE_ROOT_TESTS?=1
+DOCKER_IMAGE_TAG?=latest
+EXTRAGOARGS:=
+FIRECRACKER_BUILDER_NAME=firecracker-builder
+GOLANG_IMAGE?=golang:1.13-stretch
+FIRECRACKER_BIN=testdata/firecracker-master
+JAILER_BIN=testdata/jailer-master
+FIRECRACKER_TARGET?=x86_64-unknown-linux-musl
```
kzys: Please remove EXTRAGOARGS and GOLANG_IMAGE. We use Docker only for Firecracker.

machine_test.go
```diff
@@ -231,25 +231,8 @@ func TestJailerMicroVMExecution(t *testing.T) {
...
-		WithBin(jailerBin).
-		WithGID(IntValue(cfg.JailerCfg.GID)).
-		WithUID(IntValue(cfg.JailerCfg.UID)).
-		WithNumaNode(IntValue(cfg.JailerCfg.NumaNode)).
-		WithID(cfg.JailerCfg.ID).
-		WithChrootBaseDir(cfg.JailerCfg.ChrootBaseDir).
-		WithExecFile(cfg.JailerCfg.ExecFile).
-		WithStdout(os.Stdout).
-		WithStderr(os.Stderr).
-		Build(ctx)
```
kzys: Why do you want to remove these lines?

Makefile
xibz: `EXTRAGOARGS` are needed as we use those to pass go arguments to go commands in the make file.

However, `GOLANG_IMAGE` has been deleted.

machine_test.go
xibz: This was due to merging with #180 which adds the jailer path in the configuration. So it is no longer needed elsewhere, ie the builder.

PR #182, opened 2020-01-31
Make sure Machine#Wait() gurantee that the underlying Firecracker process is stopped

kzys:
The method should guarantee that the underlying process is dead,
but it didn't.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -516,14 +516,19 @@ func (m *Machine) startVMM(ctx context.Context) error {
 
 		return err
 	}
-	go func() {
-		select {
-		case <-ctx.Done():
-			m.fatalErr = ctx.Err()
-		case err := <-errCh:
-			m.fatalErr = err
-		}
 
+	// This goroutine is used to kill the process by context cancelletion,
+	// but doesn't tell anyone about that.
+	go func() {
+		<-ctx.Done()
+		m.stopVMM()
+	}()
+
+	// This goroutine is used to tell clients that the process is stopped
+	// (gracefully or not).
+	go func() {
+		m.fatalErr = <-errCh
+		m.logger.Debugf("closing the exitCh %v", m.fatalErr)
 		close(m.exitCh)
 	}()
 
--- a/machine.go
+++ b/machine.go
@@ -338,7 +338,10 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) 
 	return m, nil
 }
 
-// Start will iterate through the handler list and call each handler. If an
+// Start actually start a Firecracker microVM.
+// The context must not be cancelled while the microVM is running.
+//
+// It will iterate through the handler list and call each handler. If an
 // error occurred during handler execution, that error will be returned. If the
 // handlers succeed, then this will start the VMM instance.
 // Start may only be called once per Machine.  Subsequent calls will return
@@ -521,7 +524,10 @@ func (m *Machine) startVMM(ctx context.Context) error {
 	// but doesn't tell anyone about that.
 	go func() {
 		<-ctx.Done()
-		m.stopVMM()
+		err := m.stopVMM()
+		if err != nil {
+			m.logger.WithError(err).Errorf("failed to stop vm %q", m.Cfg.VMID)
+		}
 	}()
 
 	// This goroutine is used to tell clients that the process is stopped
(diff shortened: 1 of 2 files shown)


machine_test.go
```diff
@@ -1081,37 +1081,92 @@ func TestCaptureFifoToFile_leak(t *testing.T) {
...
+			alive, err := isProcessAlive(pid)
+			require.False(t, alive, "pid=%d is still there", pid)
+		})
+	}
+}
+
+func isProcessAlive(pid int) (bool, error) {
+	// Using kill(2) with signal=0 to check the existence of the process,
+	// because os.FindProcess always returns a process, regardless of whether the process is
+	// alive or not.
```
mxpv: Can you elaborate on this? 
As far as I see go's FindProcess does the same:
https://golang.org/src/os/exec_unix.go#L88
https://golang.org/src/os/exec_unix.go#L73
and its expected to receive err `os: process already finished`

mxpv: Clarification: you can use it like `err := os.FindProcess(x).Signal(0)`

machine.go
```diff
@@ -505,14 +505,19 @@ func (m *Machine) startVMM(ctx context.Context) error {
...
+	// but doesn't tell anyone about that.
 	go func() {
-		select {
-		case <-ctx.Done():
-			m.fatalErr = ctx.Err()
-		case err := <-errCh:
-			m.fatalErr = err
-		}
+		<-ctx.Done()
+		m.stopVMM()
```
xibz: Should we be ignoring this error?

machine_test.go
kzys: The error from Process#Signal() is not typed. The only way to check the type of the error is string comparison. So I'm unsure that is reliable over time.

https://github.com/golang/go/blob/753d56d3642eb83848aa39e65982a9fc77e722d7/src/os/exec_unix.go#L53

machine.go
kzys: The error from stopVMM(). Yeah, I think we should not ignore the error. Let me fix.

kzys: The last commit logs the error. Passing that to errCh seems harder than I thought, since the channel will be closed after the underlying process is finished.

machine_test.go
xibz: Don't think you need to cast the error
https://github.com/golang/go/blob/753d56d3642eb83848aa39e65982a9fc77e722d7/src/os/exec_unix.go#L72-L74

kzys: kill(2) returns EINVAL, EPERM or ESRCH as an error, but Go's Signal() doesn't distinguish them as different errors.

Would it be better to do something like below? Personally I don't want to depend on the string representation because it might change.

```
if err.Error() == "os: process already finished" {
  ...
}
```

Thoughts?

xibz: According to `Signal` it will only return `os: process already finished` error when `syscall.ESRCH` occurs. So, to me I think it doesn't matter either way. I think the string comparison is more ugly than checking against `ESRCH`, but I am fine with either solution.

kzys: > Apparently it sets done based on Wait, which may return the already finished error, which complicates the checks.

And all of these are happening inside os.Process. syscall.Kill is much simpler I think.

mxpv: I'd use os.Process, but I agree that strings comparison is ugly. Unfortunately sometimes its hard to get away from it (for instance there is a similar checker in containerd: https://github.com/containerd/containerd/blob/b0821c801dc2225bb7478f91e967888a353fb60a/pkg/process/utils.go#L131)

kzys: Okay. The last revision uses os.Process instead.

machine_test.go
```diff
@@ -1081,37 +1081,91 @@ func TestCaptureFifoToFile_leak(t *testing.T) {
...
+			go func() {
+				defer wg.Done()
+				c.stop(m, vmCancel)
+			}()
+
+			err = m.Wait(ctx)
+			require.Error(t, err, "Firecracker was killed and it must be reported")
+			t.Logf("err = %v", err)
+
+			proc, _ := os.FindProcess(pid) // os.FindProcess() won't fail on Unix systems
```
xibz: I understand that for unix systems this is a noop, but I'd still like the error check even if it is to never happen.

Other than that, these changes look good.

kzys: Thanks. Will do.

PR #191, opened 2020-02-06
Don't count pings while a guest OS is booting 

kzys:
*Issue #, if available:*

NA, but master build is failing due to this test.

*Description of changes:*

This test is sending ICMP pings to microVMs.

However since Machine#Start() only *starts* a microVM and not aware
about the status of the guest OS, some packets are sent while the OS is
booting.

### DIFF
--- a/network_test.go
+++ b/network_test.go
@@ -453,12 +453,23 @@ func startCNIMachine(t *testing.T, ctx context.Context, m *Machine) string {
 }
 
 func testPing(t *testing.T, ip string, count int, timeout time.Duration) {
+	// First, send one ping to make sure the machine is up
 	pinger, err := ping.NewPinger(ip)
 	require.NoError(t, err, "failed to create pinger")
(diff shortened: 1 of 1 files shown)


network_test.go
```diff
@@ -453,12 +453,22 @@ func startCNIMachine(t *testing.T, ctx context.Context, m *Machine) string {
 }
 
 func testPing(t *testing.T, ip string, count int, timeout time.Duration) {
+	// First, send one ping to make sure the machine is up
 	pinger, err := ping.NewPinger(ip)
 	require.NoError(t, err, "failed to create pinger")
+	pinger.SetPrivileged(true)
+
+	pinger.Count = 1
```
IRCody: Would we want to set a sort of timeout on this pinger also? It looks like the default is pretty high.

kzys: Sounds good. Will do.

kzys: Fixed.

Issue #200, opened 2020-02-19
machine.Start() context cancellation ignores clean exit


thepwagner:
Since https://github.com/firecracker-microvm/firecracker-go-sdk/pull/182 , I'm seeing a red herring message logged at "level=error" by the SDK:

<pre>
time="2020-02-19T13:06:51Z" level=info msg="VMM logging and metrics disabled."
time="2020-02-19T13:06:51Z" level=info msg="refreshMachineConfiguration: [GET /machine-config][200] getMachineConfigurationOK  &{CPUTemplate:T2 HtEnabled:0xc00f367494 MemSizeMib:0xc00f367488 VcpuCount:0xc00f367480}"
[... boring boot logs]
time="2020-02-19T13:06:51Z" level=info msg="startInstance successful: [PUT /actions][204] createSyncActionNoContent " 
[... machine completes, signals restart to terminate VM]
time="2020-02-19T13:07:03Z" level=info msg="firecracker exited: status=0"
<b>time="2020-02-19T13:07:03Z" level=error msg="failed to stop vm \"35bd7d10-99f1-4f47-99ff-903efeee8d03\"" error="os: process already finished"</b>
</pre>

I'm nesting all firecracker operations behind a `ctx, cancel := context.WithTimeout(...); defer cancel()`.
That deferred cancellation happens _after_ the VM has completed.

I'd like to patch this by string matching against the returned [err from `Signal()` inside `stopVMM()`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/machine.go#L581), like [containerd](https://github.com/containerd/containerd/blob/b0821c801dc2225bb7478f91e967888a353fb60a/pkg/process/utils.go#L131-L142) does.

kzys:
The SDK is writing `firecracker exited: status=0`, meaning that the SDK itself is aware about that the process is no longer aliver. Can't we use the signal to skip sending `Signal()` or hiding the error message?

thepwagner:
@kzys I started down that path in #201 , but the method of firing the signal and recovering had less races to deal with.

PR #201, opened 2020-02-19
stopVMM: don't signal if stopped

thepwagner:
*Issue #, if available:* https://github.com/firecracker-microvm/firecracker-go-sdk/issues/200

*Description of changes:*

1. Turn `TestWait/StopVMM_+_Context_Cancel` into a failing test case: after the VM is terminated (via exit(0) or signal) and context is cancelled, an error is logged.
1. Patch `machine.stopVMM()` to not send `SIGTERM` if we've already seen the process has exited.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -23,6 +23,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


machine.go
```diff
@@ -578,7 +579,11 @@ func (m *Machine) StopVMM() error {
 func (m *Machine) stopVMM() error {
 	if m.cmd != nil && m.cmd.Process != nil {
```
xibz: Out of curiosity, could we use [`ProcessState`](https://golang.org/pkg/os/#ProcessState) to check whether or not the process has exited?

kzys: I think, the process might be finished between checking (`m.cmd.ProcessState.Exited()`) and signaling (`m.cmd.Process.Signal(syscall.SIGTERM)`).

xibz: Ah, true there is a race there. Bleh, was trying to avoid the error string comparison :(.

xibz:
@thepwagner - Ive went ahead and merged the changes! Thank you for taking the time to do this.

Issue #217, opened 2020-03-13
VMM unable to create socket since 0.21


jeromegn:
I keep getting the following messages when trying to start a VMM:

```
DEBU[0001] Called NewMachine()
DEBU[0001] Called Machine.Start()
DEBU[0001] Marking Machine as Started
DEBU[0001] Running handler validate.NetworkCfg
DEBU[0001] Running handler validate.JailerCfg
DEBU[0001] Running handler fcinit.SetupNetwork
DEBU[0001] Running handler fcinit.SetupKernelArgs
DEBU[0001] Running handler fcinit.StartVMM
INFO[0001] Called startVMM(), setting up a VMM on /home/jerome/src/github.com/superfly/nomad-firecracker/firecrackers/62fab4aa/firecracker-0.21.1/62fab4aa/root/run/firecracker.socket
DEBU[0001] VMM started socket path is /home/jerome/src/github.com/superfly/nomad-firecracker/firecrackers/62fab4aa/firecracker-0.21.1/62fab4aa/root/run/firecracker.socket

DEBU[0004] GET /machine-config HTTP/1.1
Host: localhost
User-Agent: Go-http-client/1.1
Accept: application/json
Accept-Encoding: gzip

<snip>

WARN[0004] Failed handler "fcinit.StartVMM": Firecracker did not create API socket /home/jerome/src/github.com/superfly/nomad-firecracker/firecrackers/62fab4aa/firecracker-0.21.1/62fab4aa/root/run/firecracker.socket: context deadline exceeded
```

I've upgraded my firecracker and jailer to 0.21.1, and upgraded this go package to 0.21.0. This worked fine with 0.20.

Here's a strace for the whole thing if that's of any interest:

<details><summary>strace</summary>

```
execve("/usr/local/bin/jailer-0.21.1", ["/usr/local/bin/jailer-0.21.1", "--id", "62fab4aa", "--uid", "123", "--gid", "100", "--exec-file", "/usr/local/bin/firecracker-0.21."..., "--node", "0", "--chroot-base-dir", "/home/jerome/src/github.com/supe"..., "--", "--seccomp-level", "0", "--api-sock", "/run/firecracker.socket"], 0x7fffc464e6e8 /* 57 vars */) = 0
mmap(NULL, 560, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fc873e38000
[...]

PR #219, opened 2020-03-17
Don't hard-code machine api socket path, use value from machine config

jeromegn:
Issue: #217 

*Description of changes:*

Uses machine's `SocketPath` instead of always using the default `/run/firecracker.socket`, allowing for shorter paths.

### DIFF
--- a/jailer.go
+++ b/jailer.go
@@ -30,6 +30,8 @@ const (
 	defaultJailerBin  = "jailer"
 
 	rootfsFolderName = "root"
+
+	defaultSocketPath = "/run/firecracker.socket"
 )
 
 var (
@@ -288,7 +290,14 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 		jailerWorkspaceDir = filepath.Join(defaultJailerPath, filepath.Base(cfg.JailerCfg.ExecFile), cfg.JailerCfg.ID, rootfsFolderName)
 	}
 
-	cfg.SocketPath = filepath.Join(jailerWorkspaceDir, "run", "firecracker.socket")
+	var machineSocketPath string
+	if cfg.SocketPath != "" {
+		machineSocketPath = cfg.SocketPath
+	} else {
+		machineSocketPath = defaultSocketPath
+	}
+
+	cfg.SocketPath = filepath.Join(jailerWorkspaceDir, machineSocketPath)
 
 	stdout := cfg.JailerCfg.Stdout
 	if stdout == nil {
@@ -310,6 +319,7 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 		WithDaemonize(cfg.JailerCfg.Daemonize).
 		WithFirecrackerArgs(
 			"--seccomp-level", cfg.SeccompLevel.String(),
+			"--api-sock", machineSocketPath,
 		).
 		WithStdout(stdout).
 		WithStderr(stderr)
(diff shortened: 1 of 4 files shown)


kzys:
Can you take a look the below errors? 

* git log validation: You don't have a `Signed-off-by` line on your commit.
* tests
* root tests

"test against firecracker master" and "go mod tidy" are okay to fail right now.

jeromegn:
Should be fixed now!

kzys:
@jeromegn Thanks! Looks good now. Could you squash commits into one or two (combine 9f090f8 and f42b48a). We'd like to keep each commit self-contained.

jeromegn:
@kzys done!

jailer.go
```diff
@@ -288,7 +290,18 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
...
+	var machineSocketPath string
+	if cfg.SocketPath != "" {
+		machineSocketPath = cfg.SocketPath
+	} else {
+		machineSocketPath = defaultSocketPath
+	}
+
+	cfg.SocketPath = filepath.Join(jailerWorkspaceDir, machineSocketPath)
+
+	if err := os.MkdirAll(filepath.Dir(cfg.SocketPath), 0600); err != nil {
```
xibz: This needs to be done in a separate handler due to Firecracker creating the initial jailer workspace. You could probably just do it in the `LinkFilesHandler`, for now.

jeromegn: @xibz I thought about it, but the LinkFilesHandler does not have access to the config. It only knows about what was passed to the NaiveChrootStrategy. I would rather not have to add another argument to that constructor, especially since it's repeating the SocketPath.

Open to suggestions!

xibz: It does have access to the config, luckily. I actually prefer it in the handler, so if we could move it there, that'd be awesome.

The [handler](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/handlers.go#L280) is a closure that passes the `Machine` into it. From there you should be able to access the config.

xibz: Actually, after some thought, I prefer this in the jail function. So this is fine as is. Ill take a look at why this is failing in our CI

jailer.go
```diff
@@ -288,7 +292,24 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
...
+
+	cfg.SocketPath = filepath.Join(jailerWorkspaceDir, machineSocketPath)
+
+	jailedSocketBasePath := filepath.Dir(cfg.SocketPath)
+
+	if err := os.MkdirAll(jailedSocketBasePath, 0600); err != nil {
+		return errors.Wrap(err, "failed to create socket path directories")
+	}
+
+	if err := os.Chown(jailedSocketBasePath, *m.Cfg.JailerCfg.UID, *m.Cfg.JailerCfg.GID); err != nil {
```
xibz: `Machine` at this point does not have the config, so we need to use `cfg` here.

jeromegn:
I'm not sure why the jailer tests are failing. There are no `FAIL` block, but make still returns an error.

Not sure about the root test failure either.

kzys:
@jeromegn The below are actual issues.

* https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/797#437e7bbf-86a7-4a27-8724-f3b3363fd172/108-114
* https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/797#03584f1b-4915-4b4f-a3e3-eb41d74c0e6e/356

jeromegn:
I see the issues, but I'm not entirely sure how to fix.

xibz:
@jeromegn - I can take a look at what is happening. I will probably just make the changes and push them up to your branch.

jeromegn:
Sounds good

xibz:
@jeromegn - Don't have permission to push to your fork, so I went ahead and created a new PR, #230. Going to go ahead to close this in favor of #230.

PR #233, opened 2020-05-05
Update client/swagger.yaml from 0.21.0 to b922c7d

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Firecracker is making a few breaking changes. I'd like to estimate how much changes we may need, before their official release.


machine.go
```diff
@@ -613,9 +613,8 @@ func (m *Machine) setupLogging(ctx context.Context) error {
 	}
 
 	l := models.Logger{
-		LogFifo:       String(m.Cfg.LogFifo),
+		LogPath:       String(m.Cfg.LogFifo),
```
kzys, line 616: It may be better to rename m.Cfg.LogFifo to LogPath as well.

kzys:
Let me terminate the EC2 instance. It seems unhealthy. https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/873

machine.go
```diff
@@ -755,13 +761,14 @@ func (m *Machine) createNetworkInterface(ctx context.Context, iface NetworkInter
 	}
 
 	m.logger.Debugf("createNetworkInterface returned for %s", iface.StaticConfiguration.HostDevName)
+	m.networkInterfaces[ifaceID] = ifaceCfg
 	return err
 }
 
 // UpdateGuestNetworkInterfaceRateLimit modifies the specified network interface's rate limits
 func (m *Machine) UpdateGuestNetworkInterfaceRateLimit(ctx context.Context, ifaceID string, rateLimiters RateLimiterSet, opts ...PatchGuestNetworkInterfaceByIDOpt) error {
```
kzys, line 769: The opts will be ignored on the PUT path. How about removing `opts`? Nobody is using the method, at least on GitHub.

https://github.com/search?q=UpdateGuestNetworkInterfaceRateLimit&type=Code

machine.go
```diff
@@ -886,11 +911,21 @@ func (m *Machine) GetMetadata(ctx context.Context, v interface{}) error {
 // parameters of the partialDrive.
 func (m *Machine) UpdateGuestDrive(ctx context.Context, driveID, pathOnHost string, opts ...PatchGuestDriveByIDOpt) error {
```
kzys, line 912: This `opts` will be ignored on the PUT path as well, but nobody is using that on GitHub.

https://github.com/search?q=UpdateGuestDrive&type=Code

machine.go
kzys: I've added LogPath and MetricsPath on c1286946e641dc4fc817e59c0f7455ed31d59942.

machine.go
kzys: @xibz @IRCody What do you think? I think removing that (while nobody is using, at least on GitHub) is okay.

IRCody: If we remove it are we forcing breaking API changes in the future if FC adds an option here?

kzys: It depends how we model their changes on the Go SDK. If Firecracker extends NetworkInterface / PartialNetworkInterface and adding opts makes sense from our perspective, then yes.

Another option is supporting only PatchGuestNetworkInterface. We need to update our tests and we might break other clients (technically Firecracker's change breaks the clients), but it is simpler and we are making less arbitrary decisions.

IRCody: You mean only supporting PatchGuestNetworkInterface in UpdateGuestNetworkInterfaceRateLimit?

kzys: Yes. And doing the same on UpdateGuestDrive()

kzys: I've updated the implementation to only support PatchGuestNetworkInterface (post-boot configuration).

machine.go
kzys: I've reverted the change.

machine.go
```diff
@@ -78,6 +78,9 @@ type Config struct {
 	// should be created.
 	SocketPath string
 
+	// LogPath defines the file path where the Firecracker log is located.
+	LogPath string
+
 	// LogFifo defines the file path where the Firecracker log named-pipe should
 	// be located.
 	LogFifo string
```
kzys, line 86: @xibz @IRCody I'm supporting both LogPath (what Firecracker supports now) and LogFifo (what Firecracker supported before) here. What do you think? Would it be better to let clients break and keep our client thin?

IRCody: I agree lets not break clients for this.

machine.go
```diff
@@ -619,10 +626,32 @@ func (m *Machine) setupLogging(ctx context.Context) error {
...
-		m.Cfg.LogFifo,
-		m.Cfg.MetricsFifo,
-	)
+	m.logger.Debugf("Configured VMM logging to %s", path)
+
+	return nil
+}
+
+func (m *Machine) setupMetrics(ctx context.Context) error {
+	fmt.Println("setupMetrics")
```
IRCody, line 635: m.logger.printf if we still want this message?

machine_test.go
```diff
@@ -418,6 +418,60 @@ func TestStartVMM(t *testing.T) {
 	_, closed := <-m.exitCh
 	assert.False(t, closed)
 }
+func TestLogAndMetrics(t *testing.T) {
```
IRCody, line 421: nit: newline between } and func here.

.buildkite/pipeline.yml
```diff
@@ -85,7 +85,8 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
-
+    soft_fail:
```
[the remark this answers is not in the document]
kzys: After this change, this repos' master cannot be used with Firecracker 0.21.0. So we are going to allow non-zero exit status here.

machine.go
kzys: Right. Will fix.

machine_test.go
kzys: Will do.

PR #237, opened 2020-05-22
Removes the Debug field

xibz:
This commit remove the debug field as it is a really confusing field and
instead relies on WithLogger and WithClient to enable the debugging at
the SDK level


machine.go
```diff
@@ -327,15 +324,12 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
...
-		if cfg.Debug {
-			logger.SetLevel(log.DebugLevel)
-		}
 
 		m.logger = log.NewEntry(logger)
 	}
 
 	if m.client == nil {
-		m.client = NewClient(cfg.SocketPath, m.logger, cfg.Debug)
+		m.client = NewClient(cfg.SocketPath, m.logger, false)
```
kzys, line 332: So the way of setting the flag is giving a new client?

go.mod
```diff
@@ -18,5 +18,7 @@ require (
 	github.com/sparrc/go-ping v0.0.0-20190613174326-4e5b6552494c
 	github.com/stretchr/testify v1.5.1
 	github.com/vishvananda/netlink v1.1.0
+	golang.org/x/crypto v0.0.0-20190617133340-57b3e21c3d56 // indirect
```
IRCody, line 21: are these and the sum changes part of this change?

machine.go
xibz: Yes

go.mod
xibz: Yea, buildkite was complaining

go.sum
```diff
@@ -19,8 +19,6 @@ github.com/containerd/fifo v0.0.0-20191213151349-ff969a566b00/go.mod h1:jPQ2IAeZ
 github.com/containernetworking/cni v0.7.1/go.mod h1:LGwApLUm2FpoOfxTDEeq8T9ipbpZ61X79hmU3w8FmsY=
 github.com/containernetworking/cni v0.7.2-0.20190807151350-8c6c47d1c7fc h1:zUNdrf9w09mWodVhZ9hX4Yk4Uu84n/OgdfPattAwwt8=
 github.com/containernetworking/cni v0.7.2-0.20190807151350-8c6c47d1c7fc/go.mod h1:LGwApLUm2FpoOfxTDEeq8T9ipbpZ61X79hmU3w8FmsY=
```
kzys, line 21: Please separate go.mod/sum changes next time. Doing that in every PR will cause more conflicts between PRs.

go.mod
xibz: This occurs when that auto version bump tool creates PRs. It doesn't do a tidy.

PR #253, opened 2020-07-24
Use Config.VMID as Firecracker's instance ID

kzys:
Firecracker is internally has an instance ID, but the SDK didn't have
the way to configure the ID. This change connects Config.VMID to the
instance ID.

*Issue #, if available:*

NA

*Description of changes:*

* Connects Config.VMID to Firecracker's Instance ID.
* Alternatively we can add Config.InstanceID but I think that would be too confusing.
* Renaming VMID to InstanceID is another option and has less impedance mismatch, but I don't want to break existing customers, including firecracker-containerd.

### DIFF
--- a/machine_test.go
+++ b/machine_test.go
@@ -420,6 +420,9 @@ func TestStartVMM(t *testing.T) {
 }
 
 func TestLogAndMetrics(t *testing.T) {
+	const logLevel = "DEBUG"
+	const vmID = "UserSuppliedVMID"
+
 	dir, err := ioutil.TempDir("", t.Name())
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
@@ -427,6 +430,7 @@ func TestLogAndMetrics(t *testing.T) {
 	socketPath := filepath.Join(dir, "fc.sock")
 
 	cfg := Config{
+		VMID:              vmID,
 		SocketPath:        socketPath,
 		DisableValidation: true,
 		KernelImagePath:   getVmlinuxPath(t),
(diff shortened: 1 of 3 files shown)


machine_test.go
```diff
@@ -420,13 +420,17 @@ func TestStartVMM(t *testing.T) {
 }
 
 func TestLogAndMetrics(t *testing.T) {
+	const logLevel = "DEBUG"
```
xibz, line 423: Should be Debug

kzys:
As always, this is more complicated than I thought.

1. Firecracker's Instance ID doesn't have to be unique because it is only used inside a micro VM.
2. Firecracker Jailer's ID has to be unique, but the jailer passes the ID as its micro VM's instance ID.
3. The SDK's VMID has to be unique.

Using SDK's VMID as a micro VM's instance ID limits what Firecracker can technically do (using the same Instance ID for multiple VMs), while the limitation is enforced by Firecracker's Jailer as well though.

machine_test.go
kzys: The input log level became case-insensitive (https://github.com/firecracker-microvm/firecracker/commit/75587b8b6f7bfefddff4baa7402ff0121367b2af) recently and the output is always uppercase (https://github.com/firecracker-microvm/firecracker/blob/db84b948461bddcc89c0b69aa5639368a7d7c3ef/src/logger/src/logger.rs#L82).

xibz: Ah, cool!

PR #259, opened 2020-09-13
Fix Benchmark Goroutine

alrs:
Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>

From go doc: `testing.B.Fatal()`: "Fatal is equivalent to Log followed by FailNow."

More from `go doc testing.B.FailNow()`:

>  FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.

This ought fix the benchmark tests to not attempt `b.Fatal()` inside a goroutine.

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -3,11 +3,11 @@ package firecracker
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


kzys (approved):

kzys:
Thanks for fixing!

PR #290, opened 2020-11-22
Add support for the MMDS configuration.

yjuba:
The goal of this PR is to support the configuration of MMDS in the SDK.

*Issue #, if available:*

*Description of changes:*
- Add a handler to call `mmds/config` in handlers.go.
- Add a wrapper to call `mmds/config` in firecracker.go.
- Add a wrapped function call to machine.go.

### DIFF
--- a/firecracker.go
+++ b/firecracker.go
@@ -332,6 +332,16 @@ func (f *Client) PatchMmds(ctx context.Context, metadata interface{}, opts ...Pa
 	return f.client.Operations.PatchMmds(params)
 }
 
+// PutMmdsConfig is a wrapper for the swagger generated client to make calling of the
+// API easier.
+func (f *Client) PutMmdsConfig(ctx context.Context, config *models.MmdsConfig) (*ops.PutMmdsConfigNoContent, error) {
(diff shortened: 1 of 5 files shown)


kzys:
The handler looks good. How about adding the IP address to Config struct?

yjuba:
Thank you for your comment.
I was wondering if I should add the IP address to the Config struct, so I would like to add it.

kzys:
Sorry for the late reply. Looks good to me. Can you rebase the branch against master? You don't have to change go.mod and go.sum if they conflict. This PR doesn't add any new dependencies.

yjuba:
Thank you for your comment.
Please check it because it has been rebased.

I haven't made any changes to the code, but the test timed out.
Do you have any idea of ​​the cause?

kzys:
That's odd. Let me restart the BuildKite test.

kzys:
The error we have right now is `find: ‘./testdata/TestCNI/cni.cache/results’: Permission denied` which is not related to this PR.

kzys (approved):

PR #291, opened 2020-11-24
Remove root-owned files

kzys:

*Issue #, if available:*

*Description of changes:*

We run some tests as root user whereas BuildKite's agent is running as non-root due to security concerns. Because of the mismatch, we sometimes have files that cannot be deleted from BuildKite's agent.

This PR does the following, which is larger than what I had expected:
- Clean up files from tests correctly
- Update .buildkite/pipeline.yml to download binaries as non-root
- Update .buildkite/hooks/pre-exit to find files as root
- Remove .buildkite/hooks/pre-checkout, since the hook cannot be executed before cloning this Git repos, which is too late.

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -1,4 +1,11 @@
 #!/bin/bash
+# Since some tests are running as root, they may create root-owned files.
+# These files cannot be deleted by BuildKite Agent since the agent is running
+# as non-root.
 set -euo pipefail
 
-find . -user root -print0 | xargs -0 sudo rm -rf '{}'
+files=$(sudo sh -c "find $BUILDKITE_BUILD_CHECKOUT_PATH -user root -print0 | xargs -0 rm -rfv '{}'")
+
+# If the command above removes files, show and exit with 1.
+echo "$files" | tr "\0" "\n"
+test -z "$files"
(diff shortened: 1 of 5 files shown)


.buildkite/hooks/pre-exit
```diff
@@ -1,4 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-find . -user root -print0 | xargs -0 sudo rm -rf '{}'
+# Remove all root-owned files.
+# Since BuildKite's Agent is running as non-root. Root-owned files cannot be
+# deleted by the agent.
+sudo sh -c 'find . -user root -print0 | xargs -0 rm -rf "{}"'
```
alakesh, line 7: Anything with rm -rf sounds scary. Can we do it in a more controlled way? For example collect the list of file first and then delete each one of them logging them on each deletion?

kzys: How about `rm -rfv`? So that we can log what we've deleted.

FFMMM:
Could we get more context as to the issue we are trying to solve here? What's creating those `root` files to begin with? Why do we need to change this now?

kzys:
@FFMMM Sure. We run some tests as `root` user whereas BuildKite's agent is running as non-root due to security concerns. Because of the mismatch, we sometimes have files that cannot be deleted from BuildKite's agent.

PR #294, opened 2020-11-24
Run tests on GitHub Actions

kzys:
While GitHub Actions' environment doesn't have KVM, we can run
a few tests on there to make sure that the SDK works with different
Go versions.

It doesn't run Firecracker, but running some tests is better than
running no tests.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -25,5 +25,5 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: Build
-      run: make build
+    - name: Test
+      run: make test EXTRAGOARGS=-v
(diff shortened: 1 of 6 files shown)


FFMMM:
Hey @kzys what's the status w this PR -- do we still want to do this work?

Also, would you mind adding a bit more context on the issue we're seeing w Github Actions today? Thanks!

FFMMM:
> what was the issue you've seen?

My bad I wasn't clear -- what I meant was to ask for more context around this PR. Things like:

* why are we creating this PR
* what problem is it solving
* etc.

kzys:
I've update the commit message of 28f8498a78e99293d812041654151f3b0450e099 and the PR's description, but seems we need to merge #342 first.

FFMMM (approved):
lgtm! 🚀

PR #295, opened 2020-11-26
Fix InitrdPath usage

jeromegn:
*Description of changes:*

The way `InitrdPath` was implemented is buggy:

- When you set a `initrd_path` with firecracker, [you do not want a root drive](https://github.com/firecracker-microvm/firecracker/issues/186#issuecomment-725957979).
- There was a type (I believe that's what it was) which prevented `initrdFilename` to ever be set to anything other than `""`. This would make the config for `InitrdPath` useless.

With both these changes, I was able to use `initrd_path` successfully.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -63,7 +63,7 @@ var JailerConfigValidationHandler = Handler{
 			return nil
 		}
 
-		hasRoot := false
+		hasRoot := m.Cfg.InitrdPath != ""
 		for _, drive := range m.Cfg.Drives {
 			if BoolValue(drive.IsRootDevice) {
 				hasRoot = true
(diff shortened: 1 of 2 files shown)


jeromegn:
Actually, I wonder if having a root drive is at odds with an initrd. As in: you can't have a root drive if you have an initrd.

jeromegn:
Anyone want to look into this?

kzys:
Hey. Sorry for the late reply. We've fixed the Go 1.11 issue on master. Could you rebase this branch?

kzys:
Looks good to me. Can you remove `go mod tidy` from this PR? Your PR doesn't bring any new dependencies and go.mod changes would cause conflict with others such as #290.

jeromegn:
My bad, I wanted to make all the tests pass. Looks like another one is failing in the firecracker suite.

kzys:
Yes. There were a few root-owned files which BuildKite Agent couldn't remove.

kzys (approved):

PR #329, opened 2021-03-12
Update download link for kernel, firecracker and jailer

RoyceDavison:
Signed-off-by: Royce Zhao <qiqinzha@amazon.com>

*Issue #, if available:*

*Description of changes:*
Update the download link for kernel, firecracker and jailer in Makefile to keep consistent with Firecracker team. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,13 @@ FIRECRACKER_TARGET?=x86_64-unknown-linux-musl
 UID = $(shell id -u)
 GID = $(shell id -g)
 
+firecracker_version=v0.24.2
+arch=$(shell uname -m)
+
 # The below files are needed and can be downloaded from the internet
-testdata_objects = testdata/vmlinux testdata/root-drive.img testdata/firecracker testdata/jailer
-firecracker_version = v0.23.0
+release_url=https://github.com/firecracker-microvm/firecracker/releases/download/$(firecracker_version)/firecracker-$(firecracker_version)-$(arch).tgz
+testdata_objects = testdata/vmlinux testdata/root-drive.img testdata/binaries testdata/jailer testdata/firecracker
+testdata_dir = testdata/firecracker.tgz testdata/firecracker_spec-$(firecracker_version).yaml testdata/LICENSE testdata/NOTICE testdata/THIRD-PARTY
 
 # --location is needed to follow redirects on github.com
 curl = curl --location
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -56,15 +59,17 @@ distclean: clean
...
 
 testdata/firecracker:
-	$(curl) -o $@ https://github.com/firecracker-microvm/firecracker/releases/download/$(firecracker_version)/firecracker-$(firecracker_version)-x86_64
-	chmod +x $@
+	$(curl) -o testdata/firecracker.tgz ${release_url}
+	tar -xvzf testdata/firecracker.tgz -C ./testdata
+	mv testdata/firecracker-$(firecracker_version)-$(arch) testdata/firecracker
+	chmod +x testdata/firecracker
 
 testdata/jailer:
```
kzys, line 70: How does it get `testdata/jailer-$(firecracker_version)-$(arch)`?

go.sum
```diff
@@ -23,8 +23,7 @@ github.com/containerd/fifo v0.0.0-20191213151349-ff969a566b00 h1:lsjC5ENBl+Zgf38
 github.com/containerd/fifo v0.0.0-20191213151349-ff969a566b00/go.mod h1:jPQ2IAeZRCYxpS/Cm1495vGFww6ecHmMk1YJH2Q5ln0=
```
kzys, line 23: Since you only updated Makefile, your change shouldn't need to modify go.sum/go.mod.

kzys (commented):
Please remove `"..."` around your email address.

Makefile
RoyceDavison: It is includes in the testdata/firecracker.tgz.

go.sum
RoyceDavison: Ok. I will revert this change.

RoyceDavison:
> Please remove `"..."` around your email address.

Makefile
IRCody: If we are downloading them together in one step does it make sense for these to be separate in the Makefile?

IRCody: Maybe the target could be testdata/binaries?

RoyceDavison:
> Please remove `"..."` around your email address.

Fixed it.

Makefile
RoyceDavison: Thanks for @IRCody helps. I move all of them under testdata/binaries

Makefile
```diff
@@ -52,19 +55,20 @@ clean::
 	rm -fr build/
 
 distclean: clean
-	rm -rf $(testdata_objects)
+	rm -rf $(testdata_objects) testdata/firecracker.tgz testdata/firecracker_spec-$(firecracker_version).yaml testdata/LICENSE testdata/NOTICE testdata/THIRD-PARTY
```
FFMMM, line 58: Mostly a nit but I'm just wondering if there is a way we can further simplify this.

Could we `rm -rfv`a `testdata_dir` ?

RoyceDavison: Good catch. Fixed it.

Makefile
```diff
@@ -25,8 +25,11 @@ UID = $(shell id -u)
 GID = $(shell id -g)
 
 # The below files are needed and can be downloaded from the internet
-testdata_objects = testdata/vmlinux testdata/root-drive.img testdata/firecracker testdata/jailer
-firecracker_version = v0.23.0
+firecracker_version=v0.24.2
+arch=$(shell uname -m)
+# The below files are needed and can be downloaded from the internet
```
IRCody, line 30: This comment is duplicated from line 27.

IRCody (changes requested):
LGTM minus the small nit here.

Makefile
kzys: The testdata/ directory has some files. Removing the directory entirely may not work. https://github.com/firecracker-microvm/firecracker-go-sdk/tree/master/testdata

Makefile
RoyceDavison: Fixed it.

Makefile
RoyceDavison: I just pushed an up-to-date Makefile, I create a target test_dir which includes all of file that need to be removed. You can find it [HERE](https://github.com/RoyceDavison/firecracker-go-sdk/blob/66460c9184b6ce7df8f03b6f3f477e197b9e4ed5/Makefile#L33). This will not remove the entire files under /testdatda

kzys (approved):

IRCody (approved):

PR #334, opened 2021-03-25
Don't symlink the files in all steps

kzys:
Since the steps will run in parallel, making symlinks on all of them
would cause a race condition.

*Issue #, if available:*

*Description of changes:*


.buildkite/pipeline.yml
```diff
@@ -40,18 +42,17 @@ steps:
...
       hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
 
+  # Since make test-images mutates the source directory, it cannot be run with other steps.
+  - wait
+
   - label: ':linux: build docker images'
     commands:
-      - 'make test-images'
-      - 'buildkite-agent artifact upload testdata/firecracker-master'
-      - 'buildkite-agent artifact upload testdata/jailer-master'
```
FFMMM, line 47: out of curiosity, why are we not `artifact upload`-ing these anymore?

kzys: We were using that to pass artifacts between tests, but now we use `/tmp/buildkite_build_${BUILDKITE_BUILD_NUMBER}_testdata` instead.

.buildkite/pipeline.yml
```diff
@@ -148,11 +143,12 @@ steps:
   - wait: ~
     continue_on_failure: true
 
-  - label: ':linux: cleanup taps'
+  - label: ':linux: cleanup'
     commands:
       - 'sudo ip tuntap del fc-test-tap${BUILDKITE_BUILD_NUMBER} mode tap'
       - 'sudo ip tuntap del fc-root-tap${BUILDKITE_BUILD_NUMBER} mode tap'
       - 'sudo ip tuntap del fc-mst-tap${BUILDKITE_BUILD_NUMBER} mode tap'
+      - 'sudo rm -fr ${FC_TEST_DATA_PATH}'
```
FFMMM, line 151: any good having a `-v` option passed as well?

FFMMM (approved):
overall, this LGTM! thanks for fixing this 🥇

PR #337, opened 2021-04-20
fix use of (*T).Fatalf from non testing goroutine

alakesh:
*Issue #, if available:*

*Description of changes:*

Using t.Fatalf in the goroutine will only exit the goroutine. Use
t.Errorf instead to return failure to the testing function.


RoyceDavison:
LGTM 🚀

kzys (approved):

PR #338, opened 2021-04-20
Add describeInstance functionality for reading vm info

alakesh:
This will help use cases where we need to check guest's running state
and other information before taking another action.

Signed-off-by: Alakesh Haloi <alakeshh@amazon.com>

*Issue #, if available:*

*Description of changes:*

Getting running state of the VM from firecracker is needed many use cases. Implementing DescribeInstance similar to how GetMetadata is implemented.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -946,6 +946,23 @@ func (m *Machine) UpdateGuestDrive(ctx context.Context, driveID, pathOnHost stri
 	return nil
 }
 
+func (m *Machine) DescribeInstanceInfo(ctx context.Context) (models.InstanceInfo, error) {
+	var instanceInfo models.InstanceInfo
+	resp, err := m.client.GetInstanceInfo(ctx)
+	if err != nil {
+		m.logger.Errorf("Getting Instance Info: %s", err)
+		return instanceInfo, err
+	}
+
+	instanceInfo = *resp.Payload
+	if err != nil {
+		m.logger.Errorf("Getting Instance info failed parsing payload: %s", err)
+	}
+
+	m.logger.Printf("GetInstanceInfo successful")
+	return instanceInfo, err
+}
+
 // refreshMachineConfiguration synchronizes our cached representation of the machine configuration
 // with that reported by the Firecracker API
 func (m *Machine) refreshMachineConfiguration() error {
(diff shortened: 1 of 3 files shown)


machine.go
```diff
@@ -934,6 +934,27 @@ func (m *Machine) GetMetadata(ctx context.Context, v interface{}) error {
 	return nil
 }
 
+// DescribeInstanceInfo gets the VM info and unmarshals it into v
+func (m *Machine) DescribeInstanceInfo(ctx context.Context, v interface{}) error {
```
kzys, line 938: Does it have to take `v interface{}`? The type of the value is fixed, I believe.

kzys (changes requested):

machine.go
alakesh: For these API calls, using v interface{} would give the flexibility for the value to be not tied to one fixed type. I see that, this is the pattern used in rest of the API calls as well.

kzys: We only have GetMetadata() as a getter and that confuses everyone apparently :)

So, GetMetadata() is the wrapper of [MMDS](https://github.com/firecracker-microvm/firecracker/blob/main/docs/mmds/mmds-user-guide.md), which can have any arbitrary data.

DescribeInstanceInfo is different since it is wrapping [this API](https://github.com/firecracker-microvm/firecracker/blob/6292f3289a8a5acdbd15d7e042a89ece95c31fb4/src/api_server/swagger/firecracker.yaml#L800), where we already know the type.

So you should return `(some data, error)` rather than returning `interface{}`. Go is a statically-typed language and types give developers hints about how to consume information correctly.

RoyceDavison: I had the same problem before, and tried to keep consistent with GetMetadata(). @alakesh You may find some useful information from [GetBalloonConfig()](https://github.com/RoyceDavison/firecracker-go-sdk/blob/3e004c7509e3c21a6d772673d06a2eafee7eb8ac/machine.go#L1090)

alakesh: Thanks @RoyceDavison  @kzys, only worry is having inconsistency in the way the APIs are implemented. Updated the PR to use fixed type.

kzys (approved):

PR #341, opened 2021-04-30
Update slack link on README.md

### LINKED ISSUE #340: Outdated Slack link
On https://github.com/firecracker-microvm/firecracker-go-sdk/blob/main/README.md#questions, the Slack link doesn't work: https://tinyurl.com/firecracker-microvm

image

I suppose it should be updated as in https://github.com/firecracker-microvm/firecracker/commit/7b0f0161532692f23d648cdfd1d4cdea2c8fb65b :)

RoyceDavison:
To fix #340: Outdated Slack link

Signed-off-by: Royce Zhao <qiqinzha@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ JAILER_BIN=$(FC_TEST_DATA_PATH)/jailer-master
 UID = $(shell id -u)
 GID = $(shell id -g)
 
-firecracker_version=v0.24.2
+firecracker_version=v0.24.3
 arch=$(shell uname -m)
 
 # The below files are needed and can be downloaded from the internet
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ GitHub issues, but instead follow
 [Firecracker's security reporting guidelines](https://github.com/firecracker-microvm/firecracker/blob/master/SECURITY.md).
 
 Other discussion: For general discussion, please join us in the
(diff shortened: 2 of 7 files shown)


kzys (commented):
While it is better than what we have - the broken link, I have two questions;

- Can we update `tinyurl.com/firecracker-microvm` instead?
- If we cannot, how about using a direct link rather than relying on tinyurl.com? URL Shorteners don't give us much benefits here.

RoyceDavison:
> While it is better than what we have - the broken link, I have two questions;
> 
> * Can we update `tinyurl.com/firecracker-microvm` instead?
> * If we cannot, how about using a direct link rather than relying on tinyurl.com? URL Shorteners don't give us much benefits here.

I tried with the url you just mentioned, unfortunately we can't use it. I replaced the tinyurl with the original url.

kzys (approved):

kzys:
You can mention the issue in the commit message. Then GitHub automatically closes the issue upon merge.

https://guides.github.com/features/issues/#notifications

RoyceDavison:
> You can mention the issue in the commit message. Then GitHub automatically closes the issue upon merge.
> 
> https://guides.github.com/features/issues/#notifications

Good to know that, will use this trick in the next time.

PR #347, opened 2021-05-16
makefile: use devtool for building Firecracker

gc-plp:
Signed-off-by: Gabriel Ionescu <gbi@amazon.com>

*Description of changes:*
This PR makes use of devtool to build the Firecracker and jailer binaries as they're generated upon release.
Devtool is an entry point for launching Firecracker tests, building release binaries and running all the house work around Firecracker.

In this PR the changes consist of removing the Docker container step during the build process and generally simplifying the `make test-images` effort.

The PR currently proposed as an RFC because the `strip` command is being fixed and the `distclean` command is being reviewed. I will mark the PR as ready for review as soon as Firecracker can handle everything proposed here.

Let me know your thoughts on using devtool!  


RoyceDavison:
@gc-plp Thank you for bringing it up! A quick question regarding this, as we chatted on slack, this GO-SDK intentionally build Firecracker `main` to see whether the Go SDK works with Firecracker’s in-development version or not. With this PR, we shouldn't require to use `sudo` for it, correct?

gc-plp:
@RoyceDavison, given that the user which is running the make command can run docker without extra privileges, then yes you shoudn't need `sudo`.

Makefile
```diff
@@ -85,38 +85,24 @@ $(FC_TEST_DATA_PATH)/ltag:
...
-		localhost/$(FIRECRACKER_BUILDER_NAME):$(DOCKER_IMAGE_TAG) \
-		cargo build --release \
-		--target-dir=/artifacts --target $(FIRECRACKER_TARGET) \
-		-p firecracker -p jailer
-	cp build/$(FIRECRACKER_TARGET)/release/firecracker $(FIRECRACKER_BIN)
-	cp build/$(FIRECRACKER_TARGET)/release/jailer $(JAILER_BIN)
+$(FIRECRACKER_BIN) $(JAILER_BIN): $(FIRECRACKER_DIR)
+	# Strip currently commented out because it's being fixed in Firecracker
+	# and is under review.
+	$(FIRECRACKER_DIR)/tools/devtool -y build --release #&& \
```
kzys, line 97: What we need to run the devtool script? Python 2 or 3 and Docker?

Makefile
```diff
@@ -85,38 +85,24 @@ $(FC_TEST_DATA_PATH)/ltag:
...
+		#$(FIRECRACKER_DIR)/tools/devtool strip
+	cp $(FIRECRACKER_DIR)/build/cargo_target/$(FIRECRACKER_TARGET)/release/firecracker $(FIRECRACKER_BIN)
+	cp $(FIRECRACKER_DIR)/build/cargo_target/$(FIRECRACKER_TARGET)/release/jailer $(JAILER_BIN)
```
kzys, line 100: How about having `install` target on the devtool script? Right now the script is hiding Cargo only partially.

kzys (commented):
Overall I like the approach better. I only have a few questions...

Makefile
RoyceDavison: I think they are using Docker. https://github.com/firecracker-microvm/firecracker/blob/391eb1afc2a66b1bdccac0ea083cf56b2ec55a4d/tools/devtool

gc-plp: @kzys as @RoyceDavison pointed out, it's using Docker behind the scenes launched through a bash script.

Makefile
gc-plp: How would you see the `install` command working for multiple binaries?

gc-plp:
`distclean` and `strip` command are now working in the Firecracker repo. Marking this PR as ready for review.

RoyceDavison:
LGTM 🤖

Makefile
kzys: Installing both of them under `/usr/local/bin`? The location should be customizable though.

gc-plp: That would work. Something like `./devtool install --path /path/to/custom/dir` is doable.

We will also need to add the `seccompiler` binary to the install process.

RoyceDavison: @kzys Why do you prefer `install` over `cp`? Also could we merge this update first and see if it get the pipeline fixed or maybe wait for a little bit?

kzys: I'm fine merging the change as is and will do.

Using the devtool is good because it abstracts away Firecracker's build process. We don't have to know Cargo, Rust and other stuff, so Compute Capsule team can do whatever they want, without worrying about breaking downstream dependencies.

However if we hardcode the location of the binaries, [the abstraction is leaky](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/) than what it could be.

kzys (approved):

Makefile
RoyceDavison: Thanks @kzys. Always learn from you. Like this sentence from the article, "So the abstractions save us time working, but they don't save us time learning."

gc-plp: Added a PR in Firecracker to support the `install` subcommand [here](https://github.com/firecracker-microvm/firecracker/pull/2575).

PR #348, opened 2021-07-05
Add LoadSnapshot implementation

gc-plp:
*Description of changes:*
- Added implementation for loading snapshots
- Exposed waitsocket as public
- Added tests for load snapshots

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -385,7 +385,7 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) 
 // handlers succeed, then this will start the VMM instance.
 // Start may only be called once per Machine.  Subsequent calls will return
 // ErrAlreadyStarted.
-func (m *Machine) Start(ctx context.Context) error {
+func (m *Machine) Start(ctx context.Context, opts ...Opt) error {
 	m.logger.Debug("Called Machine.Start()")
 	alreadyStarted := true
 	m.startOnce.Do(func() {
@@ -405,6 +405,10 @@ func (m *Machine) Start(ctx context.Context) error {
 			}
 		}
 	}()
+
+	for _, opt := range opts {
+		opt(m)
+	}
 
 	err = m.Handlers.Run(ctx, m)
 	if err != nil {
--- a/machine.go
+++ b/machine.go
@@ -1085,10 +1085,26 @@ func (m *Machine) CreateSnapshot(ctx context.Context, memFilePath, snapshotPath 
 	return nil
 }
 
+// LoadSnapshot load a snapshot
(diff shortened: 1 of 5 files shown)


machine_test.go
```diff
@@ -1727,6 +1730,83 @@ func TestCreateSnapshot(t *testing.T) {
...
+
+		err = m.cmd.Start()
+		require.NoError(t, err)
+
+		errCh := make(chan error)
+		err = m.WaitForSocket(time.Duration(firecrackerSocketWait)*time.Second, errCh)
+		require.NoError(t, err)
+
+		err = m.LoadSnapshot(ctx, memPath, snapPath)
+		require.NoError(t, err)
```
kzys, line 1799: How about having `StartFromSnapshot(ctx, memPath, snapPath)`? I don't want to expose WaitForSocket().

kzys (changes requested):

machine_test.go
gc-plp: @kzys, submitted another version for the patch as suggested.

I'm not sure if the `WithSnapshot` should ask for `ctx` as parameter, because it requires the `Start()` call to pass `ctx` twice when loading snapshots. WDYT?

machine.go
```diff
@@ -1085,10 +1089,26 @@ func (m *Machine) CreateSnapshot(ctx context.Context, memFilePath, snapshotPath
 	return nil
 }
 
+// LoadSnapshot load a snapshot
+func (m *Machine) LoadSnapshot(ctx context.Context, memFilePath, snapshotPath string, opts ...LoadSnapshotOpt) error {
```
kzys, line 1093: It would be great if we don't have to expose the method. LoadSnapshot() would fail most of the case once the VM is started.

That being said, I don't have a good idea right now.

opts.go
```diff
@@ -47,3 +48,10 @@ func WithProcessRunner(cmd *exec.Cmd) Opt {
 		machine.cmd = cmd
 	}
 }
+
+// WithSnapshot will allow for the machine to start using a given snapshot.
+func WithSnapshot(ctx context.Context, memFilePath, snapshotPath string) Opt {
+	return func(machine *Machine) {
+		machine.LoadSnapshot(ctx, memFilePath, snapshotPath)
```
kzys, line 55: We must not ignore the error here.

machine.go
gc-plp: That's correct, it's not very useful the way it is now.

austinvazquez:
Superseded by #414. Big thanks to @gc-plp who helped pave the way for load snapshot in the SDK.

PR #363, opened 2021-10-12
Drop Go 1.13

kzys:
github.com/go-openapi/runtime's latest version doesn't work with Go
1.13 due to net/http.

Since Go 1.13 is no longer maintained by the Go team, it would be
better to just drop the version rather than asking the package owner
to support the version.

https://pkg.go.dev/net/http#Header.Values

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go: ['1.13', '1.14', '1.15', '1.16', '1.17']
+        go: ['1.14', '1.15', '1.16', '1.17']
         os: ['ubuntu-18.04', 'ubuntu-20.04']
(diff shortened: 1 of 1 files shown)


Kern--:
Ubuntu 18.04 only has ships up to go 1.13 and Ubuntu 20.04 up to go 1.4.

Follow up from this should be to define support policy which probably also defines which distros we consider to be major and at which versions. I think that decision will apply to firectl and firecracker-containerd as well.

kzys:
firecracker-containerd is slightly easier since containerd upgrades Go somewhat aggressively. This SDK and firectl are harder, or we have more freedom to support which Go versions.

I've cut #364 to define supported Go versions. For now, I'd like to drop Go 1.13, update OpenAPI packages to update mongo-driver.

Issue #364, opened 2021-10-13
Define supported Go versions


kzys:
The SDK currently doesn't have a policy regarding supported Go versions. We have been supporting Go 1.11 for a while even Go team doesn't support the version.

Right now, the motivation tends to come from dependencies. If they are having vulnerabilities and we want to upgrade them, we tend to realize that newer dependencies don't support older Go versions.

We probably should have a version supporting policy upfront, instead of deciding case-by-case basis. If supporting old Go versions have some value, we may be okay to ask our dependencies to support them.

ginglis13:
Seeing this on other recent PRs (e.g. #440). Looks like our dependency `ltag` does not support Go <= 1.15:
```
GO111MODULE=off GOBIN=/home/runner/work/firecracker-go-sdk/firecracker-go-sdk/testdata \
go get github.com/kunalkushwaha/ltag
# github.com/kunalkushwaha/ltag
Error: ../../../go/src/github.com/kunalkushwaha/ltag/bash.go:25:15: undefined: os.ReadFile
Error: ../../../go/src/github.com/kunalkushwaha/ltag/dockerfile.go:23:15: undefined: os.ReadFile
Error: ../../../go/src/github.com/kunalkushwaha/ltag/golang.go:23:15: undefined: os.ReadFile
Error: ../../../go/src/github.com/kunalkushwaha/ltag/makefile.go:15:15: undefined: os.ReadFile
make: *** [Makefile:125: testdata/ltag] Error 2
Error: Process completed with exit code 2.
```
`os.ReadFile` was added in Go 1.16

Joffref:
Maybe, version of the ltag binary must be setted using `go get github.com/kunalkushwaha/ltag@version` in the [Makefile](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/dd9dd9395c36b19cff8000a9c3e72b772805f005/Makefile#L124-L127) to provides two versions, one for 1.15 <= and one for 1.16 >=  ?

austinvazquez:
To me the way is to drop support for Go <= v1.16. v1.17 just ended support in August 2022 and users may still be in transition.

Also we should pin the dependency versions for reproducible builds but that is a separate issue.

Kern--:
[...]

PR #370, opened 2022-01-18
Upgrade Go from 1.13 to 1.15

kzys:
- The first commit is upgrading Go from 1.13 to 1.15. `go-openapi` needs Go 1.15, which we probably should adopt rather than stop updating. Go 1.15 and Go 1.13 are no longer supported according to https://go.dev/doc/devel/release. We need to discuss #364 later.
- The second commit is making Firecracker's main branch build "soft fail", meaning that the build breakage wouldn't make the whole build broken. Firecracker is making some breaking changes before 1.0. So we cannot make the SDK work with the main branch and released versions.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 env:
-  PATH: "/usr/lib/go-1.13/bin:/usr/bin"
+  PATH: "/usr/lib/go-1.15/bin:/usr/bin"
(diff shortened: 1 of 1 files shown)


kzys:
The `main` branch build wouldn't work due to #371.

kzys:
Another blocker is https://github.com/firecracker-microvm/firecracker/pull/2876.

Firecracker 1.0 is coming. So we need to take these changes anyway later.

aarongoulet (approved):

Kern-- (approved):

austinvazquez (approved):

Issue #379, opened 2022-02-07
Refactor firecracker-containerd's vsock module to be reused for other firecracker applications


austinvazquez:
I would like to use firecracker-containerd's internal [vsock](https://github.com/austinvazquez/firecracker-containerd/blob/main/internal/vm/vsock.go) module.

This issue is to refactor that module into a package here in the sdk.

PR #380, opened 2022-02-07
Refactor: pull in firecracker-containerd's internal vsock module as a package

austinvazquez:
*Issue #, if available:*
#379 

*Description of changes:*
Pull in firecracker-containerd's internal [vsock](https://github.com/austinvazquez/firecracker-containerd/blob/main/internal/vm/vsock.go) module and refactor into a reusable package.


vsock/dialer.go
```diff
@@ -0,0 +1,207 @@
...
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/containerd/containerd/log"
```
kzys, line 24: We shouldn't use containerd from this package.

vsock/listener.go
```diff
@@ -0,0 +1,109 @@
...
+// permissions and limitations under the License.
+
+package vsock
+
+import (
+	"context"
+	"net"
+	"time"
+
+	"github.com/containerd/containerd/log"
```
kzys, line 21: Ditto.

kzys (changes requested):

vsock/dialer.go
austinvazquez: Ah, fair point. Let me rework that dependency.

vsock/dialer.go
```diff
@@ -0,0 +1,207 @@
...
+// path and port. It will retry connect attempts if a temporary error is encountered (up
+// to a fixed timeout) or the provided request is canceled.
+func Dial(ctx context.Context, udsPath string, port uint32) (net.Conn, error) {
+	d := vsockDialer{ctx: ctx, timeout: DefaultTimeouts()}
+	return d.Dial(udsPath, port)
+}
+
+type vsockDialer struct {
+	timeout VsockTimeout
+	ctx     context.Context
```
Kern--, line 57: Generally, it's an anti-pattern to store a context in a struct because it is possible to construct and use an object in different contexts (e.g. a smaller scoped context with a shorter timeout). I know it technically doesn't matter for this because only vsock.Dial is exposed, but it's good to keep convention for easier refactoring.

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

https://pkg.go.dev/context

vsock/dialer.go
```diff
@@ -0,0 +1,207 @@
...
+	"net"
+	"strings"
+	"time"
+
+	"github.com/containerd/containerd/log"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type VsockTimeout struct {
```
Kern--, line 29: Since this package is vsock, we don't need `Vsock` in struct/function names.

e.g. `vsock.Timeout` is fine, we don't need `vsock.VsockTimeout`

https://go.dev/doc/effective_go#package-names

vsock/dialer.go
kzys: Good point. Dial() could take a context. It is different from other Dial functions do, but these Dial functions are implemented before context.

austinvazquez: Removed this change and pass the context for dial via input parameter. Listener is another story because we implement the net.Listener interface. At the very least this change doesn't make the problem worse.

kzys (approved):

kzys:
Oh 480f6510150a6489c8c3a5b2bfd979926c136805 doesn't have Signed-off line.

austinvazquez:
> Oh [480f651](https://github.com/firecracker-microvm/firecracker-go-sdk/commit/480f6510150a6489c8c3a5b2bfd979926c136805) doesn't have Signed-off line.

Pushed commit with sign-off.

Kern-- (approved):

PR #384, opened 2022-02-17
fix duplicate test imports

alrs:
`github.com/firecracker-microvm/firecracker-go-sdk/client/operations` was being imported twice in `machine_test.go`.


kzys (approved):
Looks good! I wonder why we couldn't notice that...

PR #401, opened 2022-04-27
Add a DriveOpt func for configuring CacheType

swagatbora90:
*Issue #, if available:*

*Description of changes:*
Adding a optional function to set "CacheType" of a Drive. "CacheType" is an optional field that we can now set for models.Drive and supports 2 values: "Unsafe" and "Writeback".

The aim is to make use of this function in firecracker-containerd if client requests CacheType to be set for the root drive or the other mounted drives

*Testing:*
make test

### DIFF
--- a/drives.go
+++ b/drives.go
@@ -103,3 +103,10 @@ func WithRateLimiter(limiter models.RateLimiter) DriveOpt {
 		d.RateLimiter = &limiter
 	}
 }
+
+// WithCacheType sets the cache strategy for the block device
+func WithCacheType(cacheType string) DriveOpt {
+	return func(d *models.Drive) {
+		d.CacheType = String(cacheType)
+	}
+}
--- a/drives_test.go
+++ b/drives_test.go
@@ -32,7 +32,7 @@ func TestDrivesBuilder(t *testing.T) {
 
 	drives := NewDrivesBuilder(expectedPath).Build()
 	if e, a := expectedDrives, drives; !reflect.DeepEqual(e, a) {
-		t.Errorf("expected drives %v, but received %v", e, a)
+		t.Errorf("expected drives %+v, but received %+v", e, a)
 	}
 }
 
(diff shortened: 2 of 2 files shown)


drives.go
```diff
@@ -103,3 +103,15 @@ func WithRateLimiter(limiter models.RateLimiter) DriveOpt {
 		d.RateLimiter = &limiter
 	}
 }
+
+// WithCacheType sets the cache strategy for the block device
+func WithCacheType(cache_type string) DriveOpt {
```
kzys, line 108: Generally variables in Go are camelCase.

drives.go
```diff
@@ -103,3 +103,15 @@ func WithRateLimiter(limiter models.RateLimiter) DriveOpt {
...
+	if cache_type == "" {
+		return func(_ *models.Drive) {
+			// no-op
+		}
+	}
```
kzys, line 113: Seems we don't do this kind of check in most of helpers here. Do you have specific reasons to do here?

One concern I have is that if someone pass `""` unintentionally, it would be ignored. Should we pass `""` as is and let Firecracker error out?

kzys (changes requested):

drives.go
swagatbora90: Ack. Will change to camelCase

drives.go
swagatbora90: I do see your point about letting Firecracker handle everything as is. I added this input validation since I might have to use this later from FC-CD and pass `""`. Though I can add a helper in FC-CD to do just that. 

Will remove this check from here.

kzys (commented):
The change itself looks good. Can you squash the two commits into one? It may not be much meaningful for future readers.

kzys (approved):

aarongoulet (approved):
Looks good to me. :shipit:

PR #410, opened 2022-06-14
Added API functionality + testing to get Firecracker Version

sondavidb:
Added the ability for users to get the current version of Firecracker being used, and added tests in both firecracker_test.go and machine_test.go.

Functions were added in where it made some logical sense to me. If it's not consistent with the overall structure please let me know, as I couldn't seem to get the exact structure of the order of the functions in the files. Also, please pay attention to how I tested the function in machine_test.go, as I'm not super confident that's where it should be, but it feels right.

### DIFF
--- a/firecracker.go
+++ b/firecracker.go
@@ -75,6 +75,22 @@ func NewClient(socketPath string, logger *logrus.Entry, debug bool, opts ...Clie
 	return c
 }
 
+// GetFirecrackerVersionOpt is a functional option to be used for the
+// GetFirecrackerVersion API in setting any additional optional fields.
+type GetFirecrackerVersionOpt func(*ops.GetFirecrackerVersionParams)
+
+// GetFirecrackerVersion is a wrapper for the swagger generated client to make
+// calling of the API easier.
+func (f *Client) GetFirecrackerVersion(ctx context.Context, opts ...GetFirecrackerVersionOpt) (*ops.GetFirecrackerVersionOK, error) {
+	params := ops.NewGetFirecrackerVersionParams()
+	params.SetContext(ctx)
+	for _, opt := range opts {
+		opt(params)
+	}
+
+	return f.client.Operations.GetFirecrackerVersion(params)
+}
+
 // PutLoggerOpt is a functional option to be used for the PutLogger API in
 // setting any additional optional fields.
 type PutLoggerOpt func(*ops.PutLoggerParams)
--- a/machine.go
+++ b/machine.go
@@ -429,6 +429,18 @@ func (m *Machine) Wait(ctx context.Context) error {
 	}
 }
 
+// GetFirecrackerVersion gets the machine's firecracker version and returns it
+func (m *Machine) GetFirecrackerVersion(ctx context.Context) (string, error) {
+	resp, err := m.client.GetFirecrackerVersion(ctx)
+	if err != nil {
+		m.logger.Errorf("Getting firecracker version: %s", err)
+		return "", err
+	}
+
+	m.logger.Debug("GetFirecrackerVersion successful")
+	return *resp.Payload.FirecrackerVersion, nil
+}
+
 func (m *Machine) setupNetwork(ctx context.Context) error {
 	err, cleanupFuncs := m.Cfg.NetworkInterfaces.setupNetwork(ctx, m.Cfg.VMID, m.Cfg.NetNS, m.logger)
 	m.cleanupFuncs = append(m.cleanupFuncs, cleanupFuncs...)
--- a/machine_test.go
+++ b/machine_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"flag"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
@@ -25,6 +26,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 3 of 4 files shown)


firecracker.go
```diff
@@ -75,6 +75,22 @@ func NewClient(socketPath string, logger *logrus.Entry, debug bool, opts ...Clie
 	return c
 }
 
+// GetFirecrackerVersionOpt is a functional option to be used for the
+// GetFirecrackerVersion API in setting any additional optional fields.
+type GetFirecrackerVersionOpt func(*ops.GetFirecrackerVersionParams)
+
+// CreateSnapshot is a wrapper for the swagger generated client to make
```
austinvazquez, line 82: Comment here needs to be `GetFirecrackerVersion`

machine_test.go
```diff
@@ -620,6 +621,13 @@ func testCreateMachine(ctx context.Context, t *testing.T, m *Machine) {
 	}
 }
 
+func testGetFirecrackerVersion(ctx context.Context, t *testing.T, m *Machine) {
+	_, err := m.getFirecrackerVersion(ctx)
```
austinvazquez, line 625: @kzys for assertion purposes would we prefer `firecracker_version` in Makefile be an environment variable so we can read it? Or perhaps using some regex to grab it dynamically `firecracker --version`? I think I prefer the former.

machine.go
```diff
@@ -429,6 +429,18 @@ func (m *Machine) Wait(ctx context.Context) error {
...
 
+// GetFirecrackerVersion gets the machine's firecracker version and returns it
+func (m *Machine) getFirecrackerVersion(ctx context.Context) (string, error) {
+	resp, err := m.client.GetFirecrackerVersion(ctx)
+	if err != nil {
+		m.logger.Errorf("Getting firecracker version: %s", err)
+		return "", err
+	}
+
+	m.logger.Printf("getFirecrackerVersion successful")
```
ginglis13, line 440: nit: probably want this as a debug message `m.logger.Debug`

machine_test.go
sondavidb: I just made a change so that testGetFirecrackerVersion is now as follows: 

`func testGetFirecrackerVersion(ctx context.Context, t *testing.T, m *Machine) {
	ver, err := m.GetFirecrackerVersion(ctx)

	if err != nil {
		t.Errorf("GetFirecrackerVersion: %v", err)
	}

	if ver != "1.0.0" {
		t.Errorf("GetFirecrackerVersion: Exppected version 1.0.0, got version %v", ver)
	}
}`

This runs fine on my end using Firecracker version 1.0.0, but when tested on here it fails, as you can see below, and I'm not entirely sure if I can get the expected firecracker version from the makefile. Any suggestions?

sondavidb: A bit more info:

Testing this against (I assume) the code in the branch seems to run perfectly fine, with no errors.

However, the pipeline also tests on the mainline of Firecracker, and that's when I get the following error:

```
=== RUN   TestMicroVMExecution/TestGetFirecrackerVersion
    machine_test.go:632: GetFirecrackerVersion: Exppected version 1.0.0, got version 1.1.0-35-g0fa080b1-dirty<!--
```

Looks like the failing test is the one running against the tip of Firecracker instead of any specific one specified by the Makefile, which I guess scratches off the idea of taking the version from the Makefile.

Even more tempted now to re-remove `ver` and subsequent checks on it but I just want to double check with you guys before going through with it.

austinvazquez: @sondavidb how about we try the following:

1. Use `firecracker_version` as an environment variable when running the tests.
2. Read the expected value from the environment variable during test execution.
3. Create a test utility function for sanitizing the dirty version returned by firecracker. For reference see https://github.com/firecracker-microvm/firecracker/blob/b3cc331794013419726446f9eb42d6e64f65d135/tests/framework/utils.py#L641

machine_test.go
```diff
@@ -620,6 +621,18 @@ func testCreateMachine(ctx context.Context, t *testing.T, m *Machine) {
...
 
+func testGetFirecrackerVersion(ctx context.Context, t *testing.T, m *Machine) {
+	ver, err := m.GetFirecrackerVersion(ctx)
+
+	if err != nil {
+		t.Errorf("GetFirecrackerVersion: %v", err)
+	}
+
+	if ver != "1.0.0" {
+		t.Errorf("GetFirecrackerVersion: Exppected version 1.0.0, got version %v", ver)
```
kzys, line 632: Typo: Exppected

Also you can use require.Equal/assert.Equal instead.

machine_test.go
austinvazquez: Spoke with @sondavidb offline on why the environment variable would prove challenging in pipeline steps were we run against firecracker main.

@sondavidb, how about the following approach to read the version from the currently installed firecracker binary. Should make for more robust testing.

```go
func parseVersionFromStdout(stdout []byte) (string, error) {
	pattern := regexp.MustCompile(`Firecracker v(?P<version>[0-9]\.[0-9]\.[0-9])`)
	matches := pattern.FindStringSubmatch(string(stdout))
	if len(matches) != 2 {
		return "", fmt.Errorf("Unable to parse firecracker version from stdout")
	}
	return matches[pattern.SubexpIndex("version")], nil
}

func getFirecrackerVersion() (string, error) {
	cmd := exec.Command("firecracker", "--version")
	stdout, err := cmd.Output()
	if err != nil {
		return "", err
	}
	return parseVersionFromStdout(stdout)
}
```

machine.go
```diff
@@ -429,6 +429,18 @@ func (m *Machine) Wait(ctx context.Context) error {
...
 
+// GetFirecrackerVersion gets the machine's firecracker version and returns it
+func (m *Machine) GetFirecrackerVersion(ctx context.Context) (string, error) {
+	resp, err := m.client.GetFirecrackerVersion(ctx)
+	if err != nil {
+		m.logger.Errorf("Getting firecracker version: %s", err)
+		return "", err
+	}
+
+	m.logger.Debug("getFirecrackerVersion successful")
```
austinvazquez, line 440: Minor nit: GetFirecrackerVersion*

austinvazquez (approved):

ginglis13 (approved):
LGTM!

kzys (approved):
LGTM!

PR #411, opened 2022-06-17
Added API functionality + testing to change engine IO type

sondavidb:
The Drive model has a variable called `IoEngine` that determines if the block device runs commands synchronously or asynchronously (default synchronous). This PR implements a function to change said variable alongside a test to show that it works.

The default value seems to be nil and I assume that the handling of that is done in Firecracker itself, but if this is not the case, please let me know.

### DIFF
--- a/drives.go
+++ b/drives.go
@@ -110,3 +110,12 @@ func WithCacheType(cacheType string) DriveOpt {
 		d.CacheType = String(cacheType)
 	}
 }
+
+// WithIoEngine sets the io engine of the drive
+// Defaults to Sync, Async is in developer preview at the moment
+// https://github.com/firecracker-microvm/firecracker/blob/v1.1.0/docs/api_requests/block-io-engine.md
+func WithIoEngine(ioEngine string) DriveOpt {
+	return func(d *models.Drive) {
+		d.IoEngine = String(ioEngine)
+	}
+}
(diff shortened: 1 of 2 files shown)


drives.go
```diff
@@ -110,3 +110,10 @@ func WithCacheType(cacheType string) DriveOpt {
 		d.CacheType = String(cacheType)
 	}
 }
+
+// WithIoEngine sets the io engine of the drive
+func WithIoEngine(ioEngine string) DriveOpt {
```
austinvazquez, line 115: Would it be better to perhaps have two methods, one for each setting. e.g. `WithSyncIoEngine` and `WithAsyncIoEngine`. Also it might be worth documenting that `Sync` mode is default and `Async` mode is currently in developer preview with link to firecracker documentation.

sondavidb: It would make it inconsistent with the other functions in `drive.go`. Other functions are similar in that they effectively take enumerators and have functions to validate their values to match those in the enumerators. Based on that I don't see a reason to split the functions.

The block of code now reads as follows:
```
// WithIoEngine sets the io engine of the drive
// Defaults to Sync, Async is in developer preview at the moment
// https://github.com/firecracker-microvm/firecracker/blob/main/docs/api_requests/block-io-engine.md
func WithIoEngine(ioEngine string) DriveOpt {
	return func(d *models.Drive) {
		d.IoEngine = String(ioEngine)
	}
}
```

austinvazquez: Yeah I guess that is fair enough. What you have looks good to me then, just point the link to a pinned version of firecracker. e.g. `https://github.com/firecracker-microvm/firecracker/blob/v1.1.0/docs/api_requests/block-io-engine.md`

ginglis13 (approved):
LGTM

austinvazquez (approved):

PR #414, opened 2022-06-22
Implemented snapshot loading capabilities

sondavidb:

Taking most of the code from #348 and modifying it a bit, I was able to implement the ability to load snapshots as an option within the `Start` function in `machine.go` via `WithSnapshot`. This fixes the issue addressed in #348, where `machine.go` exposing `LoadSnapshot` function in said PR was deemed suboptimal due to its guaranteed failure outside of the scope of `Start`.

The actual implemention in `opts.go` initially felt a bit messy, as I felt the extra added imports made it clunky, but @austinvazquez and I looked over it today and decided it was fine given that it ultimately did similar things compared to other options, but feel free to disagree below if you feel otherwise.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -78,6 +78,9 @@ steps:
       - ln -s /var/lib/fc-ci/rootfs.ext4 ${FC_TEST_DATA_PATH}/root-drive.img
       # Download Firecracker and its jailer.
       - make deps
+      # Build a rootfs with SSH enabled.
+      - sudo -E FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make ${FC_TEST_DATA_PATH}/root-drive-ssh-key
+      - stat ${FC_TEST_DATA_PATH}/root-drive-ssh-key ${FC_TEST_DATA_PATH}/root-drive-with-ssh.img
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
@@ -110,7 +113,7 @@ steps:
 
   - label: ':hammer: root tests'
     commands:
-      - "sudo PATH=$PATH FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make test EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
+      - "sudo -E PATH=$PATH FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make test EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ root-drive.img
 TestPID.img
 build/
 testdata/fc.stamp
+testdata/bin/
 testdata/logs/
 testdata/ltag
 testdata/release-*
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ FIRECRACKER_DIR=build/firecracker
 FIRECRACKER_TARGET?=x86_64-unknown-linux-musl
 
 FC_TEST_DATA_PATH?=testdata
+FC_TEST_BIN_PATH:=$(FC_TEST_DATA_PATH)/bin
 FIRECRACKER_BIN=$(FC_TEST_DATA_PATH)/firecracker-main
 JAILER_BIN=$(FC_TEST_DATA_PATH)/jailer-main
 
@@ -37,7 +38,17 @@ $(FC_TEST_DATA_PATH)/vmlinux \
 $(FC_TEST_DATA_PATH)/root-drive.img \
 $(FC_TEST_DATA_PATH)/jailer \
 $(FC_TEST_DATA_PATH)/firecracker \
-$(FC_TEST_DATA_PATH)/ltag
+$(FC_TEST_DATA_PATH)/ltag \
+$(FC_TEST_BIN_PATH)/ptp \
+$(FC_TEST_BIN_PATH)/host-local \
+$(FC_TEST_BIN_PATH)/static \
+$(FC_TEST_BIN_PATH)/tc-redirect-tap
+
+# Enable pulling of artifacts from S3 instead of building
+# TODO: https://github.com/firecracker-microvm/firecracker-go-sdk/issues/418
+ifeq ($(GID), 0)
+testdata_objects += $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img $(FC_TEST_DATA_PATH)/root-drive-ssh-key
+endif
 
 testdata_dir = testdata/firecracker.tgz testdata/firecracker_spec-$(firecracker_version).yaml testdata/LICENSE testdata/NOTICE testdata/THIRD-PARTY
 
@@ -81,6 +92,34 @@ $(FC_TEST_DATA_PATH)/fc.stamp:
 
 $(FC_TEST_DATA_PATH)/root-drive.img:
 	$(curl) -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/fsfiles/hello-rootfs.ext4
+
+$(FC_TEST_DATA_PATH)/root-drive-ssh-key $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img: 
+# Need root to move ssh key to testdata location
+ifeq ($(GID), 0)
+	$(MAKE) $(FIRECRACKER_DIR)
+	$(FIRECRACKER_DIR)/tools/devtool build_rootfs
+	cp $(FIRECRACKER_DIR)/build/rootfs/bionic.rootfs.ext4 $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img
+	cp $(FIRECRACKER_DIR)/build/rootfs/ssh/id_rsa $(FC_TEST_DATA_PATH)/root-drive-ssh-key
+	rm -rf $(FIRECRACKER_DIR)
+else
+	$(error unable to place ssh key without root permissions)
+endif
+
+$(FC_TEST_BIN_PATH)/ptp:
+	GO111MODULE=off GOBIN=$(abspath $(FC_TEST_BIN_PATH)) \
+	go get github.com/containernetworking/plugins/plugins/main/ptp
+
+$(FC_TEST_BIN_PATH)/host-local:
+	GO111MODULE=off GOBIN=$(abspath $(FC_TEST_BIN_PATH)) \
+	go get github.com/containernetworking/plugins/plugins/ipam/host-local
+
+$(FC_TEST_BIN_PATH)/static:
+	GO111MODULE=off GOBIN=$(abspath $(FC_TEST_BIN_PATH)) \
+	go get github.com/containernetworking/plugins/plugins/ipam/static
+
+$(FC_TEST_BIN_PATH)/tc-redirect-tap:
+	GO111MODULE=off GOBIN=$(abspath $(FC_TEST_BIN_PATH)) \
+	go get github.com/awslabs/tc-redirect-tap/cmd/tc-redirect-tap
 
 $(FC_TEST_DATA_PATH)/ltag:
 	GO111MODULE=off GOBIN=$(abspath $(FC_TEST_DATA_PATH)) \
--- a/machine.go
+++ b/machine.go
@@ -151,6 +151,13 @@ type Config struct {
 	// It is possible to use a valid IPv4 link-local address (169.254.0.0/16).
 	// If not provided, the default address (169.254.169.254) will be used.
 	MmdsAddress net.IP
+
+	// Configuration for snapshot loading
+	Snapshot SnapshotConfig
+}
+
+func (cfg *Config) hasSnapshot() bool {
+	return cfg.Snapshot.MemFilePath != "" || cfg.Snapshot.SnapshotPath != ""
 }
 
 // Validate will ensure that the required fields are set and that
@@ -381,7 +388,7 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) 
 // handlers succeed, then this will start the VMM instance.
 // Start may only be called once per Machine.  Subsequent calls will return
 // ErrAlreadyStarted.
-func (m *Machine) Start(ctx context.Context) error {
+func (m *Machine) Start(ctx context.Context, opts ...StartOpt) error {
 	m.logger.Debug("Called Machine.Start()")
 	alreadyStarted := true
 	m.startOnce.Do(func() {
@@ -401,6 +408,10 @@ func (m *Machine) Start(ctx context.Context) error {
 			}
 		}
 	}()
+
+	for _, opt := range opts {
+		opt(m)
+	}
 
 	err = m.Handlers.Run(ctx, m)
 	if err != nil {
@@ -854,6 +865,10 @@ func (m *Machine) addVsock(ctx context.Context, dev VsockDevice) error {
 }
 
 func (m *Machine) startInstance(ctx context.Context) error {
+	if m.Cfg.hasSnapshot() {
+		return nil
+	}
+
 	action := models.InstanceActionInfoActionTypeInstanceStart
 	info := models.InstanceActionInfo{
 		ActionType: &action,
@@ -1102,6 +1117,23 @@ func (m *Machine) CreateSnapshot(ctx context.Context, memFilePath, snapshotPath 
 	}
 
 	m.logger.Debug("snapshot created successfully")
+	return nil
+}
+
+// loadSnapshot loads a snapshot of the VM
+func (m *Machine) loadSnapshot(ctx context.Context, snapshot *SnapshotConfig) error {
+	snapshotParams := &models.SnapshotLoadParams{
+		MemFilePath:         &snapshot.MemFilePath,
+		SnapshotPath:        &snapshot.SnapshotPath,
+		EnableDiffSnapshots: snapshot.EnableDiffSnapshots,
+		ResumeVM:            snapshot.ResumeVM,
+	}
+
+	if _, err := m.client.LoadSnapshot(ctx, snapshotParams); err != nil {
+		return fmt.Errorf("failed to load a snapshot for VM: %v", err)
+	}
+
+	m.logger.Debug("snapshot loaded successfully")
 	return nil
 }
 
--- a/machine_test.go
+++ b/machine_test.go
@@ -38,6 +38,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/ssh"
 
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 	ops "github.com/firecracker-microvm/firecracker-go-sdk/client/operations"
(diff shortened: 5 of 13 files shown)


opts.go
```diff
@@ -47,3 +49,20 @@ func WithProcessRunner(cmd *exec.Cmd) Opt {
...
+		snapshotParams := &models.SnapshotLoadParams{
+			MemFilePath:  String(memFilePath),
+			SnapshotPath: String(snapshotPath),
+		}
+
+		if _, err := m.client.LoadSnapshot(ctx, snapshotParams, opts...); err != nil {
+			m.logger.Errorf("failed to load a snapshot for VM: %v", err)
+			return
+		}
+
```
kzys, line 65: If err != nil, I think Start() should return the error.

Right now, the tests are passing but logging

```
time="2022-06-22T18:18:23Z" level=error msg="failed to load a snapshot for VM: Put \"http://localhost/snapshot/load\": dial unix /tmp/TestCreateLoadSnapshot688360910/TestCreateLoadSnapshot.load: connect: no such file or directory"
```

kzys (changes requested):

opts.go
sondavidb: I understand and think we should find a way to do that, but the current `Opt` type doesn't return anything, so we would have to change the type to return an error. It doesn't seem that the scope of this change would be a big deal (currently only seeing references to the type within `opts.go`), but as these functions are exposed, in theory it could break something as we're moving from returning nothing to returning an error. I think this is alright, but wanted to make sure you agreed before moving forward with the changes.

I feel this change would allow for better error handling down the line should we add any more `WithFoo`-esque commands that require more complex computations than setting a variable; however, the above concern still remains.

An alternative approach is to add some machine variables, have `WithSnapshot` set aforementioned parameters, then checks if those parameters have been set after all the options are called, then call `client.LoadSnapshot` with said parameters if so.

kzys: Well does it make sense to have StartOpt instead of reusing Opt? NewMachine wouldn't take WithSnapshot and Start wouldn't take WithProcessRunner.

sondavidb: That makes sense, yes. I'll make the adjustments and do some debugging once I figure out why the snapshot is failing to load.

sondavidb: Deleted my previous comment as it contained incorrect information.

@austinvazquez and I combed through the code and figured out that snapshots were not loading with the previous implementation due to the lack of a vsock when attempting to load the snapshot. However, trying to do so after the handler is not allowed, so we had to make the snapshot loader a handler that would be dynamically inserted before the handler that starts the VMM.

I also dropped the requirement for `StartOpt()` to return an error, as it now only appends to the handler array. If `LoadSnapshot()` itself fails, it will be handled via `m.Handlers.Run()` now.

It seems to pass the tests after these changes, as the handler's `Run()` function requires no errors, but when adding metadata to a machine before creating the snapshot, it is not loaded onto the snapshot, leading me to believe that either snapshots do not preserve metadata or `LoadSnapshot()` is somehow not functioning correctly on the Firecracker side, so I presume in the scope of this problem this is not something I should worry about, but feel free to correct me if my assumption is wrong.

machine_test.go
```diff
@@ -1728,6 +1728,76 @@ func TestCreateSnapshot(t *testing.T) {
...
+	defer os.Remove(snapPath)
+	defer os.Remove(memPath)
+
+	// Tee logs for validation:
+	var logBuffer bytes.Buffer
+	machineLogger := logrus.New()
+	machineLogger.Out = io.MultiWriter(os.Stderr, &logBuffer)
+
+	// Create a snapshot
+	{
```
kzys, line 1755: You should use t.Run(...) instead.

machine_test.go
```diff
@@ -1728,6 +1728,76 @@ func TestCreateSnapshot(t *testing.T) {
...
+		err = m.Start(ctx, WithSnapshot(memPath, snapPath))
+		require.NoError(t, err)
+
+		err = m.ResumeVM(ctx)
+		require.NoError(t, err)
+
+		err = m.StopVMM()
+		require.NoError(t, err)
+	}
+
```
kzys, line 1798: Can you add a test case where you don't have a valid snapshot? Does it return an error correctly?

opts.go
```diff
@@ -47,3 +48,11 @@ func WithProcessRunner(cmd *exec.Cmd) Opt {
 		machine.cmd = cmd
 	}
 }
+
+// WithSnapshot will allow for the machine to start using a given snapshot.
+func WithSnapshot(memFilePath, snapshotPath string, opts ...LoadSnapshotOpt) StartOpt {
+	return func(m *Machine) {
+		m.Handlers.FcInit.AppendAfter("SetupKernelArgsHandler",
+			NewLoadSnapshotHandler(memFilePath, snapshotPath, opts...))
+	}
```
kzys, line 57: 1) Don't we need to remove existing handlers when you have the snapshot handler? For example, you cannot start Linux after loading a snapshot.

2) Is it possible to validate some of the options in the snapshot case? For example, the kernel image won't be loaded if the VM is started from a snapshot.

sondavidb:
Made changes that included both a config file and an extra handler to load snapshots, so that we can choose when to call it during device setup.

I tried to do @kzys's approach of having LoadSnapshotHandler after AttachDrivesHandler but having either one in front of the other makes Firecracker unhappy and gives me an error saying that whatever handler is called second cannot be called after start config is set (e.g. AttachDrivesHandler -> LoadSnapshotHandler makes Firecracker complain about LoadSnapshotHandler, while LoadSnapshotHandler -> AttachDrivesHandler makes Firecracker complain about LoadSnapshot handler, both for the same reason). Based on this information, I feel like either a. Attaching drives is handled by loading the snapshot, or b. I'm not fully understanding how LoadSnapshot works.

In any case, the above seems to run fine, so take that as you will. I'm planning on adding some testing tomorrow to make sure that the proper snapshot is being loaded. This will be done as follows:

- Create MicroVM
- ssh into it
- Run `sleep 492` (or some other weird uncommon value)
- Run CreateSnapshot
- Close the VM
- Create another MicroVM loaded from that snapshot
- ssh into that
- see if `ps -aux | grep "sleep 492"` returns more than one line

 If said test succeeds, we can assume that said snapshot is properly loaded and forget about it, but I also would like to make sure I fully understand this to ensure we're not skipping over anything crucial.

machine.go
```diff
@@ -814,6 +833,10 @@ func (m *Machine) UpdateGuestNetworkInterfaceRateLimit(ctx context.Context, ifac
 
 // attachDrive attaches a secondary block device
 func (m *Machine) attachDrive(ctx context.Context, dev models.Drive) error {
+	if m.Cfg.hasSnapshot() {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
austinvazquez: My concern and something I will need to dive into `LoadSnapshot` to understand further is the two caveats must be guaranteed from a Firecracker user perspective (e.g. firecracker-go-sdk)

1. guest did not mount the device
2. guest does not read or write from the raw block device /dev/vdX during the update sequence

machine.go
```diff
@@ -826,6 +837,10 @@ func (m *Machine) UpdateGuestNetworkInterfaceRateLimit(ctx context.Context, ifac
 
 // attachDrive attaches a secondary block device
 func (m *Machine) attachDrive(ctx context.Context, dev models.Drive) error {
+	if m.Cfg.hasSnapshot() {
+		return nil
+	}
```
austinvazquez, line 842: This check can be removed. The `AttachDrivesHandler` is not ran in the new `loadSnapshotHandlerList`

.gitignore
```diff
@@ -6,7 +6,7 @@ vmlinux
 root-drive.img
 TestPID.img
 build/
-testdata/fc.stamp
```
austinvazquez, line 9: Oops, need to add this back.

Makefile
```diff
@@ -35,9 +36,22 @@ release_url=https://github.com/firecracker-microvm/firecracker/releases/download
...
+$(FC_TEST_BIN_PATH)/ptp \
+$(FC_TEST_BIN_PATH)/host-local \
+$(FC_TEST_BIN_PATH)/static \
+$(FC_TEST_BIN_PATH)/tc-redirect-tap
+
+# Enable pulling of artifacts from S3 instead of building
+# TODO: https://github.com/firecracker-microvm/firecracker-go-sdk/issues/418
+ifeq ($(GID), 0)
+testdata_objects += $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img \
+					$(FC_TEST_DATA_PATH)/root-drive-ssh-key
```
austinvazquez, line 53: The formatting here ended up uglier than I wanted. Perhaps this would be better.

```suggestion
testdata_objects += $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img $(FC_TEST_DATA_PATH)/root-drive-ssh-key
```

machine_test.go
```diff
@@ -1772,6 +1821,261 @@ func TestCreateSnapshot(t *testing.T) {
...
+				err = m.PauseVM(ctx)
+				require.NoError(t, err)
+
+				err = m.CreateSnapshot(ctx, memPath, snapPath)
+				require.NoError(t, err)
+			},
+
+			loadSnapshot: func(ctx context.Context, machineLogger *logrus.Logger, socketPath, memPath, snapPath string) {
+				time.Sleep(10 * time.Second) // Let the system free IP
```
austinvazquez, line 1996: @sondavidb, I did some digging and found a more deterministic way to have this work. So effectively we need to create a temporary cache directory and use it in the network interface configuration for both save and load snapshot.

```go
				networkInterface := NetworkInterface{
					CNIConfiguration: &CNIConfiguration{
						NetworkName: networkName,
						IfName:      ifName,
						ConfDir:     cniConfDir,
						CacheDir: cniCacheDir,
						BinPath:     cniBinPath,
						Args:        [][2]string{{"IP", ipToRestore + networkMask}},
						VMIfName:    "eth0",
					},
				}
```
Then we will need to monitor this directory for the IP to be freed before continuing to load the snapshot.

austinvazquez (changes requested):

Makefile
```diff
@@ -82,6 +96,34 @@ $(FC_TEST_DATA_PATH)/fc.stamp:
 $(FC_TEST_DATA_PATH)/root-drive.img:
 	$(curl) -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/fsfiles/hello-rootfs.ext4
 
+$(FC_TEST_DATA_PATH)/root-drive-with-ssh.img: $(FIRECRACKER_DIR)
+	$(FIRECRACKER_DIR)/tools/devtool build_rootfs
+	cp $(FIRECRACKER_DIR)/build/rootfs/bionic.rootfs.ext4 $@
+
+$(FC_TEST_DATA_PATH)/root-drive-ssh-key: $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img
+# Need root to move ssh key to testdata location
```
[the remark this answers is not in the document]
sondavidb: devtools creates the rootfs's ssh key in a privileged container, so we can't copy the ssh key without root perms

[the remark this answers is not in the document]
ginglis13: Might benefit future readers to permalink to running this privileged container or mention that here https://github.com/firecracker-microvm/firecracker/blob/firecracker-v1.1/tools/devtool#L2188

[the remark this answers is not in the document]
sondavidb: Good idea. Added and will be in the next force push

Makefile
```diff
@@ -82,6 +93,33 @@ $(FC_TEST_DATA_PATH)/fc.stamp:
 $(FC_TEST_DATA_PATH)/root-drive.img:
 	$(curl) -o $@ https://s3.amazonaws.com/spec.ccfc.min/img/hello/fsfiles/hello-rootfs.ext4
 
+$(FC_TEST_DATA_PATH)/root-drive-ssh-key $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img: $(FIRECRACKER_DIR)
+# Need root to move ssh key to testdata location
+ifeq ($(GID), 0)
+	$(FIRECRACKER_DIR)/tools/devtool build_rootfs
+	cp $(FIRECRACKER_DIR)/build/rootfs/bionic.rootfs.ext4 $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img
+	cp $(FIRECRACKER_DIR)/build/rootfs/ssh/id_rsa $(FC_TEST_DATA_PATH)/root-drive-ssh-key
```
austinvazquez, line 101: Because the copies rely on a successful build let's chain them so if a build failure occurs we stop early.

```suggesstion
	$(FIRECRACKER_DIR)/tools/devtool build_rootfs && \
	cp $(FIRECRACKER_DIR)/build/rootfs/bionic.rootfs.ext4 $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img && \
	cp $(FIRECRACKER_DIR)/build/rootfs/ssh/id_rsa $(FC_TEST_DATA_PATH)/root-drive-ssh-key
```

kzys: You don't have to do that in make.

https://www.gnu.org/software/make/manual/html_node/Errors.html

> After each shell invocation returns, make looks at its exit status. If the shell completed successfully (the exit status is zero), the next line in the recipe is executed in a new shell; after the last line is finished, the rule is finished.

.buildkite/pipeline.yml
```diff
@@ -110,7 +113,7 @@ steps:
 
   - label: ':hammer: root tests'
     commands:
-      - "sudo PATH=$PATH FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make test EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
+      - "sudo -E PATH=$PATH:/usr/sbin FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER} FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make test EXTRAGOARGS='-v -count=1 -race' DISABLE_ROOT_TESTS="
```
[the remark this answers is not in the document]
sondavidb: Yes, the dev buildtools uses mkfs.ext4, which is only found in the root user’s PATH, but we are overriding it with the user’s PATH, so we have to append this to the end of it (or figure out an equivalent solution, as I feel this is a tad messy)

machine_test.go
```diff
@@ -1772,6 +1821,327 @@ func TestCreateSnapshot(t *testing.T) {
...
+
+				session, err := client.NewSession()
+				require.NoError(t, err)
+				defer session.Close()
+
+				var b bytes.Buffer
+				session.Stdout = &b
+				err = session.Run(`ps -aux | grep "sleep 422" | wc -l`)
+				require.NoError(t, err)
+				require.Equal(t, "3\n", b.String())
```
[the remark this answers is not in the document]
sondavidb: The third command is what is sent via ssh.
B7C00FE2-5D2F-4F0C-8EDB-C502DFC572D2

.buildkite/pipeline.yml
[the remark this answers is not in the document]
austinvazquez: To be clear, dependencies are being rebuilt in this step which should have been built in install step. Adding `/usr/sbin` here and below in test steps should be considered a PoC and not a long term solution. e.g. we built the rootfs and ssh-key pair 3 times in the passing build [here](https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/2490).

opts.go
```diff
@@ -47,3 +48,14 @@ func WithProcessRunner(cmd *exec.Cmd) Opt {
 		machine.cmd = cmd
 	}
 }
+
+// WithSnapshot will allow for the machine to start using a given snapshot.
+func WithSnapshot(memFilePath, snapshotPath string, opts ...LoadSnapshotOpt) StartOpt {
+	return func(m *Machine) {
+		m.Cfg.Snapshot.MemFilePath = memFilePath
+		m.Cfg.Snapshot.SnapshotPath = snapshotPath
+		m.Cfg.Snapshot.Opts = opts
```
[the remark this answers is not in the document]
sondavidb: We technically could, I just feel it would clutter the code here when it can be done much cleaner in [firecracker.go](https://github.com/sondavidb/firecracker-go-sdk/blob/snapshot-tester/firecracker.go?#L290) and makes it more consistent with other functions in that file. Doing it here would require having the user re-provide the context they're already using for Start as well as require us to import context.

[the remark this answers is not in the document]
kzys: I don't get the context part. Applying options means just mutating m.Cfg.Snapshot (or LoadSnapshotParams).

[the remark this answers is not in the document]
sondavidb: Ah, I see, my bad. I agree with the idea, but this would require the WithSnapshot opt to take in a different option type, i.e. make a new option type in opts.go or machine.go and trash LoadSnapshotOpt in firecracker.go. I feel given all the other functions in firecracker.go it would be inconsistent to have one function that doesn't have or use the opt type. But if you think it makes more sense to handle these options within opts.go, I can certainly do that.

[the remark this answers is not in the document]
kzys: Do you have a function to modify EnableDiffSnapshots? Shouldn't it take SnapshotConfig?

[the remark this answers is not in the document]
sondavidb: We looked into it a bit more, and while this is doable via a new option type, the main caveat is that it removes the user's ability to set some client-side configurations when reaching firecracker.go (e.g. WithContext(), WithTimeout(), etc). However, not being able to change the variables (timeout, Context, and HTTPClient) in the scope of LoadSnapshot sounds fine to me? But I'm not really sure — what do you think? Is this alright?

Makefile
```diff
@@ -35,9 +36,22 @@ release_url=https://github.com/firecracker-microvm/firecracker/releases/download
 testdata_objects = \
 $(FC_TEST_DATA_PATH)/vmlinux \
 $(FC_TEST_DATA_PATH)/root-drive.img \
+$(FC_TEST_DATA_PATH)/root-drive-with-ssh.img \
+$(FC_TEST_DATA_PATH)/root-drive-ssh-key \
 $(FC_TEST_DATA_PATH)/jailer \
 $(FC_TEST_DATA_PATH)/firecracker \
-$(FC_TEST_DATA_PATH)/ltag
+$(FC_TEST_DATA_PATH)/ltag \
+$(FC_TEST_BIN_PATH)/ptp \
```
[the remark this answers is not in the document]
austinvazquez: I looked into this and it didn't work out-of-the-box. Mainly I didn't see the bridge plugin creating the IP route on the host like the ptp plugin does. So packets were not being routed to the bridge. More than likely I am miss configuring though. @sondavidb would you like to give it a shot?

machine_test.go
```diff
@@ -1772,6 +1821,327 @@ func TestCreateSnapshot(t *testing.T) {
...
+				err = m.CreateSnapshot(ctx, memPath, snapPath)
+				require.NoError(t, err)
+			},
+
+			loadSnapshot: func(ctx context.Context, machineLogger *logrus.Logger, socketPath, memPath, snapPath string) {
+				var ipFreed bool = false
+				var err error
+
+				for i := 0; i < maxRetries; i++ {
+					if _, err = os.Stat("/var/lib/cni/networks/fcnet/" + ipToRestore); err == nil {
```
austinvazquez, line 2048: 1. Let's create a method variable `const cniCacheDir = "/var/lib/cni"` around line 1898 with the network name. 
2. Also let's add a comment along the lines of `// Using default cache directory to ensure collision avoidance on IP allocations`
3. Finally here it will look like `os.Stat(fmt.Sprintf("%s/networks/%s/%s", cniCacheDir, networkName, ipToRestore))`

austinvazquez (changes requested):

machine_test.go
```diff
@@ -1467,8 +1470,48 @@ func TestWaitWithNoSocket(t *testing.T) {
...
-	return Config{
+type machineConfigOpt func(c *Config)
+
+func withRootDrive(rootfs string) machineConfigOpt {
+	return func(c *Config) {
+		var drives []models.Drive
+
+		inserted := false
+		for _, drive := range c.Drives {
+			if *drive.DriveID == "root" {
```
kzys, line 1481: How about checking IsRootDevice instead?

kzys (approved):

machine_test.go
```diff
@@ -1772,6 +1821,329 @@ func TestCreateSnapshot(t *testing.T) {
...
+		  }
+		]
+	  }`, networkName, subnet)), 0644)
+}
+
+func TestLoadSnapshot(t *testing.T) {
+	fctesting.RequiresKVM(t)
+	fctesting.RequiresRoot(t)
+
+	dir, err := ioutil.TempDir("", t.Name())
```
ginglis13, line 1876: [T.TempDir()](https://pkg.go.dev/testing#T.TempDir) ?

sondavidb: It seems the entirety of machine_test.go uses ioutil.TempDir as opposed to its testing.T counterpart. So, I think it's fine to leave it as is.

ginglis13 (approved):

sondavidb:
Worth mentioning the install step seems to be flaky — sometimes it doesn't set up the loop device during mounting (e.g. the install step in [this build](https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/2496), line 83), but just restarting the test makes it run fine. Unsure of cause but good to document in my opinion.

austinvazquez (approved):

Makefile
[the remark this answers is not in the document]
Kern--: Both the bridge plugin and ptp create veth pairs, so it should be just a slightly different config to switch, but I don't feel strongly enough about which we use to block on that. It's fine to ship this as is.

Kern-- (approved):

Issue #423, opened 2022-07-16
Document feature support and limitations for microVM snapshot feature.


austinvazquez:
It would be good to have a writeup for users with the following information.

- [x] Current snapshot support from the SDK
- [x] Known limitations

PR #431, opened 2022-07-29
Added snapshot documentation + pipeline example code

sondavidb:

*Issue #, if available:*
#423 

*Description of changes:*
 Documenting changes made in #414 as well as creating a working example of snapshots in the SDK that is tested in the pipeline to ensure relevancy. It may be redundant due to it being very similar to `TestLoadSnapshot()` in `machine.go` but I feel having a prepackaged way for users to easily test snapshotting capabilities is very useful. Feel free to discuss below.


examples/cmd/snapshotting/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Note the user PATH variable is different from the root user's PATH variable, hence the need for `-E env PATH=$PATH`.
+
+Upon running, the VM logs will be printed to the console, as well as the IP of the VM. It will then show that it is sending the marker (in our case, `sleep 422`).
+
+Afterwards, the snapshot is created and the machine is terminated. The snapshot files are saved in the snapshotssh folder created in the directory.
+
+Then, a new machine is created, booted with the snapshot that was just taken, and the IP of the VM will once again be printed to the console (which should be the same as the last machine). The output of searching for the marker (in this case `ps -aux | grep "sleep 422"`) is then printed to the console and the user can confirm that the snapshot loaded properly.
+
+Note: Lines 223 and 349 have `fmt.Scanln()` commented out. Uncommenting them pauses the program, allowing you to ssh into the machine in a new terminal window with the following command:
```
ginglis13, line 32: instead of leaving those commented out / referencing those line numbers (which could be subject to change), could you provide an example? i.e.: "You can pause the execution of the program and ssh into the machine <...etc etc...> by adding a blocking function call to the example after launching the respective VMs like:

```go
ipToRestore = m.Cfg.NetworkInterfaces[0].StaticConfiguration.IPConfiguration.IPAddr.IP.String()
fmt.Printf("IP of VM: %v\n", ipToRestore)
fmt.Scanln() // block, allows you to ssh from another shell

...

fmt.Println("Snapshot loaded")
fmt.Printf("IP of VM: %v\n", ipToRestore)
fmt.Scanln() // block, allows you to ssh from another shell
```

docs/snapshotting.md
```diff
@@ -0,0 +1,58 @@
...
+    ...
+
+}
+m, _ := sdk.NewMachine(ctx, cfg)
+
+m.Start(ctx, sdk.WithSnapshot(memPath, snapPath))
+m.ResumeVM(ctx)
+```
+
+Check out examples/cmd/snapshotting for a quick example that can be run on your machine.
```
ginglis13, line 58: nit: link to examples/cmd/snapshotting

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,393 @@
...
+
+		err = m.ResumeVM(ctx)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		fmt.Println("Snapshot created")
+	}
+
+	var loadSnapshotSSH = func() {
```
ginglis13, line 280: just from glancing at this, I can review more next week - for this and the create snapshot function, is there a reason you chose to define these in `main` rather than as their own functions? I'm guessing maybe to keep some variables in scope?

ginglis13 (changes requested):

examples/cmd/snapshotting/Makefile
```diff
@@ -0,0 +1,67 @@
...
+bin/ptp: bin
+	GO111MODULE=off GOBIN=$(PWD)/bin \
+	go get github.com/containernetworking/plugins/plugins/main/ptp
+
+bin/host-local: bin
+	GO111MODULE=off GOBIN=$(PWD)/bin \
+	go get github.com/containernetworking/plugins/plugins/ipam/host-local
+	
+image:
+	- sudo cp ${FC_TEST_DATA_PATH}/root-drive-with-ssh.img root-drive-with-ssh.img
```
austinvazquez, line 41: This assumes the user is able to escalate root permissions. It is better practice to check permissions and error.

examples/cmd/snapshotting/README.md
austinvazquez, line 32: From a documentation perspective this feels tedious to maintain, but a little more beyond that the commented lines of code are unnecessary.

docs/snapshotting.md
austinvazquez: Also what is cmd in this context?

austinvazquez (changes requested):

examples/cmd/snapshotting/example_demo.go
sondavidb: Made the changes and refactored some to make the functions separate, this ended up with me creating a global variable but I think that should be fine since it's just example code.

examples/cmd/snapshotting/README.md
sondavidb: Per what @ginglis13 suggested I made it more of a general case, it still involves a particular code snippet but should be better? What do you think?

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,410 @@
...
+// Using default cache directory to ensure collision avoidance on IP allocations
+const cniCacheDir = "/var/lib/cni"
+const networkName = "fcnet"
+const ifName = "veth0"
+
+const networkMask string = "/24"
+const subnet string = "10.168.0.0" + networkMask
+
+const maxRetries int = 10
+const backoffTimeMs time.Duration = 500
```
ginglis13, line 43: let's put this all in one block

```go
const(
    cniCacheDir = "/var/lib/cni"
    ...
)
```

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,410 @@
...
+
+var ipToRestore string // Don't use global variables in production!
```
ginglis13, line 45: I'm thinking rather than keeping this as a global var you could have `createSnapshotSSH` return the ip address to restore. Then, in main, you can get this ip from calling `createSnapshotSSH` and pass that into your load snapshot function

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,410 @@
...
+	cniConfDir := filepath.Join(dir, "cni.conf")
+	err = os.Mkdir(cniConfDir, 0777)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(cniConfDir)
+
+	// Setup socket and snapshot + memory paths
+	tempdir, err := ioutil.TempDir("", "FCGoSDKSnapshotExample")
+	defer os.Remove(tempdir)
```
ginglis13, line 394: nit: check this error

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,410 @@
...
+func main() {
+	// Check for kvm and root access
+	err := unix.Access("/dev/kvm", unix.W_OK)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if x, y := 0, os.Getuid(); x != y {
+		log.Fatal("Root acccess denied")
+	}
```
ginglis13, line 378: are these essentially doing the same thing? IIRC `/dev/kvm` is only accessible by root anyhow

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,410 @@
...
+
+	sshKeyPath := filepath.Join(dir, "root-drive-ssh-key")
+
+	var client *ssh.Client
+	for i := 0; i < maxRetries; i++ {
+		client, err = connectToVM(m, sshKeyPath)
+		if err != nil {
+			time.Sleep(backoffTimeMs * time.Millisecond)
+		} else {
+			fmt.Println(i)
```
ginglis13, line 345: any reason to print the number of retries?

examples/cmd/snapshotting/example_demo.go
```diff
@@ -0,0 +1,410 @@
...
+	)
+
+	// Use the firecracker binary
+	cmd := sdk.VMCommandBuilder{}.WithSocketPath(fmt.Sprintf("%s.load", socketPath)).WithBin(filepath.Join(dir, "firecracker")).Build(ctx)
+
+	m, err := sdk.NewMachine(ctx, cfg, sdk.WithProcessRunner(cmd))
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(socketPath + ".load")
```
ginglis13, line 312: nit: can we have a variable `... = fmt.Sprintf("%s.load", socketPath)` and use that 3x ? similar comment for the create snapshot function.

ginglis13 (changes requested):

examples/cmd/snapshotting/example_demo.go
sondavidb: Agreed, but the testing suite checks for both. I suppose it can't hurt to have both to cover any potential edge cases (if they even exist?)

examples/cmd/snapshotting/example_demo.go
sondavidb: Oops, forgot to remove some debug code 🙃

austinvazquez (approved):

docs/snapshotting.md
```diff
@@ -0,0 +1,58 @@
+# Snapshotting
+
+Snapshotting is currently supported in the Firecracker Go SDK using Firecracker v1.0.0's API.
+
+Due to [known issues and limitations](https://github.com/firecracker-microvm/firecracker/blob/firecracker-v1.0/docs/snapshotting/snapshot-support.md#known-issues-and-limitations), it is currently not recommended to use snapshots in production.
+
+Snapshots created in this version only save the following:
+- guest memory
+- emulated hardware state (both KVM & Firecracker eumlated hardware)
```
ginglis13, line 9: nit: eumlated -> emulated

ginglis13 (changes requested):

ginglis13 (approved):

Issue #432, opened 2022-08-09
Loading snapshot should not require kernel image path on machine config


austinvazquez:
When loading a snapshot, if no kernel image path is provided in machine config, then an error occurs.

```
WARN[0000] Failed handler "validate.Cfg": failed to stat kernel image path, "": stat : no such file or directory
```

This is likely the config validation being pedantic / not designed for loading snapshots.

The task here is to 
1) Verify a snapshot can be loaded without providing the kernel image path
2) Create a new validator for loading snapshots

PR #433, opened 2022-08-09
Removed unnecessary requirements for loading snapshots

sondavidb:

*Issue #, if available:* #432

*Description of changes:*
Created changes to the snapshot opt so that LoadSnapshot only validates network config, socket path, drive location, and snapshot paths. These are the only configurations required to load a snapshot.


handlers.go
```diff
@@ -57,6 +58,16 @@ var ConfigValidationHandler = Handler{
 	},
 }
 
+// ConfigValidationHandler is used to validate that required fields are
+// present. This validator is to be used when the jailer is turned off.
```
ginglis13, line 62: nit on the comment: 

1/ LoadSnapshotConfigValidationHandler
2/ This comment mentions the validator is to be used when the jailer is turned off, is that accurate in the case of loading snapshots?

machine_test.go
```diff
@@ -1958,6 +1958,55 @@ func TestLoadSnapshot(t *testing.T) {
...
+					// Rewriting m.cmd partially wouldn't work since Cmd has
+					// some unexported members
+					args := m.cmd.Args[1:]
+					m.cmd = exec.Command(getFirecrackerBinaryPath(), args...)
+				}, WithLogger(logrus.NewEntry(machineLogger)))
+				require.NoError(t, err)
+
+				m.Cfg.KernelImagePath = ""
+				m.Cfg.MachineCfg = models.MachineConfiguration{}
+
```
ginglis13, line 1999: I may be misunderstanding the test. With your changes, should attempting to load a snapshot with KernelImagePath set to "" result in a config validation error?

sondavidb: No. When loading a snapshot, many of the fcinit handlers are removed as much of the setup of a new machine can be ignored. For example, machine config is not needed, as the snapshot already saves this config, as well as the kernel. However, not specifying them previously would cause the validator to fail. This change will make the validator ignore these fields when loading a snapshot.

ginglis13: Ah I see now - nice, LGTM then :)

handlers.go
sondavidb: Oh, I forgot to change this. No, this will just be called whenever LoadSnapshot is called. I'll investigate more on the jailer later, as I'm unsure how the jailer setup interacts with snapshotting.

ginglis13 (approved):

machine.go
```diff
@@ -204,6 +204,37 @@ func (cfg *Config) Validate() error {
 	return nil
 }
 
+func (cfg *Config) LoadSnapshotValidate() error {
```
austinvazquez, line 207: There seems to be slight inconsistencies with if the validation is done as apart of the handler of as a method of `Config`.

At the very least I think the method name should be verb first. e.g. `ValidateLoadSnapshot()`

machine_test.go
```diff
@@ -1958,6 +1958,55 @@ func TestLoadSnapshot(t *testing.T) {
 				require.NoError(t, err)
 			},
 		},
+		{
+			name: "TestLoadSnapshot without unnecessary config",
```
austinvazquez, line 1962: Instead of a new test with minimum config I think it is better for readers to just update the current test to utilize minimum config. Imagine the user who only reads the first test for how to use the feature and gets the impression that additional config is required.

austinvazquez (changes requested):
It would be nice to update the examples as well to only use minimum configuration.

machine.go
sondavidb: I see what you're saying, but I think this would still be the right place to validate these options. The other handlers are in their specified file because the only part of the config it validates is its own field. Validate and LoadSnapshotValidate both require multiple areas of the config to validate. So, I think this would be the right place to put it, and as no other function calls it, I don't really see an issue keeping it as is.

sondavidb:
> It would be nice to update the examples as well to only use minimum configuration.

Correct. I made a change to the example's `go.mod` so that the local installation of the SDK can be used in the example code.

examples/cmd/snapshotting/example_demo.go
```diff
@@ -301,6 +301,8 @@ func loadSnapshotSSH(ctx context.Context, socketPath, memPath, snapPath, ipToRes
 
 	socketFile := fmt.Sprintf("%s.load", socketPath)
 	cfg := createNewConfig(socketFile, withNetworkInterface(networkInterface))
+	cfg.KernelImagePath = ""
+	cfg.MachineCfg = models.MachineConfiguration{}
```
austinvazquez, line 305: This looks weird. Perhaps making the kernel image path and machine config an option would be more clear.

machine_test.go
```diff
@@ -2083,6 +2095,9 @@ func TestLoadSnapshot(t *testing.T) {
 					withNetworkInterface(networkInterface),
 				)
 
+				cfg.MachineCfg = models.MachineConfiguration{}
+				cfg.KernelImagePath = ""
```
austinvazquez, line 2099: Same here. Maybe consider moving to options.

machine.go
```diff
@@ -204,6 +204,37 @@ func (cfg *Config) Validate() error {
...
+	for _, drive := range cfg.Drives {
+		if BoolValue(drive.IsRootDevice) {
+			rootPath := StringValue(drive.PathOnHost)
+			if _, err := os.Stat(rootPath); err != nil {
+				return fmt.Errorf("failed to stat host drive path, %q: %v", rootPath, err)
+			}
+
+			break
+		}
+	}
```
austinvazquez, line 221: It is interesting we validate any drives the user passes in via configuration but none of the handlers which run in a load snapshot do anything with the drives.

examples/cmd/snapshotting/example_demo.go
austinvazquez, line 305: Perhaps consider making these options.

machine_test.go
```diff
@@ -1939,7 +1939,19 @@ func TestLoadSnapshot(t *testing.T) {
...
+				cfg := Config{
+					SocketPath: socketPath + ".load",
+					Drives: []models.Drive{
+						{
+							DriveID:      String("root"),
+							IsRootDevice: Bool(true),
+							IsReadOnly:   Bool(true),
+							PathOnHost:   String(testRootfs),
+						},
+					},
```
austinvazquez, line 1952: Do we need to specify root drive here? Which handler requires this?

machine.go
sondavidb: None of them do, but if the drives are not in place, the LoadSnapshot call will fail, so I think it is right to validate this.

examples/cmd/snapshotting/example_demo.go
sondavidb: I just made a config in place that has those empty, as LoadSnapshot is the only function that would use such an option.

machine_test.go
sondavidb: I feel it would be best to keep them here just so that `ValidateLoadSnapshot()` can ensure that it's in place, but leaving it out still makes it pass, so you're technically correct. I would like to leave this here because leaving out the drives would not allow the user to interact with it until it adds said drives. What do you think?

examples/cmd/snapshotting/example_demo.go
austinvazquez: Can we remove these two lines then?

machine_test.go
```diff
@@ -2078,10 +2090,24 @@ func TestLoadSnapshot(t *testing.T) {
...
-				cfg := createValidConfig(t, fmt.Sprintf("%s.load", socketPath),
-					withRootDrive(rootfsPath),
-					withNetworkInterface(networkInterface),
-				)
+
+				cfg := Config{
+					SocketPath: socketPath + ".load",
+					// Specifying drives in the cfg is not strictly necessary,
+					// but it's nice to have the validator ensure that
+					// the drives are still in place
```
austinvazquez, line 2098: Can we remove this comment? Feels slightly cheeky.

austinvazquez (approved):

PR #434, opened 2022-08-10
Refactored NewMachine and Start functions

sondavidb:

*Issue #, if available:*

*Description of changes:*
Firecracker can only start a VM once per instance. Because of this, it made sense to me to keep all configuration changes to the `NewMachine` call instead of the `Start` call, as `Start` should simply do that — start the VM. Everything should be in place when the machine starts, so there should be no reason to have an option in said call.

As a result of these changes, `WithSnapshot` is now an option used in `NewMachine`. This also required me to refactor the machine options, putting the options changes at the end of the `NewMachine` call, as putting it where it originally was would not allow me to change the config. This should also allow for more flexibility should any future functions want to change machine configuration via `NewMachine`.

Documentation has been updated to reflect this new change.

This should not change any functionality and is simply a design change that I feel is more streamlined and easier to understand, but if this would change any functionality and/or make anything more confusing, feel free to let me know.

### DIFF
--- a/opts.go
+++ b/opts.go
@@ -21,7 +21,6 @@ import (
 
 // Opt represents a functional option to help modify functionality of a Machine.
 type Opt func(*Machine)
-type StartOpt func(*Machine)
 
 // WithClient will use the client in place rather than the client constructed
 // during bootstrapping of the machine. This option is useful for mocking out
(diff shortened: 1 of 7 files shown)


opts.go
```diff
@@ -54,7 +53,7 @@ func WithProcessRunner(cmd *exec.Cmd) Opt {
 type WithSnapshotOpt func(*SnapshotConfig)
```
ginglis13, line 53: not necessarily related to these changes, but is this needed? I don't see it being used anywhere

ginglis13 (approved):

opts.go
sondavidb: Not in testing, but if you wanted to change an option (e.g. whether or not you want to resume the VM on snapshot load) that gives you the option to do so.

austinvazquez (changes requested):
I don't agree with the examples being pinned to a previous commit. This will be difficult to maintain. e.g. right now the API is changing and no automation is saying the examples should be updated.

opts.go
austinvazquez: Maybe it would be nice to have a test use this. Perhaps we can have the SSH test use the resume option.

austinvazquez (approved):

opts.go
ginglis13: +1, @sondavidb can you create an issue that describes adding this test so we can track it?

PR #439, opened 2022-08-21
Added mount directory to build_rootfs command

sondavidb:
Signed-off-by: David Son <son.david.b@gmail.com>

*Issue #, if available:*

*Description of changes:*
Firecracker's `build_rootfs` command now supports [specifying a mount directory](https://github.com/firecracker-microvm/firecracker/pull/3086) via the `-m` flag. This should (hopefully) make the pipeline a little less flaky if this is indeed the reason for the command to seemingly randomly fail.

It's worth nothing that if this is the real issue (which it seems to not be), this means we can remove the pipeline concurrency group. I think I shared some info with the team as to what behaviors are shown before the loop device setup fails (specifically, when creating the working directory in the first pipeline step, the ones that fail to setup the loop device seem to have a different setup process than the ones that pass), so maybe a deeper dive into that would help fix pipeline flakiness. Still, this is nice to have.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ $(FC_TEST_DATA_PATH)/root-drive-ssh-key $(FC_TEST_DATA_PATH)/root-drive-with-ssh
 # Need root to move ssh key to testdata location
 ifeq ($(GID), 0)
 	$(MAKE) $(FIRECRACKER_DIR)
-	$(FIRECRACKER_DIR)/tools/devtool build_rootfs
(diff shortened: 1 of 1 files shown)


ginglis13:
Thanks @sondavidb 👍 Can you rebase against main now that #445 has been merged? Buildkite is failing the install step because of this

sondavidb:
@ginglis13 All good I think, had some issues rebasing but I think I did it

kzys (approved):

ginglis13 (approved):

PR #440, opened 2022-08-26
Remove io/ioutil since it has been deprecated

kzys:
The package has been deprecated since 1.16.

Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #445, opened 2022-08-30
Pin version of ltag for go <= v1.15 compatibility

Joffref:
Signed-off-by: Mathis Joffre <mariusjoffre@gmail.com>

*Issue #, if available:* #364

Sorry for multiple PR opening I had some troubles with my branches and squashing. This the final PR for this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,8 +122,8 @@ $(FC_TEST_BIN_PATH)/tc-redirect-tap:
 	go get github.com/awslabs/tc-redirect-tap/cmd/tc-redirect-tap
 
 $(FC_TEST_DATA_PATH)/ltag:
-	GO111MODULE=off GOBIN=$(abspath $(FC_TEST_DATA_PATH)) \
-	go get github.com/kunalkushwaha/ltag
+	GO111MODULE=on GOBIN=$(abspath $(FC_TEST_BIN_PATH)) \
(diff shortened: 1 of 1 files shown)


Kern-- (commented):
Thanks for this!

If v0.2.3 still works with newer versions of go, I would prefer we just pin it for all go versions. We're already looking at removing support for <=1.15 in https://github.com/firecracker-microvm/firecracker-go-sdk/pull/440 at which point we can update the pin.

Joffref:
I think it is the best option I will made changes, plus we should consider to pin a version for each dependency to avoid situation like this one.

Makefile
```diff
@@ -144,6 +144,6 @@ firecracker-clean:
 
 lint: deps
 	gofmt -s -l .
-	$(FC_TEST_DATA_PATH)/ltag -check -v -t .headers
+	$(FC_TEST_DATA_PATH)/bin/ltag -check -v -t .headers
```
ginglis13, line 147:
```suggestion
	$(FC_TEST_BIN_PATH)/ltag -check -v -t .headers
```

ginglis13 (approved):
small nit, otherwise LGTM

Kern-- (approved):

kzys (approved):

PR #456, opened 2022-10-05
Removing CPU template from Config for non-intel instance

vvejell1:
Signed-off-by: Vaishnavi Vejella <vvejella@amazon.com>

*Issue:* While testing **AMD** instance against the Firecracker Go SDK, the BuildKite's pipeline tests are getting failed due to CPUTemplate which is not supported by non-intel instances.
*Description of changes:* Added two files [cpu_template.go](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/internal/cpu_template.go 
), [cpu_template_test.go ](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/internal/cpu_template_test.go)by taking reference from firecracker-containerd to determine that CPU template is only supported by Intel. 

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -51,7 +51,6 @@ func createMachine(ctx context.Context, name string, forwardSignals []os.Signal)
 		LogLevel:        "Info",
 		MachineCfg: models.MachineConfiguration{
 			VcpuCount:   Int64(1),
-			CPUTemplate: models.CPUTemplate(models.CPUTemplateT2),
 			MemSizeMib:  Int64(256),
 			Smt:         Bool(false),
 		},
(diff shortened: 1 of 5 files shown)


client/models/machine_configuration.go
```diff
@@ -30,9 +30,6 @@ import (
 // swagger:model MachineConfiguration
 type MachineConfiguration struct {
 
-	// cpu template
```
ginglis13, line 33: I think we still want to keep the `CPUTemplate` as an option here to allows those that do wish to specify this value to do so. same goes for `Validate` and `validateCPUTemplate` functions

ginglis13 (changes requested):
some minor changes requested but otherwise LGTM, thanks!

Could you update your commit message to have the body be descriptive like the description you've provided in this PR?

kzys:
Can you add one test specifically for CPUTemplate? We can disable the test on non-Intel platforms.

After merging this change, even if we break CPUTemplate, we wouldn't notice.

kzys (approved):

kzys:
@vvejell1 Can you change your name on the commit? You would need something like

```
% git config user.name 'Vaishnavi Vejella' 
% git commit --amend --reset-author 
```

to rewrite the name.

PR #458, opened 2022-10-19
Fix goroutine leak in startVMM

bduffany:
*Issue #, if available:*

N/A

*Description of changes:*

The statement `<-ctx.Done()` without a `select {}` wrapping it will cause a goroutine leak if `ctx` is `context.Background()` or `context.TODO()` (or any other context that doesn't get cancelled, intentionally or not).

This change allows the goroutine to exit if the VM exits on its own, in addition to waiting for `ctx.Done()`, preventing a goroutine leak in case people do pass in a context that doesn't get canceled, but are otherwise managing the VM lifecycle correctly.

---

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -613,10 +613,16 @@ func (m *Machine) startVMM(ctx context.Context) error {
 		return err
 	}
 
-	// This goroutine is used to kill the process by context cancelletion,
+	// This goroutine is used to kill the process by context cancellation,
 	// but doesn't tell anyone about that.
 	go func() {
-		<-ctx.Done()
+		select {
+		case <-ctx.Done():
+			break
+		case <-m.exitCh:
+			// VMM exited on its own; no need to stop it.
+			return
+		}
 		err := m.stopVMM()
 		if err != nil {
 			m.logger.WithError(err).Errorf("failed to stop vm %q", m.Cfg.VMID)


machine.go
```diff
@@ -613,10 +613,16 @@ func (m *Machine) startVMM(ctx context.Context) error {
...
 
-	// This goroutine is used to kill the process by context cancelletion,
+	// This goroutine is used to kill the process by context cancellation,
 	// but doesn't tell anyone about that.
 	go func() {
-		<-ctx.Done()
+		select {
+		case <-ctx.Done():
+			break
+		case <-m.cleanupCh:
```
Kern--, line 622: nitpick: `m.cleanupCh` works, but to to make it clear that what we really care about is whether the VM exited, can you use `m.exitCh`?

Kern--:
Thanks, just one minor comment :)

Kern--:
Actually, one more, can you please add a signed off line to your commit message like:

```
Signed-off-by: Firstname Lastname <email@provider.com>
```

machine.go
bduffany: Good call; done

bduffany:
> can you please add a signed off line to your commit message

Done

bduffany:
I'm not sure how to interpret the failing CI tests; it looks like maybe TestLoadSnapshot timed out. Is that a known issue @Kern-- or is it likely this PR could've caused it?

Kern-- (approved):

austinvazquez (approved):

PR #469, opened 2022-12-14
Adding requirements to support ARM instance

vvejell1:
Signed-off-by: Vaishnavi Vejella <vvejella@amazon.com>

*Issue:*
While trying to perform ARM integration test against firecracker-go-sdk, found that **SendCtrlAltDel** is not supported on aarch64 and **TestNetworkMachineCNIWithConfFile** test is failing because **testPing** is not supported when we are running the test on multiple VM's.

*Description of changes:*
If the architecture is **ARM64**, we are not able to send **SendCtrlAltDel** (API) on aarch64, so stopping the machine using `StopVMM()` and also skipping the test by not passing the VM's then **TestNetworkMachineCNIWithConfFile** will get passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,8 @@ DISABLE_ROOT_TESTS?=1
 DOCKER_IMAGE_TAG?=latest
 EXTRAGOARGS:=
 FIRECRACKER_DIR=build/firecracker
-FIRECRACKER_TARGET?=x86_64-unknown-linux-musl
+arch=$(shell uname -m)
+FIRECRACKER_TARGET?=$(arch)-unknown-linux-musl
 
 FC_TEST_DATA_PATH?=testdata
 FC_TEST_BIN_PATH:=$(FC_TEST_DATA_PATH)/bin
@@ -28,7 +29,6 @@ UID = $(shell id -u)
 GID = $(shell id -g)
 
 firecracker_version=v1.0.0
-arch=$(shell uname -m)
 
 # The below files are needed and can be downloaded from the internet
 release_url=https://github.com/firecracker-microvm/firecracker/releases/download/$(firecracker_version)/firecracker-$(firecracker_version)-$(arch).tgz
--- a/machine.go
+++ b/machine.go
@@ -24,6 +24,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -64,6 +65,9 @@ type SeccompConfig struct {
 // ErrAlreadyStarted signifies that the Machine has already started and cannot
 // be started again.
 var ErrAlreadyStarted = errors.New("firecracker: machine already started")
+
+// ErrGraceShutdown signifies that the Machine will shutdown gracefully and SendCtrlAltDelete is unable to send
+//var ErrGraceShutdown = errors.New("Shutdown gracefully: SendCtrlAltDelete is not supported if the arch is ARM64")
 
 type MMDSVersion string
 
@@ -456,7 +460,11 @@ func (m *Machine) Start(ctx context.Context) error {
 // Shutdown requests a clean shutdown of the VM by sending CtrlAltDelete on the virtual keyboard
 func (m *Machine) Shutdown(ctx context.Context) error {
 	m.logger.Debug("Called machine.Shutdown()")
-	return m.sendCtrlAltDel(ctx)
+	if runtime.GOARCH != "arm64" {
+		return m.sendCtrlAltDel(ctx)
+	} else {
+		return m.StopVMM()
+	}
 }
 
 // Wait will wait until the firecracker process has finished.  Wait is safe to
--- a/network_test.go
+++ b/network_test.go
@@ -20,6 +20,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
@@ -253,7 +254,7 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 	}
 	fctesting.RequiresRoot(t)
 
-	cniBinPath := []string{"/opt/cni/bin", testDataBin}
+	cniBinPath := []string{testDataBin, "/opt/cni/bin"}
 
 	dir, err := ioutil.TempDir("", fsSafeTestName.Replace(t.Name()))
 	require.NoError(t, err)
@@ -312,6 +313,9 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 		require.NoError(t, err, "cni conf should parse")
 	}
 
+	if runtime.GOARCH == "arm64" {
+		return
+	}
 	numVMs := 10
 	vmIPs := make(chan string, numVMs)
 
@@ -405,7 +409,6 @@ func newCNIMachine(t *testing.T,
 		MachineCfg: models.MachineConfiguration{
 			VcpuCount:  Int64(2),
 			MemSizeMib: Int64(256),
-			Smt:        Bool(true),
 		},
 		Drives: []models.Drive{
 			{
--- a/.hack/go.mod
+++ b/.hack/go.mod
@@ -12,4 +12,5 @@ go 1.11
 require (
 	github.com/awslabs/tc-redirect-tap v0.0.0-20220715050423-f2af44521093 // indirect
 	github.com/containernetworking/plugins v1.1.1 // indirect
+	github.com/kunalkushwaha/ltag v0.2.3 // indirect
 )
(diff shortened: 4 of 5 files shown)


network_test.go
```diff
@@ -253,7 +254,7 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 	}
 	fctesting.RequiresRoot(t)
 
-	cniBinPath := []string{"/opt/cni/bin", testDataBin}
+	cniBinPath := []string{"/testdata/bin", testDataBin}
```
ginglis13, line 257:
```suggestion
	cniBinPath := []string{testDataBin, "/opt/cni/bin"}
```

`testDataBin` is defined here and should provide the path to these CNI plugins built for ARM: https://github.com/firecracker-microvm/firecracker-go-sdk/blob/4648d28a3cc3018cf576cc8ae3cf696cb4c0dbc4/machine_test.go#L71

I don't think we can guarantee "/testdata" exists.

network_test.go
```diff
@@ -344,7 +345,9 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 
 			assert.FileExists(t, expectedCacheDirPath, "CNI cache dir doesn't exist after vm startup")
 
-			testPing(t, vmIP, 3, 5*time.Second)
+			if runtime.GOARCH != "arm64" {
```
ginglis13, line 348: worth checking for `arm` as well? 

https://pkg.go.dev/internal/goarch#GOARCH

Makefile
```diff
@@ -75,7 +75,7 @@ unit-tests: $(testdata_objects)
 	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test -short ./... $(EXTRAGOARGS)
 
 all-tests: $(testdata_objects)
-	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test ./... $(EXTRAGOARGS)
+	DISABLE_ROOT_TESTS=$(DISABLE_ROOT_TESTS) go test -v ./... $(EXTRAGOARGS)
```
ginglis13, line 78: nit: callers of `make all-tests` sholud set `-v` in `EXTRAGOARGS`

network_test.go
vvejell1: I have tried by printing `runtime.GOARCH` to check if the arch is ARM64

Makefile
vvejell1: Noted and removed `-v` in `EXTRAGOARGS`

machine.go
```diff
@@ -456,7 +460,10 @@ func (m *Machine) Start(ctx context.Context) error {
 // Shutdown requests a clean shutdown of the VM by sending CtrlAltDelete on the virtual keyboard
 func (m *Machine) Shutdown(ctx context.Context) error {
```
kzys, line 461: Doesn't Firecracker return an error? Does the SDK need to know that?

.buildkite/pipeline.yml
```diff
@@ -108,8 +108,8 @@ steps:
   
   - label: ':book: examples'
     commands:
-      - "sudo -E PATH=$PATH FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make -C examples/cmd/snapshotting run"
-      - "sudo -E PATH=$PATH FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make -C examples/cmd/snapshotting clean"
+      - "sudo -E PATH=$FC_TEST_DATA_PATH:$PATH FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make -C examples/cmd/snapshotting run"
```
kzys, line 111: Hmm, why?

network_test.go
kzys: Unlike sending Ctrl + Alt + Del, networking should work on aarch64. Do you know the root cause?

kzys:
Your PR description is good. Please put some of that in the commit message itself.

network_test.go
vvejell1: TestPing() is giving internet gateway error such as "Destination Host Unreachable" and not able to ping multiple VM's at a time.

.buildkite/pipeline.yml
vvejell1: To get Arm version binaries like Jailer and Firecracker which are located in TEST_DATA_PATH and added it to PATH

network_test.go
vvejell1: Previously CNI plugins are set towards `/opt/cni/bin` which consists of binaries in X86 format. But when we require the  ARM related binaries and they exists in `testdata/bin`.

ginglis13: Yes, we should put the testdata binaries on PATH before "/opt/cni/bin"; `testDataBin` gives us a reference to the path to testdata binaries. "/testdata/bin" is not guranteed to exist on all setups so we should use the `testDataBin` variable https://github.com/firecracker-microvm/firecracker-go-sdk/blob/8378e6a6b128ad730215c58ea00c9da2f26df87a/machine_test.go#L69-L71

vvejell1: Changed the variable to `testDataBin`

network_test.go
```diff
@@ -253,7 +254,7 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 	}
 	fctesting.RequiresRoot(t)
 
-	cniBinPath := []string{"/opt/cni/bin", testDataBin}
+	cniBinPath := []string{testDataBin, "/opt/cni/bin"}
```
kzys, line 257: Does /opt/cni/bin have x86 binaries in our BuildKite host? Or is it empty?

network_test.go
```diff
@@ -312,63 +313,65 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 		require.NoError(t, err, "cni conf should parse")
 	}
 
-	numVMs := 10
-	vmIPs := make(chan string, numVMs)
+	if runtime.GOARCH != "arm64" {
```
kzys, line 316: - Let's use t.Skip instead
- If possible, we should skip this test entirely. It doesn't make much sense to setup the VM, if we don't run any validations against the VM.

vvejell1: sure, let me try in that case.

network_test.go
vvejell1: I think, x86 CNI plugin binaries should be present under `/opt/cni/bin`

kzys: Can you check the host? It is odd if we have x86 CNI plugins on our ARM host.

vvejell1: I checked and not able to find any PATH related to  `/opt/cni/bin`. Previously we changed the PATH from `/opt/cni/bin` which is not there and set the PATH to `./testdata/bin` where we installed arm related Binaries.

network_test.go
vvejell1: After using t.skip() here, **TestMicroVMExecution** test got failed in the BuildKite because we are skipping and not allowing the VM's to perform the test.
https://buildkite.com/firecracker-microvm/vvejella-firecracker-go-sdk-arm/builds/27#01852ca2-9f4c-41aa-b2a4-0d6e834c6399:~:text=FAIL%3A-,TestMicroVMExecution,-(30.03s)

kzys: I don't think they are related. The VM here and TestMicroVMExecution's VM are different.

vvejell1: May be it is causing flaky test and now I test is fine

network_test.go
kzys: You meant, we don't have the directory? If so, we don't have to have this change to specifically prioritize `testdata/bin` over `/opt/cni/bin`.

network_test.go
```diff
@@ -312,63 +313,67 @@ func testNetworkMachineCNI(t *testing.T, useConfFile bool) {
 		require.NoError(t, err, "cni conf should parse")
 	}
 
-	numVMs := 10
-	vmIPs := make(chan string, numVMs)
+	if runtime.GOARCH == "arm64" {
+		t.Skip()
```
kzys, line 317: Let's return from here. So that we don't have to nest the lines below.

https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_return_early_rather_than_nesting_deeply

machine.go
kzys: I still want to remove this check if possible. While we know Firecracker cannot send Ctrl + Alt + Del on aarch64 microVMs, it would be better to let Firecracker make a call, rather than doing that in SDK.

network_test.go
vvejell1: Can we just keep like this `cniBinPath := []string{testDataBin}`

kzys: It may work, but let's do that in another PR.

machine.go
vvejell1: ok, let me create an Issue for this check in Firecracker repository.

network_test.go
vvejell1: Changed the skip by replacing it with return without nesting the lines below.

network_test.go
ginglis13: the `test` rule of the Makefile has `testdata_objects` as a dependency, which includes the CNI plugins https://github.com/firecracker-microvm/firecracker-go-sdk/blob/49c83706e3a8993212165c18be4b22c458193800/Makefile#L36-L45

Those plugins will be installed to `testDataBin` regardless of platform, so just `cniBinPath := []string{testDataBin}` should be fine for the sake of testing.

ginglis13 (approved):
LGTM

Issue #418, opened 2022-07-06
Intermittent LoadSnapshot integration test and examples failures due to bad rootfs.


austinvazquez:
For testing `CreateSnapshot` and `LoadSnapshot` support in the SDK, we started utilizing `firecracker`'s devtool to create a microVM rootfs and SSH key for spinning up microVMs we could run commands inside using SSH. 

It would be nice to have these artifacts publicly available via S3 bucket for download compared to building it at runtime.

austinvazquez:
The load snapshot integration test can sometimes fail with the following:

```
=== CONT  TestLoadSnapshot

    machine_test.go:2134: 

        	Error Trace:	/local/home/buildkite-agent/builds/ip-10-0-0-177-12/firecracker-microvm/firecracker-go-sdk/machine_test.go:2134

        	            				/local/home/buildkite-agent/builds/ip-10-0-0-177-12/firecracker-microvm/firecracker-go-sdk/machine_test.go:2260

        	Error:      	Received unexpected error:

        	            	dial tcp 10.168.0.3:22: connect: no route to host

        	Test:       	TestLoadSnapshot

time="2022-08-31T15:03:28Z" level=info msg="Sent instance shutdown request: [PUT /actions][204] createSyncActionNoContent "
```

This is a result of a bug in Firecracker's devtool causing the rootfs build to silently fail.

fangn2:
As #488 merged, this issue can be closed as resolved.

PR #488, opened 2023-03-16
Fix CI build by using pre-built rootfs image

fangn2:
*Issue #, if available:*
#418 
The main branch is failing due to hardcoded rootfs name as the name changed with upstream firecracker moving from bionic to jammy.  Also devtool used to build rootfs is unstable. 
```
cp: cannot stat 'build/firecracker/build/rootfs/bionic.rootfs.ext4': No such file or directory
```

*Description of changes:*
1. Not specify the ubuntu version name but using regex to match it. The approach is safe since we will only have one rootfs built at a time.
2. Use prebuilt rootfs and ssh key from S3 instead of building from devtool on fly which seems pretty unstable.
3.  Update readme of snapshotting example to reflect current status. 
4.  Minor update for the snapshotting example code to clean up generated files. 

Signed-off-by: Tony Fang <nhfang@amazon.com>

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -61,11 +61,9 @@ steps:
       # Copy vmlinux and root-drive.img.
       - ln -s /var/lib/fc-ci/vmlinux.bin ${FC_TEST_DATA_PATH}/vmlinux
       - ln -s /var/lib/fc-ci/rootfs.ext4 ${FC_TEST_DATA_PATH}/root-drive.img
-      # Download Firecracker and its jailer.
-      - make deps
-      # Build a rootfs with SSH enabled.
-      - sudo -E FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make ${FC_TEST_DATA_PATH}/root-drive-ssh-key
-      - sudo chown $USER ${FC_TEST_DATA_PATH}/root-drive-ssh-key ${FC_TEST_DATA_PATH}/root-drive-with-ssh.img
+      # Download Firecracker and its jailer
+      # Download pre-built rootfs with SSH enabled and its ssh key.
+      - make deps  
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
@@ -90,6 +88,7 @@ steps:
   
   - label: ':book: examples'
     commands:
+      # Copy those deps from install step to save time re-downloading
       - cp ${FC_TEST_DATA_PATH}/root-drive-ssh-key ${FC_TEST_DATA_PATH}/root-drive-with-ssh.img ${FC_TEST_DATA_PATH}/vmlinux examples/cmd/snapshotting
       - cd examples/cmd/snapshotting
       - make all
@@ -112,7 +111,7 @@ steps:
     commands:
       - export FC_TEST_TAP=fc-root-tap${BUILDKITE_BUILD_NUMBER}
       - export FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH}
-      - make test EXTRAGOARGS="-exec 'sudo -E' -count=1 -race" DISABLE_ROOT_TESTS=
+      - make test EXTRAGOARGS="-exec 'sudo -E' -v -count=1 -race" DISABLE_ROOT_TESTS=
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
       distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
--- a/Makefile
+++ b/Makefile
@@ -34,25 +34,21 @@ firecracker_version=v1.0.0
 release_url=https://github.com/firecracker-microvm/firecracker/releases/download/$(firecracker_version)/firecracker-$(firecracker_version)-$(arch).tgz
 
 testdata_objects = \
-$(FC_TEST_DATA_PATH)/vmlinux \
-$(FC_TEST_DATA_PATH)/root-drive.img \
-$(FC_TEST_DATA_PATH)/jailer \
 $(FC_TEST_DATA_PATH)/firecracker \
+$(FC_TEST_BIN_PATH)/host-local \
+$(FC_TEST_DATA_PATH)/jailer \
 $(FC_TEST_DATA_PATH)/ltag \
 $(FC_TEST_BIN_PATH)/ptp \
-$(FC_TEST_BIN_PATH)/host-local \
+$(FC_TEST_DATA_PATH)/root-drive.img \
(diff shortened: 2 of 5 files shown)


ginglis13 (approved):

Makefile
```diff
@@ -110,7 +110,7 @@ $(FC_TEST_DATA_PATH)/root-drive-ssh-key $(FC_TEST_DATA_PATH)/root-drive-with-ssh
 ifeq ($(GID), 0)
 	$(MAKE) $(FIRECRACKER_DIR)
 	$(FIRECRACKER_DIR)/tools/devtool build_rootfs -m $(FC_TEST_DATA_PATH)/mnt
-	cp $(FIRECRACKER_DIR)/build/rootfs/bionic.rootfs.ext4 $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img
+	cp $(FIRECRACKER_DIR)/build/rootfs/*.rootfs.ext4 $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img
```
swagatbora90, line 113: This looks good for now. We can also make a change to `/tools/devtool` to build the rootfs drive with a custom name provided as an argument.

swagatbora90 (approved):
LGTM. The buildkite go-sdk and go-sdk-amd CI steps  have failed. PTAL before merging.

Makefile
fangn2: The tricky part is the devtool is owned by [firecracker](https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool#L1160). We need to request changes upstream to do that.

examples/cmd/snapshotting/Makefile
```diff
@@ -50,28 +50,27 @@ bin/host-local: bin
...
+# https://github.com/firecracker-microvm/firecracker/blob/main/docs/getting-started.md
 root-drive-with-ssh.img root-drive-ssh-key:
-	- mkdir temp
-	- git clone https://github.com/firecracker-microvm/firecracker temp
-	temp/tools/devtool build_rootfs
-	cp temp/build/rootfs/bionic.rootfs.ext4 root-drive-with-ssh.img
-	cp temp/build/rootfs/ssh/id_rsa root-drive-ssh-key
-	rm -rf temp
-
+	curl -L -o root-drive-with-ssh.img https://s3.amazonaws.com/spec.ccfc.min/ci-artifacts/disks/${ARCH}/ubuntu-18.04.ext4
```
swagatbora90, line 63: `devtool` is using ubuntu 22.04, correct? We should keep the same upgraded version here imo.

fangn2: Current HEAD of firecracker uses 22.04. Older versions of firecracker use 18.04.
Current issue is firecracker didn't publish `ubuntu-18.04.ext4` in the same S3 bucket. 
I can follow up with them and address this in a follow-up PR?

.buildkite/pipeline.yml
```diff
@@ -61,11 +61,9 @@ steps:
       # Copy vmlinux and root-drive.img.
       - ln -s /var/lib/fc-ci/vmlinux.bin ${FC_TEST_DATA_PATH}/vmlinux
       - ln -s /var/lib/fc-ci/rootfs.ext4 ${FC_TEST_DATA_PATH}/root-drive.img
-      # Download Firecracker and its jailer.
-      - make deps
-      # Build a rootfs with SSH enabled.
-      - sudo -E FC_TEST_DATA_PATH=${FC_TEST_DATA_PATH} make ${FC_TEST_DATA_PATH}/root-drive-ssh-key
```
fangn2, line 67: With pre-built rootfs, we no longer need to build them with root, I added them as part of deps build.

Makefile
fangn2: Anyway,  decided to bypass the devtool which seems unstable, and uses pre-built rootfs instead following suggestion from #418.

Makefile
```diff
@@ -132,12 +125,14 @@ $(FC_TEST_BIN_PATH)/tc-redirect-tap:
 $(FC_TEST_DATA_PATH)/ltag:
 	$(call install_go,github.com/kunalkushwaha/ltag,v0.2.3)
 
-$(FIRECRACKER_DIR):
-	- git clone https://github.com/firecracker-microvm/firecracker.git $(FIRECRACKER_DIR)
```
fangn2, line 136: Move this block down to put related code together for better readability.

fangn2:
Updated few other parts(please check Description of changes for details) to pass the build.  
Cleaned up the examples code base a little bit. 

CI seems a little bit unstable when it ran step `test against Firecracker above` which uses firecracker built from HEAD of main, but I think we can monitor from there. 

@swagatbora90  @austinvazquez  @ginglis13  PTAL

ginglis13 (approved):

Makefile
```diff
@@ -42,17 +42,13 @@ $(FC_TEST_DATA_PATH)/ltag \
...
-$(FC_TEST_BIN_PATH)/tc-redirect-tap
-
-# Enable pulling of artifacts from S3 instead of building
-# TODO: https://github.com/firecracker-microvm/firecracker-go-sdk/issues/418
-ifeq ($(GID), 0)
-testdata_objects += $(FC_TEST_DATA_PATH)/root-drive-with-ssh.img $(FC_TEST_DATA_PATH)/root-drive-ssh-key
-endif
+$(FC_TEST_BIN_PATH)/tc-redirect-tap \
+$(FC_TEST_DATA_PATH)/root-drive-with-ssh.img \
+$(FC_TEST_DATA_PATH)/root-drive-ssh-key
```
swagatbora90, line 47: nit: alphabetize the order of test object. the new targets moved after L41?

examples/cmd/snapshotting/README.md
```diff
@@ -62,20 +62,5 @@ Pressing enter resumes execution of the program.
...
-```
-umount: /firecracker/build/rootfs/mnt: not mounted.
-```
-
-This is due to an issue in Firecracker's devtool command used to dynamically create an image. Fixing this is often as simple as rerunning the command.
-
-```
-sudo rm -rf root-drive-with-ssh.img root-drive-ssh-key
-sudo make image
+make clean
```
swagatbora90, line 65: what was changed here that we no longer need sudo?

swagatbora90 (approved):
Minor nits, otherwise looks good.

We are using `hello-rootfs.ext4` for non-ssh testing, which I believe is also uploaded by firecracker team. Lets confirm what OS version this is based of and whether FC team can do something similar for ssh version, i.e. with OS agnostic naming.

examples/cmd/snapshotting/README.md
fangn2: Devtool used to build rootfs uses docker --privileged, then generated rootfs file are under root, needs sudo to copy to test location. 
The new approach directly download from S3 no longer needs that.

Makefile
fangn2: Updated.

PR #495, opened 2023-07-01
Add support for snapshot memory backend

bduffany:
- Pull in upstream swagger definitions for SnapshotLoadParams and MemoryBackend
- Run `go generate`
- Update Snapshot config to allow setting MemBackend
- Forward the MemBackend through to the Firecracker client


machine.go
```diff
@@ -1171,7 +1171,8 @@ func (m *Machine) CreateSnapshot(ctx context.Context, memFilePath, snapshotPath
 // loadSnapshot loads a snapshot of the VM
 func (m *Machine) loadSnapshot(ctx context.Context, snapshot *SnapshotConfig) error {
 	snapshotParams := &models.SnapshotLoadParams{
-		MemFilePath:         &snapshot.MemFilePath,
+		MemFilePath:         snapshot.MemFilePath,
```
bduffany, line 1174: I was confused why making `mem_file_path` optional caused `MemFilePath` to now be a `string` instead of `*string`. I would have guessed it'd be the other way around - optional should imply `*string` and required should imply `string`, no?

I don't think this is a big deal, was just curious if anyone has any idea why this happens.

bduffany:
Friendly ping @kzys - we have gotten this working successfully over in https://github.com/buildbuddy-io/buildbuddy

opts.go
```diff
@@ -66,3 +69,17 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
 		m.Handlers.FcInit = loadSnapshotHandlerList
 	}
 }
+
+// MemoryBackendType sets the memory backend type to the given value.
+func MemoryBackendType(typ string) WithSnapshotOpt {
+	return func(cfg *SnapshotConfig) {
+		if cfg.MemBackend == nil {
```
fangn2, line 76: If `MemBackend` was previously set,  change type here won't change the path.

fangn2 (commented):
Thanks for your PR! 

Overall looks good to me, I left some comments. 

Regarding CI failure, rebase main should resolve it(had some changes on CI recently).

fangn2 (commented):
Thanks for your PR! 

Overall looks good to me, I left some comments. 

Regarding CI failure, rebase main should resolve it(had some changes on CI recently).

opts.go
```diff
@@ -66,3 +69,17 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
 		m.Handlers.FcInit = loadSnapshotHandlerList
 	}
 }
+
+// MemoryBackendType sets the memory backend type to the given value.
+func MemoryBackendType(typ string) WithSnapshotOpt {
```
swagatbora90, line 74: Please add unit tests specifically for Loadsnapshot in https://github.com/firecracker-microvm/firecracker-go-sdk/blob/main/machine_test.go

opts.go
```diff
@@ -66,3 +69,17 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
...
+	return func(cfg *SnapshotConfig) {
+		if cfg.MemBackend == nil {
+			cfg.MemBackend = &models.MemoryBackend{
+				BackendPath: String(cfg.MemFilePath),
+			}
+		}
+		cfg.MemBackend.BackendType = String(typ)
+		// Clear MemFilePath since only one of MemFilePath or MemBackend is allowed.
```
swagatbora90, line 82: The comment here needs to be more explicit `MemFilePath is ignored if MemBackend is set` and

snapshot.go
```diff
@@ -13,9 +13,19 @@
...
+
 type SnapshotConfig struct {
 	MemFilePath         string
+	MemBackend          *models.MemoryBackend
 	SnapshotPath        string
 	EnableDiffSnapshots bool
 	ResumeVM            bool
 }
+
+func (cfg *SnapshotConfig) GetMemFilePath() string {
```
swagatbora90, line 26: wdyt about changing this func name to `GetMemBackendPath`? Since the return here can either point to a file or an uffd.

bduffany: Sounds good, done

bduffany:
I went ahead and bumped the firecracker version to v1.4.0 since the MemoryBackend is not supported in v1.0.0. I am not sure if this will break anything, though.

opts.go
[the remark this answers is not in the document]
fangn2: Since firecracker [deprecated](https://github.com/firecracker-microvm/firecracker/blob/main/src/api_server/swagger/firecracker.yaml#L1196) `memFilePath `, I think we should use the later approach in case it got removed in the future.

snapshot.go
```diff
@@ -13,9 +13,19 @@
...
+
 type SnapshotConfig struct {
 	MemFilePath         string
+	MemBackend          *models.MemoryBackend
 	SnapshotPath        string
 	EnableDiffSnapshots bool
 	ResumeVM            bool
 }
+
+func (cfg *SnapshotConfig) GetMemBackendPath() string {
```
fangn2, line 26: How about adding some comments to the function something like `if MemBackend is not set, then MemFilePath from SnapshotConfig will be returned` ?    
Just from the name of the function, I would expect nil return if MemBackend not set.

opts.go
```diff
@@ -66,3 +75,19 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
...
+//
+// Note that if MemFilePath is already configured for the snapshot config, it
+// will be ignored, and the backendPath specified here will be used instead.
+func WithMemoryBackend(backendType, backendPath string) WithSnapshotOpt {
+	return func(cfg *SnapshotConfig) {
+		cfg.MemBackend = &models.MemoryBackend{
+			BackendType: String(backendType),
+			BackendPath: String(backendPath),
+		}
+		cfg.MemFilePath = ""
```
fangn2, line 91: nit:
What do you think not clear the `MemFilePath` since it will be ignored by firecracker any way?  
Just set the `MemoryBackend` to align the function description.  
Seems to me clear or not clear both are ok, but users may have to dig deep to find where their previous set value turned empty.

fangn2 (commented):
[CI](https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/3257#0189a195-6db9-4b68-bae8-550a22fc1cf3) is failing due to "go: command not found".

Seems to me your fork is out of sync somehow, at least PR #498 is not in your fork.

opts.go
bduffany: Seems OK to me - Done

opts.go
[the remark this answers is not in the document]
bduffany: > Since firecracker [deprecated](https://github.com/firecracker-microvm/firecracker/blob/main/src/api_server/swagger/firecracker.yaml#L1196) memFilePath , I think we should use the later approach in case it got removed in the future.

Do you mean that we should stop setting `MemFilePath`, and only set `MemBackend`? If yes, I think this will break users who are on older versions of firecracker that don't support `MemBackend`. Is that OK?

Also, I should have mentioned this earlier, but the reason that my initial implementation of `WithMemoryBackend` was so complex because it was trying to be backwards compatible with older firecracker versions. It would always use the deprecated `MemFilePath` (for backwards compatibility), unless `WithMemoryBackend` was used, which would "move" the `MemFilePath` into `MemBackend.BackendPath`. This was the code: https://github.com/firecracker-microvm/firecracker-go-sdk/blob/62f8e172aec24033fc99c007ef093d4f10008e45/opts.go#L74-L85

snapshot.go
bduffany: Done

bduffany (commented):
> [CI](https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/3257#0189a195-6db9-4b68-bae8-550a22fc1cf3) is failing due to "go: command not found".

I merged `main` into this branch - hopefully that fixes CI. Let me know if I should have rebased, or if there's anything else I should do to clean up the commits. (I assumed you will squash commits when merging into `main`.)

opts.go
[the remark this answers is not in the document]
fangn2: > Do you mean that we should stop setting MemFilePath, and only set MemBackend?

I was more leaning to who started using `MemBackend` which is only available in latest firecracker version should use 
```
WithSnapshot(
  "" /*memFile*/, snapshotPath,
  WithMemoryBackend("UFFD", "/path/to/uffd.sock"))
```
instead of 
```
WithSnapshot(
  "/path/to/uffd.sock", snapshotPath,
  WithMemoryBackend("UFFD"))
```
Those who are still using an old firecracker version(assuming they don't use `MemBackend`), they can still do, don't they? 
```
WithSnapshot(
  "" /*memFile*/, snapshotPath)
```
But since firecracker changed `memFile` from a `* string` to `string`, users want to move to new firecracker-go-sdk version anyway need to update their code(backwards compatibility broke).

fangn2:
>I merged main into this branch - hopefully that fixes CI. Let me know if I should have rebased, or if there's anything else I should do to clean up the commits. (I assumed you will squash commits when merging into main.)

CI is working as expected.  
Test [TestLoadSnapshot](https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/3263#0189b850-4490-4be6-8bc5-37d628e5f7fd/582) is failing though.

bduffany:
> Test [TestLoadSnapshot](https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/3263#0189b850-4490-4be6-8bc5-37d628e5f7fd/582) is failing though.

Fixed with https://github.com/firecracker-microvm/firecracker-go-sdk/pull/495/commits/c51a40b159af8a209ff6c1ce7bd3927cacce80b6 - this assertion was unnecessary, I think.

opts.go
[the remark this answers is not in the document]
bduffany: OK. If I understand correctly, you are saying we should keep the current code as-is. If not, please let me know.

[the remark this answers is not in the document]
fangn2: Yeah, keep the current code as it is looks good to me.

fangn2 (approved):

swagatbora90 (approved):

swagatbora90:
Please clean up the commits before merging. I think we should keep the commit to upgrade firecracker ver to v1.4.0 as is and squash the rest

swagatbora90:
nvm, looks like its already merged

PR #544, opened 2024-04-05
Enable resuming snapshot for jailed firecracker

albinvass:
*Issue #, if available:*

*Description of changes:*
This links the memory and snapshot files into the root of the jail and fixes an issue where `WithSnapshots` overrides the injected `LinkFilesHandler`. 

### DIFF
--- a/jailer.go
+++ b/jailer.go
@@ -430,7 +430,7 @@ func LinkFilesHandler(kernelImageFileName string) Handler {
 					"",
 				); err != nil {
 					return fmt.Errorf(
-						"Failed to bind mount %s to %s: %v",
+						"failed to bind mount %s to %s: %v",
 						hostPath,
 						rootfsPath,
 						err,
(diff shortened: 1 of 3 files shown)


jailer.go
```diff
@@ -426,6 +426,24 @@ func LinkFilesHandler(kernelImageFileName string) Handler {
 				m.Cfg.Drives[i].PathOnHost = String(driveFileName)
 			}
 
+			if m.Cfg.hasSnapshot() {
```
swagatbora90, line 429: `hasSnapshot` returns true if either [snapshotPath or memFilePath is non-empty,](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/d7a9b5ea4a5150bfdeda6ab9a20812f2657aac79/machine.go#L174) so we should check if the individual file paths are non-empty, before creating the file link.

swagatbora90 (commented):
Thanks for the PR. Can you please add/update unit test for this change in `jailer_test.go`.

Issue #534, opened 2024-02-05
New info/warning messages from Firecracker when using SDK


alexellis:
Hi there, I've been using Firecracker for a while, mainly via the Go SDK and noticed a bunch of new messages with the v1.5.1 release.

Is there a recommended way to disable these as they're not required and it's not entirely clear where they are coming from.

```bash
2024-02-05T18:34:55.694924617 [anonymous-instance:fc_vcpu 0] Received KVM_SYSTEM_EVENT: type: 2, event: 0
2024-02-05T18:34:55.694954408 [anonymous-instance:main] Vmm is stopping.
2024-02-05T18:34:55.695110571 [anonymous-instance:main] Vmm is stopping.
2024-02-05T18:34:55.725922300 [anonymous-instance:main] Firecracker exited successfully
```

Thanks. Hope the answer is also useful to others landing here.

alexellis:
There are also dozens of messages generated which were not shown before during the initial setup of the VM.. a short snippet:

```
2024-03-22T09:16:24.701798179 [anonymous-instance:main] Running Firecracker v1.5.1
2024-03-22T09:16:24.711814138 [anonymous-instance:fc_api] The API server received a Get request on "/machine-config".
2024-03-22T09:16:24.711860706 [anonymous-instance:fc_api] The request was executed successfully. Status code: 200 OK.
2024-03-22T09:16:24.712049926 [anonymous-instance:fc_api] The API server received a Put request on "/metrics" with body "{\"metrics_path\":\"...-2720694948/metrics.json\"}\n".
2024-03-22T09:16:24.712084928 [anonymous-instance:fc_api] The request was executed successfully. Status code: 204 No Content.
2024-03-22T09:16:24.712214834 [anonymous-instance:fc_api] The API server received a Put request on "/machine-config" with body "{\"mem_size_mib\":7000,\"smt\":false,\"vcpu_count\":8}\n".
2024-03-22T09:16:24.712245202 [anonymous-instance:fc_api] The request was executed successfully. Status code: 204 No Content.
2024-03-22T09:16:24.712379454 [anonymous-instance:fc_api] The API server received a Get request on "/machine-config".
2024-03-22T09:16:24.712407169 [anonymous-instance:fc_api] The request was executed successfully. Status code: 200 OK.
```

alexellis:
This seems like regression because we have the log level set to "Error" and none of these messages are errors.

[...]

PR #574, opened 2024-06-12
Pass `--level` through `jailer` if set

gudmundur:
*Issue #, if available:* https://github.com/firecracker-microvm/firecracker-go-sdk/issues/534

*Description of changes:* In upgrading Firecracker, I noticed a handful of new log lines regarding invocations of the Firecracker API. I've traced them down to [logging the API call](https://github.com/firecracker-microvm/firecracker/blob/e641bfb584a8a029c9585d96da16d6195354fca5/src/firecracker/src/api_server/parsed_request.rs#L69), and [a series of successful request messages](https://github.com/search?q=repo%3Afirecracker-microvm%2Ffirecracker%20%22The%20request%20was%20executed%20successfully%22&type=code). To reduce the amount of log output from Firecracker, I tried adjusting the log level to `Warning`, but it had no effect as the `jailer` and the SDK do not seem to respect the setting if no path or fifo is used. This change changes the `jailer` so that it will propagate the level through to the binary's `--level` flag if set.

### DIFF
--- a/jailer.go
+++ b/jailer.go
@@ -339,6 +339,10 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 	fcArgs := seccompArgs(cfg)
 	fcArgs = append(fcArgs, "--api-sock", machineSocketPath)
 
+	if cfg.LogLevel != "" {
+		fcArgs = append(fcArgs, "--level", cfg.LogLevel)
+	}
+
 	builder := NewJailerCommandBuilder().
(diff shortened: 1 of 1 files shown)


gs0510:
@swagatbora90 @austinvazquez Any chance y'all could look into it? Seems like you folks are the most recent contributors on AWS side. Thank you!

sondavidb (approved):
LGTM, thanks for the contribution!

swagatbora90 (approved):

PR #579, opened 2024-06-24
Revert "Pass firecracker log level through jailer"

swagatbora90:
This reverts commit fe672947d6eeda2d2c99ff1379a766cbc25e0fee.

*Issue #, if available:*
Passing `--log-level` without passing `log-path` [is not supported](https://github.com/firecracker-microvm/firecracker/blob/v1.4.1/src/firecracker/src/main.rs#L182-L186) in firecracker v1.4.1. The above PR breaks that behavior resulting in CI failures.  This behavior has been changed since firecracker v1.5.0, so we should be able to make the change once we update this SDK to support v1.5.0 api. But for now the behavior should remain consistent with v1.4.1 api. 

Adding a separate issue #580 580 to track the addition of this change once we upgrade go-sdk to support firecracker v1.5.0 

@gudmundur @austinvazquez 

*Description of changes:*

### DIFF
--- a/jailer.go
+++ b/jailer.go
@@ -339,10 +339,6 @@ func jail(ctx context.Context, m *Machine, cfg *Config) error {
 	fcArgs := seccompArgs(cfg)
 	fcArgs = append(fcArgs, "--api-sock", machineSocketPath)
 
-	if cfg.LogLevel != "" {
-		fcArgs = append(fcArgs, "--level", cfg.LogLevel)
-	}
-
 	builder := NewJailerCommandBuilder().
 		WithID(cfg.JailerCfg.ID).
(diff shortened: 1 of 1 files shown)


austinvazquez (approved):

henry118 (approved):

gudmundur:
@swagatbora90 Happy to provide a fix to this, is there a way I can see the results from the CI test suites?

swagatbora90:
Thanks @gudmundur. Here is the CI run that failed https://buildkite.com/firecracker-microvm/firecracker-go-sdk/builds/3712#01902801-d6d8-4913-ac88-149051653840

PR #591, opened 2024-08-26
ci: upgrade min go version so that newer versions of containernetworking/cni can build

pendo324:
*Issue #, if available:* failing builds on dependabot PR https://github.com/firecracker-microvm/firecracker-go-sdk/pull/578

*Description of changes:*
Updates the minimum go version since containernetworking/cni now requires go 1.21

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/firecracker-microvm/firecracker-go-sdk
 
-go 1.18
+go 1.21
 
 require (
 	github.com/containerd/fifo v1.1.0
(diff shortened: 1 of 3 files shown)


sondavidb (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/firecracker-microvm/firecracker-go-sdk
 
-go 1.18
+go 1.21
```
Shubhranshu153, line 3: any reason we are useing 1.21 and not 1.22

pendo324: I bumped it to match the minimum version requirements of our dependencies

sondavidb: As this is a library I think it's good practice to use the minimum possible version

Issue #623, opened 2025-02-05
Disabling VMM logs with WithLogger(nil) causes a panic


Gabriel-Ladzaretti:
In `firecracker.NewMachine`, the `nil` logger is checked and handled before the variadic opts are evaluated.

https://github.com/firecracker-microvm/firecracker-go-sdk/blob/03dbe20b038fc2b547ccec0faf139e589028e8d3/machine.go#L387-L391

This means that  `WithLogger(nil)` will cause a panic at line 418 once evaluated:

https://github.com/firecracker-microvm/firecracker-go-sdk/blob/03dbe20b038fc2b547ccec0faf139e589028e8d3/machine.go#L414-L418

This is obviously a misuse, as i suppose an actual logger is expected, and a `noop` logger would suffice for my use case. however, the issue described above is still possible.

Partial stack trace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x854553]

goroutine 1 [running]:
github.com/sirupsen/logrus.(*Entry).Log(0xc000498210?, 0xb?, {0xc0000b9928?, 0x3?, 0x0?})
	/home/gbi/go/pkg/mod/github.com/sirupsen/logrus@v1.9.3/entry.go:303 +0x13
github.com/sirupsen/logrus.(*Entry).Debug(...)
	/home/gbi/go/pkg/mod/github.com/sirupsen/logrus@v1.9.3/entry.go:313
github.com/firecracker-microvm/firecracker-go-sdk.NewMachine({_, _}, {{0xc000498210, 0x11}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, ...}, ...)
	/home/gbi/go/pkg/mod/github.com/firecracker-microvm/firecracker-go-sdk@v1.0.1-0.20250124212734-03dbe20b038f/machine.go:418 +0xde5
```

swagatbora90:
Since the option is supplied by the client, I feel this is more on the client side to handle and this goes for other client options as well. 
However, at least for this particular case, we can move the check for `logger == nil`  after the client options have been applied at L414-L416 to prevent the above seg-fault.

PR #626, opened 2025-02-10
fix: check logger nil after applying opts

swagatbora90:
Fixes: #623 


machine.go
```diff
@@ -415,6 +409,12 @@ func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error)
 		opt(m)
 	}
 
+	if m.logger == nil {
+		logger := log.New()
```
sondavidb, line 413: Just curious — this creates a default logger, not a no-op logger, correct?

I wonder if we might want to allow a way to attach a no-op logger, unless this already exists.

sondavidb (approved):
LGTM

PR #630, opened 2025-02-25
Do not overwrite handlers in LoadSnapshot

### LINKED ISSUE #627: FcInit modified by Jailer overwriten when SnapshotLoad
If I start a Firecracker instance from snapshot with [NewNaiveChrootStrategy](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/563df78e7f85900520cd5ee8ccb24c349cda18b5/jailer.go#L489) with `WithSnapshot` and `WithMemoryBackend` option like below,

```go
instance, err := firecracker.NewMachine(ctx, cfg,
		firecracker.WithSnapshot("", "snapshot.snapshot",
			firecracker.WithMemoryBackend("File",
				"snapshot.mem",
			),
		),
		firecracker.WithLogger(log.StandardLogger().WithContext(vmcs.Context)),
	)
...
err = instance.Start(ctx)
```

instance.Handlers.FcInit [modified by Jailer](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/563df78e7f85900520cd5ee8ccb24c349cda18b5/jailer.go#L511) is overwritten to `loadSnapshotHandlerList` by [opt](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/563df78e7f85900520cd5ee8ccb24c349cda18b5/opts.go#L65)

```go
func NewMachine(ctx context.Context, cfg Config, opts ...Opt) (*Machine, error) {
...
	if cfg.JailerCfg != nil {
		m.Handlers.Validation = m.Handlers.Validation.Append(JailerConfigValidationHandler)
		if err := jail(ctx, m, &cfg); err != nil {
			return nil, err
		}
	} else {
...

	for _, opt := range opts {
		opt(m)
```

```go
func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt {
	return func(m *Machine) {
		...
		m.Handlers.Validation = loadSnapshotValidationHandlerList
		m.Handlers.FcInit = loadSnapshotHandlerList
	}
}

```

I did a workaround by manually inserting the handler after `NewMachine()` is initialized.
Is there any other way to do it the way it is intended?

sondavidb:
*Issue #, if available:*
Closes #627

*Description of changes:*
As LoadSnapshot requires only a subset of the default handlers, it was implemented by assigning the FcInit handlers to a smaller list of handlers. However, this meant that it would overwrite the FcInit handlers given by the jailer. This PR fixes this behavior by just removing the unnecessary handlers, instead of overwriting them entirely.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -315,22 +315,21 @@ var defaultFcInitHandlerList = HandlerList{}.Append(
 	ConfigMmdsHandler,
 )
 
-var loadSnapshotHandlerList = HandlerList{}.Append(
-	SetupNetworkHandler,
-	StartVMMHandler,
-	CreateLogFilesHandler,
-	BootstrapLoggingHandler,
-	LoadSnapshotHandler,
-	AddVsocksHandler,
-)
+// When the machine starts, these handlers cannot run
+// if we plan to load a snapshot. As these handlers are
+// included in defaultFcInitHandlerList, we must remove them
+// if WithSnapshot() has been specified.
+var loadSnapshotRemoveHandlerList = []Handler{
+	SetupKernelArgsHandler,
+	CreateMachineHandler,
+	CreateBootSourceHandler,
+	AttachDrivesHandler,
+	CreateNetworkInterfacesHandler,
+	ConfigMmdsHandler,
+}
 
 var defaultValidationHandlerList = HandlerList{}.Append(
 	NetworkConfigValidationHandler,
-)
-
-var loadSnapshotValidationHandlerList = HandlerList{}.Append(
-	NetworkConfigValidationHandler,
-	LoadSnapshotConfigValidationHandler,
 )
 
 var defaultHandlers = Handlers{
--- a/opts.go
+++ b/opts.go
@@ -71,9 +71,17 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
 			opt(&m.Cfg.Snapshot)
 		}
 
-		m.Handlers.Validation = loadSnapshotValidationHandlerList
-		m.Handlers.FcInit = loadSnapshotHandlerList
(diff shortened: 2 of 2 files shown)


opts.go
```diff
@@ -71,11 +71,19 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
...
-		m.Handlers.Validation = loadSnapshotValidationHandlerList
-		m.Handlers.FcInit = loadSnapshotHandlerList
+		m.Handlers.Validation = m.Handlers.Validation.Append(LoadSnapshotConfigValidationHandler)
+		m.Handlers.FcInit = modifyHandlersForLoadSnapshot(m.Handlers.FcInit)
 	}
 }
 
+func modifyHandlersForLoadSnapshot(l HandlerList) HandlerList {
+	for _, h := range loadSnapshotRemoveHandlerList {
+		l = l.Remove(h.Name)
```
swagatbora90, line 81: Does `Remove` handle the case when `h.Name` is not present in the Handler list?

handlers.go
```diff
@@ -315,24 +315,19 @@ var defaultFcInitHandlerList = HandlerList{}.Append(
...
 
-var loadSnapshotHandlerList = HandlerList{}.Append(
-	SetupNetworkHandler,
-	StartVMMHandler,
-	CreateLogFilesHandler,
-	BootstrapLoggingHandler,
-	LoadSnapshotHandler,
-	AddVsocksHandler,
-)
+var loadSnapshotRemoveHandlerList = []Handler{
```
swagatbora90, line 318: Can you please add a comment here for future reference on why we need to have a filter list?

opts.go
sondavidb: Seems it should return the same HandlerList: https://github.com/firecracker-microvm/firecracker-go-sdk/blob/main/handlers.go#L448

swagatbora90 (approved):

chews93319 (approved):

PR #642, opened 2025-03-13
Fix BuildKite CI for kernel image path errors when loading snapshots

austinvazquez:
Signed-off-by: Austin Vazquez <macedonv@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/opts.go
+++ b/opts.go
@@ -71,7 +71,7 @@ func WithSnapshot(memFilePath, snapshotPath string, opts ...WithSnapshotOpt) Opt
 			opt(&m.Cfg.Snapshot)
 		}
 
-		m.Handlers.Validation = m.Handlers.Validation.Append(LoadSnapshotConfigValidationHandler)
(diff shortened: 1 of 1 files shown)


austinvazquez:
So my debugging revealed that the BuildKite CI failures are likely caused by this PR: https://github.com/firecracker-microvm/firecracker-go-sdk/pull/630

Previously we would override the configuration validation handler list when loading snapshots and have broken that behavior. I don't think we need to have the full configuration validation. e.g. the current error is due to not providing a kernel image path which is a valid use case for loading snapshots.

austinvazquez:
Chatted to @sondavidb offline. It seems the fix was for users that specify their own handlers so overwriting validation here might not produce the desired effect. e.g. users may have validation for their handlers. So alternatively suggesting to remove the configuration validation when loading a snapshot.

sondavidb (approved):

PR #666, opened 2025-07-23
Fix race condition with daemonized jailer in startVMM

adamthomason:
*Description of changes:*

When using jailer with Daemonize: true, the jailer parent process exits immediately after forking the child firecracker process. This caused a race condition where the SDK's process monitoring goroutine would detect the parent exit and send to errCh before waitForSocket() could establish the socket connection, causing startVMM to fail prematurely.

This change modifies the process monitoring logic to treat a clean exit (status=0) of a daemonized jailer parent as expected behavior rather than an error condition. The goroutine now returns early in this case, allowing waitForSocket() and other initialization logic to complete normally.

The actual firecracker child process continues running in daemon mode and creates the API socket as expected.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -588,6 +588,16 @@ func (m *Machine) startVMM(ctx context.Context) error {
 	errCh := make(chan error)
 	go func() {
 		waitErr := m.cmd.Wait()
+
+		// If using daemonized jailer and parent exits cleanly,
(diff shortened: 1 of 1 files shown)


adamthomason:
Thanks for the review @sondavidb. Agreed, it would be nice for it to be added to tests but I too couldn't quite think of a nice way of doing this.

